PR #4, opened 2021-04-09
Spherical unit test: Filters and balance law clean

LenkaNovak:
Added over-integration, exponential filters, advection-compatible BCs, and minor modifications to the balance law for conciseness.


christophernhill:
@LenkaNovak this looks good to merge to me. I think we are still in early stages, so as long as its not obviously broken we should keep moving....

christophernhill (approved):

christophernhill:
@LenkaNovak should we merge this? Thanks, Chris

PR #8, opened 2021-06-07
Simple Slab Land

LenkaNovak:
Dry heat advection-diffusion equation in an atmospheric single stack coupled to a prognostic surface temperature equation reflecting a slab land. Dry setup. This case is mainly for software design before implementing more complex cases. 

Design: https://github.com/CliMA/CouplerMachine/blob/ln/slabland/experiments/SlabLand/README.md
Conservation tests: [under the 16 June section](https://docs.google.com/document/d/1JKK8wFKPq3Jo3D4flXZiY3WAUPqYE19pXRvwJgpDwjw/edit)


Project.toml
```diff
@@ -13,6 +13,7 @@ KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
jb-mackay, line 16: We should keep our dependencies as trim as possible - so no `Plots` in the CouplerMachine dependencies (`Pkg.rm("Plots")` in the CouplerMachine environment). If you add `Plots` from the default Julia environment (no `--project`) then it will still be accessible when you want to use it in other projects without adding it to the dependency list:

`julia -e 'using Pkg; Pkg.add("Plots")'`

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,156 @@
...
+    - Simplifications
+        - $g_a$ is constant
+        - (μ, ν, $\kappa$) = (0, 0, 1e-5)
+        - $\vec{u} = 0$
+        - $G = 0$ and $F_a = 0$
+    - [results](https://docs.google.com/document/d/1JKK8wFKPq3Jo3D4flXZiY3WAUPqYE19pXRvwJgpDwjw/edit)
+
+ (constant conductances and ConstantViscosity( FT(0), FT(0), FT(1e-5), u = 0 )
+
+## **Pipeline** (probably delete)
```
jb-mackay, line 133: I think this pipeline could be worth holding onto for a little while - it's a clear walkthrough of what is happening during the coupling.

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,161 @@
...
+- Sensible heat flux at surface
+    $$SH = cp_d g_a (T_{sfc} - T_a) = cp_d g_a (T_{sfc} - T_a)$$
+- Latent heat flux at surface
+    $$LH = \lambda g_w (q_{sat}(T_{sfc},p_{sfc}) - q_a)$$
+- Downward-pointing short-wave radiative flux
+    $$R_{SW} = F_{sfc}^d - F_{sfc}^u = (1-\alpha)\tau F_{sol} * (1 + \sin(2π τ_d^{-1} t))$$
+- Upward-pointing net long-wave flux
+    $$R_{LW} = F_{sfc}^u - F_{sfc}^d = \epsilon \sigma \theta_{sfc}^4 - \epsilon F_{a}$$
+- Soil storage
+    $$G = \kappa_s (\theta_{sfc}-\theta_h)/h_s$$
```
kmdeck, line 81: \theta_{sfc} -> T_{sfc}

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,161 @@
...
+2. Flux accumulation
+    - In order to integrate fluxes in time consistently with the Atmos time stepper, we define a new prognostic variable `F_ρθ_accum` whose source function accumulates the fluxes, so that $F\_ρθ\_accum = \int F_{tot} dt$
+
+            @inline function source!(atmos_model::ModelSetup, _...)
+                source.F_ρθ_accum = calculate_land_sfc_fluxes(atmos_model, _...) 
+            end
+
+3. Flux / state storage and communication
+    - the coupler stores 2 fields:
+        - `EnergyFluxAtmos` which collects `F_ρθ_accum / Δt` calculated and accumulated in Atmos (with $Δt$ denoting the period of the coupling cycle), and is called by land for its $T_{sfc}$ equation
```
kmdeck, line 117: maybe rename, e.g. BoundaryEnergyFlux?

docs/src/boundary_conditions_doc.md
```diff
@@ -0,0 +1,165 @@
+# Boundary Conditions (BCs)
+
+This section is an attempt to bridge mathematical considerations of BCs and their implementations in climate models.
```
jb-mackay, line 3: I think this doc is great and very helpful, but may perhaps belong with the RQA work in ClimateMachine. What do you think?

experiments/SlabLand/coupler/put_get_calls.jl
```diff
@@ -0,0 +1,98 @@
+# Land put and get calls
```
jb-mackay, line 1: I like the coupler directory set up and am glad we are separating out the pre- and post-step function definition from the balance laws; so this is great! There will probably be some other coupler-specific configuration/definitions that we'll need to do in the future so it's good to have a separate place for it.

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,161 @@
...
+$$
+with μ, ν and κ representing the respective viscosity and diffusivity coefficients. The boundary conditions are:
+- Upper (external) boundary:
+    - ρ: `Impenetrable()` 
+    - u: `FreeSlip()`
+    - θ: `Insulating()`
+- Lower (coupled) boundary:
+    - ρ: `Impenetrable()`  
+    - u: `FreeSlip()`
+    - θ: `CoupledPrimary()`: $\frac{\partial θ}{\partial t} = - ∇ \cdot F_{tot}$ where $F_{tot}$ is the surface flux calculated by `calculate_land_sfc_fluxes(_...)` using Atmos and Land properties (see below)
```
kmdeck, line 40: Is the boundary condition $$\partial θ / \partial n = F_{tot}$$? (not the divergence of the flux)

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+1) Coupler sends its field `LandSurfaceTemperature` and transforms it into Atmos field `auxiliary.T_sfc`
+2) Atmos initializes its state according to the user-defined ICs
+3) Atmos performs all its timesteps within the coupling cycle, with the boundary conditions at the coupled boundary setting the total normal flux to be equal to the flux coming from the land (`fluxᵀn.ρθ = - F_tot / cp_d`, so that $\partial_t \rho \theta = \nabla_z \sdot F_{tot}/cp_d$)
+4) The same flux `F_tot` is calculated in the `source!` function and saved as `state.F_ρθ_accum`, which ensures that the flux will be integrated and accumulated in time, consistent with the Atmos time stepping. 
+5) Coupler converts Atmos field `state.F_ρθ_accum` to the coupler field `EnergyFluxAtmos` 
+6) Coupler sends its field `EnergyFluxAtmos` and transforms it into Land field `auxiliary.F_ρθ_prescribed`
+5) Land performs its own initialization and physics (`G`), which is then added to the  `auxiliary.F_ρθ_prescribed` (corresponding to $F_{tot}$ above) in the `source!` function to solve:
+$$T_s^{n+1} = T_s^n +\frac{F_{tot}^{n...}}{\rho_s c_s h_s}\Delta t$$
```
kmdeck, line 153: i also found this pipeline really helpful. 

the use of F_tot is inconsistent here - in line 87 it refers to the non G fluxes, here it refers to G + non G fluxes.

kmdeck: I also wonder if using the subscript n with F_tot is correct? If Im understanding, it does use land values from the previous land step, but it's an accumulation of the flux (using atmos values at intermediate t) over the whole land step. Might be worth stating that?

If one wanted to use a different explicit stepper on the land side (not forward Euler), is that possible? Would the "coupling cycle" dt be e.g. the different dts for the RK stages?

experiments/SlabLand/README.md
```diff
@@ -0,0 +1,155 @@
...
+- we're using our sequential coupler, which is described here
+### 2. Surface Flux Handling
+1. Flux calculation
+    - function `calculate_land_sfc_fluxes(_...)` is defined in `coupler/surface_fluxes.jl`, and is used for:
+        - the Atmos Robin boundary condition calculation (see above)
+        - for surface flux accumulation (next point). 
+    - it is called by Atmos, which is assumed to be the model with the shortest timestep. This means that the fluxes are only calculated once and are consistent with Atmos's time stepping.
+
+2. Flux accumulation
+    - In order to integrate fluxes in time consistently with the Atmos time stepper, we define a new prognostic variable `F_ρθ_accum` whose source function accumulates the fluxes, so that $F\_ρθ\_accum = \int F_{tot} dt$
```
kmdeck, line 104: maybe this language was causing confusion? the source function doesnt accumulate the fluxes, it is the RHS term of the ODE d(F_rho theta_accum)/dt = F_tot

experiments/SlabLand/balance_law_interface_atmos.jl
```diff
@@ -0,0 +1,197 @@
...
+
+function nodal_init_state_auxiliary!(
+    m::ModelSetup,
+    state_auxiliary,
+    tmp,
+    geom,
+)
+    init_state_auxiliary!(m, m.physics.orientation, state_auxiliary, geom)
+
+    state_auxiliary.T_sfc = 0
```
kmdeck, line 90: is this consistent with the IC of the land model?

experiments/SlabLand/balance_law_interface_slabland.jl
```diff
@@ -0,0 +1,229 @@
...
+    p = model.parameters
+    G    = p.κ_s * (state.T_sfc - p.T_h) / p.h_s # simple soil physics
+    source.T_sfc = - (aux.F_ρθ_prescribed + G) / (p.ρ_s * p.c_s * p.h_s)
+
+    return nothing
+end
+
+"""
+    Boundary conditions
+"""
```
kmdeck, line 154: Are these BC methods needed yet? Right now the PDE for land is just an ODE...

kmdeck:
it seems like a lot of the files in this PR are not specific to the slab land experiment, and looks like some are already in the repo (e.g. utilities/sphere_utils).

experiments/SlabLand/parameters_initialconditions.jl
```diff
@@ -0,0 +1,42 @@
+########
+# Set up parameters
+########
+parameters = (
+        xmax = 10000,   # horizontal extent of the SingleStack in the x direction [m]
+        ymax = 500,     # horizontal extent of the SingleStack in the y direction [m]
+        zmax = 10000,   # vertical extent of the SingleStack [m]
+        g    = 9.81,    # gravitation acceleration [m /s^2]
```
kmdeck, line 8: worth using Clima parameters at this stage?

experiments/SlabLand/README.md
LenkaNovak: Good catch! :)

docs/src/boundary_conditions_doc.md
LenkaNovak: RQA said they will add it once they've revamped their BCs, so perhaps we can leave it here until then?

experiments/SlabLand/README.md
LenkaNovak: Not quite, the way we set the BCs in our case is ultimately via the normal flux, i.e. `fluxᵀn.ρθ += - F_tot / scaling_params` (referred to as `local_flux` in different parts of the code), and then the model calculates the div of that.

experiments/SlabLand/README.md
LenkaNovak: Umm, it basically calculates `d (F_ρθ_accum) +=  \int F_{tot} dt`. I will clarify the doc, but yes, accumulating via the source function is not ideal. In the CM that's the only easy option we have, but we will be giving feedback to the CMCore team on this to make it less confusing in the new version :).

experiments/SlabLand/balance_law_interface_atmos.jl
LenkaNovak: Yes, at the coupled boundary, `aux.T_sfc` (of Atmos) is set to be equal to `state.T_sfc` (of Land) at the beginning of each Atmos coupling cycle. This line just initializes the elements of the MPIStateArray that are not at the boundary (so they should never be used). Technically, this line should not be needed but I was occasionally getting Nans if the whole field was not initialized. 🤷

experiments/SlabLand/README.md
kmdeck: hey! not being critical, just more clarifying for my own sake. I still dont understand the confusion re: using the language source - it seems as though this is exactly the same usage (a RHS term for a differential equation) as we always use it, except in this case the ODE is d(F_rho theta_accum)/dt = F_tot.

I can chat with Valeria more though since she was the one who brought it up!

experiments/SlabLand/parameters_initialconditions.jl
LenkaNovak: I think for this illustrative example it's ok to leave the params list as is, but it's definitely worth doing for the more formal test cases that we will be publishing as part of CliMA. :)

experiments/SlabLand/README.md
kmdeck: ok! that is what I thought. I was just commenting on the way it is written. (the math expression does not reflect the BC, which is a Neumann BC on flux like you wrote  - the math instead reflects the entire RHS at the boundary).

LenkaNovak:
> it seems like a lot of the files in this PR are not specific to the slab land experiment, and looks like some are already in the repo (e.g. utilities/sphere_utils, grid stuff, numerics (filters...), even physics). maybe can be reorganized to have a single copy of these files?

Good point. These are copied from the RQA interface. We are keeping these folders under each example separate, because the RQA interface was evolving as we were writing these different test cases, so we wanted to have a stable backend for each case. Once the `Aleph.jl` repo/module (where the RQA code will live) is ready though, we will import these files from there.

experiments/SlabLand/README.md
jb-mackay: @kmdeck unless I'm mistaken, yes, the choice of the land timestepper will determine how $T_s^{n+1}$ is computed, and this does not have to be forward Euler, of course.

The "coupling cycle" refers to the synchronization timescale of the two solvers. At the beginning/end of the coupling cycle, coupled field information is exchanged. Within that coupling cycle, the individual components can take as many substeps as they need, but they don't exchange information.

jb-mackay (approved):
Looks good @LenkaNovak. Once you have addressed @kmdeck's comments feel free to merge. As mentioned, we'll have to re-tidy this and some other experiments when the RQA interface if formalized.

experiments/SlabLand/balance_law_interface_slabland.jl
LenkaNovak: Eeep! Well spotted, I've just removed them! :)

experiments/SlabLand/README.md
kmdeck: thanks! @jb-mackay. I'll think about it more! May still follow up with ?s, but that sounds good.

kmdeck (approved):
looks good to me @LenkaNovak @jb-mackay. nice job!!

PR #11, opened 2021-07-09
CplSolver Unit Tests

jb-mackay:
- Adds unit tests for the CplSolver
- Adds callback storage to the CplModel struct

### DIFF
--- a/src/CplModel.jl
+++ b/src/CplModel.jl
@@ -10,7 +10,7 @@ struct CplModel{G, D, B, S, TS}
     discretization::D
     boundary::B
     state::S
-    stepper::TS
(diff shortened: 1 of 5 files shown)


LenkaNovak:
Looks great! Do we need L37-41 in `cplstate_interface.jl`, since we test units above and this is just overwriting the coupler field? Otherwise happy to merge! 👍

jb-mackay:
> Looks great! Do we need L37-41 in `cplstate_interface.jl`, since we test units above and this is just overwriting the coupler field? Otherwise happy to merge! 👍

Good catch, modified these lines so they act as intended.

PR #14, opened 2021-08-27
Move column test cases from climacore repo

LenkaNovak:
Initial move of ClimaCore test cases.

Co-authored-by: Ben Mackay <jb-mackay@users.noreply.github.com>
Co-authored-by: kmdeck <kmdeck@users.noreply.github.com>
Co-authored-by: Gregory L. Wagner <gregory.leclaire.wagner@gmail.com>

Test case 3 still needs to be optimized for a better interface.


jb-mackay:
@LenkaNovak do we need to do additional work with the ClimaCore examples for this PR?

Looking ahead, the ClimateMachine tests will break when we update the interface. We should discuss if we want to maintain some of them. The ClimaCore cases are currently independent of the CouplerMachine source but should eventually use the CouplerMachine.

PR #16, opened 2021-09-02
Coupled Timestepping Interface

jb-mackay:
Introduces the coupled timestepping interface in which component models provide `step!` methods used in the coupled system driver. Updates to coupler state interface will occur in a separate PR.

To do:
- [x] Implement basic `run!` function that executes simulation timestepping
- [x] Update unit tests
- [x] ~~Coupled atmos-ocean-land example implementation~~ (Future PR)
- [x] Update dependencies
- [x] Update docs


jb-mackay:
Documentation preview: https://clima.github.io/CouplerMachine/previews/PR16/couplerstate/

LenkaNovak (approved):
Looks good! Do we want to remove the ClimateMachine-compatible interface completely though? I wonder whether it's still useful to keep it somewhere until have a way to support the CM with the new interface. I guess we can always go back the commit history once we come to that bridge though...

jb-mackay:
> Looks good! Do we want to remove the ClimateMachine-compatible interface completely though? I wonder whether it's still useful to keep it somewhere until have a way to support the CM with the new interface. I guess we can always go back the commit history once we come to that bridge though...

I think it is straightforward to use the new interface with a ClimateMachine simulation via [`ODESolvers.general_dostep!`](https://github.com/CliMA/ClimateMachine.jl/blob/5563a272aedbb85c44b4f256ca51d58e748e41cb/src/Numerics/ODESolvers/ODESolvers.jl#L49).

PR #19, opened 2021-09-14
[WIP] SurfaceFluxes.jl plugin

LenkaNovak:
Provides coupled and uncoupled test cases that use the Nishizawa & Kitamura (2018) formulation using SurfaceFluxes.jl. 

## TODO minimal example (~TC1: heat diffusion column)
- [ ] integrate to run with ClimaCore/ ClimaAtmos interface, ensuring type compatibility
- [ ] note changes in SF and NS + revamp the SurfaceFluxes.jl repo with original developers
- [ ] test for longer simtime and compare with the bulk formula test
- [ ] note if still conservation problems due to divergence operator
- [ ] clean up @shows + document the SF functions (+ make README consistent)
- [ ] hook up to CI (along with TC1)

## TODO physical example (~TC3: Atmos Ekman Column and Land's Richards equations;  minus ocean)
- [ ] coupler interface enhancement: add coupler flux calculation (also change this for TC3)
- [ ] couple stress, moisture
- [ ] add to SF: $g_{ac}$ (and $g_{s}$) calculation. $g_s$ needed only for evaporation and latent heat fluxes
- [ ] hook up to CI


charleskawczynski:
I’ve been working on the surface fluxes issues, they should be fixed in less than a week, in case you want to avoid duplicating code.

LenkaNovak:
> I’ve been working on the surface fluxes issues, they should be fixed in less than a week, in case you want to avoid duplicating code.

Ahh nice, thanks! I've been trying to get this minimal example going, like I promised you. Do you have a PR I could check? :)

charleskawczynski:
I’ve removed NonLinearSolvers.jl’s dependence on CUDA, so you might want to just re-try using SurfaceFluxes.jl. I started the registration process for SurfaceFluxes.jl, here https://github.com/JuliaRegistries/General/pull/44816#issuecomment-918765876

LenkaNovak:
> I’ve removed NonLinearSolvers.jl’s dependence on CUDA, so you might want to just re-try using SurfaceFluxes.jl. I started the registration process for SurfaceFluxes.jl, here [JuliaRegistries/General#44816 (comment)](https://github.com/JuliaRegistries/General/pull/44816#issuecomment-918765876)

That's great, thanks for doing that! I had to deal with some type inconsistencies as well to get it to work, but it's a pretty hacky solution rn. It's probably easiest if I show you the code, when you have a mo. Lmk when works for you.

PR #24, opened 2021-09-28
SurfaceFluxes.jl plugin

LenkaNovak:
Provides a coupled test case that uses the dry Nishizawa & Kitamura (2018) formulation using `SurfaceFluxes.jl`. 

## TODO:
- [x] integrate to run with ClimaCore/ ClimaAtmos interface, ensuring type compatibility
- [x] note changes in SF and NS + revamp the SurfaceFluxes.jl repo with original developers
- [x] test for longer simtime and compare with the bulk formula test + conservation (conservation achieved)
- [x] note if still conservation problems due to divergence operator - resolved (needed to apply boundary conditions only to diffusive fluxes, not advective fluxes as was the default)
- [x] clean up @shows + document the SF functions (+ make README consistent)
- [x] hook up to CI 

Rebased from [this](https://github.com/CliMA/CouplerMachine/pull/19) PR.


experiments/ClimaCore/Project.toml
```diff
@@ -1,6 +1,5 @@
 [deps]
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
-ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
```
jb-mackay, line 3: Was there a reason for these to be removed? I think we can revert these changes to this file.

experiments/ClimaCore/tc4_atm-lnd-sfc-fluxes/Project.toml
```diff
@@ -0,0 +1,18 @@
+[deps]
```
jb-mackay, line 1: Since this shares most dependencies with the other ClimaCore examples, I think we can just add `DocStringExtensions`, `KernelAbstactions`, and `SurfaceFluxes` to the `examples/ClimaCore/Project.toml`. I guess I am also okay with each test case folder having its own `Project.toml`, but we should be consistent here.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
jb-mackay, line 8: I think it is standard now to have a `Project.toml` for the test directory that is separate from the package dependencies, so we can create that now :)

test/experiments_ClimaCore.jl
```diff
@@ -0,0 +1,80 @@
+# Experiments
+
+using Test
+using Pkg
+
+Pkg.activate(joinpath(@__DIR__,"../experiments/ClimaCore/tc4_atm-lnd-sfc-fluxes/"))
+
+Pkg.add(url="https://github.com/CliMA/ClimaAtmos.jl",rev="main" )
+Pkg.add(url="https://github.com/CliMA/SurfaceFluxes.jl",rev="sf_rc0.2" )
```
jb-mackay, line 9: We can add these to the `Project.toml`; we have the `Manifest.toml` checked in so the appropriate branches of these packages will be used.

test/experiments_ClimaCore.jl
```diff
@@ -0,0 +1,80 @@
...
+
+Pkg.instantiate()
+Pkg.precompile()
```
jb-mackay, line 12: I need to look at CI setup to see if we can avoid this kind of environment activation. We could always add all this to the `test/Project.toml` I mentioned above.

experiments/ClimaCore/tc4_atm-lnd-sfc-fluxes/Project.toml
jb-mackay: Changing my mind here; I think it will be less convoluted to have separate projects for each experiment. This will give us greater control over dependency versioning and branches. It also follows ClimaCore.

jb-mackay: We can add these as we update the other experiments in future PRs.

test/experiments_ClimaCore.jl
```diff
@@ -0,0 +1,73 @@
...
+using Test
+
+include(joinpath(@__DIR__,"../experiments/ClimaCore/tc4_atm-lnd-sfc-fluxes/experiment.jl"))
+
+
+FT = Float64
+parameters = (
+        # timestepping parameters 
+        Δt_min = 0.01, # minimum model timestep [s]
+        timerange = (0.0, 1.0),  # start time and end time [s]
```
jb-mackay, line 14: I shortened this time so the test completes more quickly; @LenkaNovak is there a minimum time we want to run this for in the CI?

LenkaNovak: Sounds good. Since this is only an integration test, 100 time steps should be enough :).

jb-mackay:
@charleskawczynski can I get your insight on why the Julia formatter is struggling here?

charleskawczynski:
Looking...

charleskawczynski:
These struct definitions are incomplete:
```julia
struct Gravity{𝒯} <: AbstractGravity{𝒯}
struct Buoyancy{𝒯} <: AbstractGravity{𝒯}
```

charleskawczynski:
One nice trick with debugging formatter problems is opening the Julia REPL and trying to load the offending file. This often gives more informative error messages (since the formatter doesn't try to reason about the code in the same way that happens during code loading).

jb-mackay:
@charleskawczynski does the JuliaFormatter action fail if the formatter hasn't been applied and then the formatted code committed? Or is this action failing for another reason?

jb-mackay (approved):
Looks good - just need the formatter to be happy :)

charleskawczynski:
> Looks good - just need the formatter to be happy :)

We're not currently depending on the success of the formatter ([here](https://github.com/CliMA/CouplerMachine/blob/main/bors.toml#L2-L5)), so we can merge this as is. If you'd like to apply the formatter for this PR, we could do that but, since the PR is already pretty large, the formatting could be applied in a separate PR (and then add the format entry in the bors toml file).

charleskawczynski:
bors ping

jb-mackay:
bors ping

PR #25, opened 2021-10-07
Test case 1 update + conversion to tutorial

LenkaNovak:
This is the simplest coupler test case that can be used for quick prototyping without interface wrappers. This PR updates the test case for the recent upstream (i.e. ClimaCore) changes, and sets up the run file as a tutorial. Once this is merged, it will be set up for automatic CI. 

TODO
- [x] setup Documenter (tutorial's [index.html](https://drive.google.com/file/d/1udu8zhlTlUnd86wvaMuBCKbF1sKxLCjJ/view?usp=sharing))
- [x] setup auto-CI (Buildkite will be hooked up in a separate PR after renaming the repo)


LenkaNovak:
@jb-mackay WYGAC LMK if you'd like me to update TC4 (I have it running locally) or if it's worth switching from ClimaCore 0.1.1 to 0.2.0 half way through `test/runtests.jl`. Thanks! :)

docs/src/run.md
```diff
@@ -0,0 +1,406 @@
+# Heat Equation + Slab Tutorial
```
jb-mackay, line 1: @LenkaNovak using Literate.jl, we should be able to only have one version of the tutorial file (the one that lives in `experiments/...`) which then generates a file to be used by `Documenter.jl` when the docs are generated. Was there a reason for adding another markdown file?

Issue #42, opened 2022-04-04
Implement prescribed sea ice


LenkaNovak:
## Purpose
Implement sea ice (prescribed sea ice thickness [e.g., 2 m] and sea ice concentration, with[ Semtner](https://www.atmosp.physics.utoronto.ca/people/guido/PHY2502/articles/seaice-landice/AJSemtner_1976.pdf) 0-layer sea ice model (simplified as in[ this code](https://github.com/sally-xiyue/fms-idealized/blob/sea_ice_v1.0/exp/sea_ice/srcmods/mixed_layer.f90), documented in[ this paper](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2021MS002671)) 

## Final Key Result
- Smooth moist BW with multiple surfaces: atmos (full) + land (slab) + hydrology (bucket) + ocean (prescribed) + sea ice (prescribed) with a land-sea mask

## Measure of success 
(literature comparison / conservation / performance benchmark)
- [x] physical tests
- [x] conservation of tracers
- [x] performance comparable to standalone run

## Risks
- no domain expertise but should be ok

## Design, Features and Dependencies
- [x] approved design [here](https://github.com/CliMA/ClimaCoupler.jl/blob/fd1eb0bc220ad34541db15478a6bb4d247a2f623/experiments/ClimaCore/sea-ice/README.md)
- [x] implement sea ice in a SCM [#46](https://github.com/CliMA/ClimaCoupler.jl/pull/46)
- [x] use the sea ice model on sphere with a land-sea mask
- [x] use prescribed ocean `T_sfc` and `h_i` data
- [x] pull sea ice out of the test case to the coupler's `src`
- [x] sea-ice mask
- [x] add file reader if want to replace sea-ice concentration and depth with prescribed values, e.g. like in CESM AMIP (https://www.cesm.ucar.edu/models/ccsm4.0/cice/doc/node23.html)
- [x] add ocean Q flux

## Driver PR(s)
- [x] sea ice in a SCM  [#46](https://github.com/CliMA/ClimaCoupler.jl/pull/46)
- [x] sea ice on sphere - prescribed #50 , #51 
- [x] sea ice on sphere - dynamical [#69](https://github.com/CliMA/ClimaCoupler.jl/pull/69)
- [x] clean-up PR
  - introduce more coupler-specific interface abstractions https://github.com/CliMA/ClimaCoupler.jl/issues/44
[...]

PR #46, opened 2022-04-18
0-layer sea-ice model in SCM

LenkaNovak:
- addresses the first stage of #42 

## Steps
- [x] [design](https://github.com/CliMA/ClimaCoupler.jl/blob/6f8704e925c0ba0fc5bd20f0e094deeb246b2f38/experiments/ClimaCore/sea-ice/README.md)
- [x] implement the Semster 1976 model
- [x] test using a single column
- [x] conservation test
Screen Shot 2022-04-27 at 5 50 59 PM

- [x] clean up

## Proposed assignees
- Lenka, Akshay, Ben

## Proposed Reviewers
- Zhaoyi, Kat

## TODO in following PRs
- add file reader if want to replace sea-ice concentration and depth with prescribed values, e.g. like in CESM AMIP (https://www.cesm.ucar.edu/models/ccsm4.0/cice/doc/node23.html)
- spherical application with land-sea mask
- add ocean Q flux
- add an existing Newton iterator for more iters (?) 
- apply with SurfaceFluxes.jl and compare with [Semtner 1976](https://www.atmosp.physics.utoronto.ca/people/guido/PHY2502/articles/seaice-landice/AJSemtner_1976.pdf)'s quantitative results

### DIFF
--- a/experiments/ClimaCore/sea-ice/README.md
+++ b/experiments/ClimaCore/sea-ice/README.md
@@ -1,0 +1,72 @@
+# **Simple Sea Ice Model**
+
+# Semtner's Zero-layer Model
+
+Sea ice is approximated as a model that does not occupy a particular gridpoint between the mixed layer (ocean) and the atmosphere. It is essentially implemented in each domain column as an ODE, with no horizontal transport between the columns. In the absence of ice, the sea-ice model reduces to the slab ocean formulation. The ice is assumed to have a negligible heat capacity (so there is no energy storage due to internal temperature changes of the ice). The only storage changes arise from the ice thickness changes, which result from temperature differences at the ice surface or at the ice base. 
+
+We followed the FMS implementation as in [Zhang et al 22](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2021MS002671), which can be found on [GitHub](https://github.com/sally-xiyue/fms-idealized/blob/sea_ice_v1.0/exp/sea_ice/srcmods/mixed_layer.f90) and is itself a modification of the 0-layer model of [Semtner 1976](https://www.atmosp.physics.utoronto.ca/people/guido/PHY2502/articles/seaice-landice/AJSemtner_1976.pdf) (appendix). Chronologically, the algorithm follows these steps, with all fluxes defined positive upward:
+
+## 1. Ice thickness, $h_i$
+$$
+L_i \frac{dh_i}{dt} = F_{atm} - F_{base}
+$$
+- with the latent heat of fusion, $L_i=3 \times 10^8$ J m$^{-3}$, and where the (upward-pointing) flux into the atmosphere is
+$$
+F_{atm} = F_{rad} + F_{SH} + F_{LH} \approx \lambda (-{T_{sfc}} - T_{atm})
+$$
+- where the latter approximation was used for testing in earlier prototypes. $F_{atm}$ will be obtained from the atmospheric model (via the coupler). The flux at the ice base from the mixed layer is
+$$
+F_{base} = F_0(T_{ml} - T_{melt})
+$$
+- where $T_{melt} = 273.16$ K is the freezing temperature, and the basal heat coefficient $F_0 = 120$ W m$^{-2}$ K$^{-1}$. 
+
+## 2. Ocean mixed layer temperature, $T_{ml}$
+- $T_{ml}$ is the standard slab ocean formulation in ice-free conditions:
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{atm}
+$$
+- while ice-covered conditions require that:
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{base}
+$$
+
+## 3. Transitions between ice free and ice covered conditions
+- If the updated $T_{ml}^{t+1} < T_{melt}$, set $T_{ml}^{t+1} = T_{melt}$ and grow ice ($h_i^{t+1}$) due to the corresponding energy deficit.
+- If the updated $h_i^{t+1} <= 0$ from a non-zero $h_i^t$, adjust $h_i^{t+1} = 0$ and use the surplus energy to warm the mixed layer. 
+
+## 4. Surface temperature ($T_s$)
+- $T_s$ is determined implicitly using a balance between $F_{atm}(T_s)$ and the conductive heat flux through the ice slab, $F_{ice}$:
+$$
+F_{atm} = F_{ice} = k_i \frac{T_{melt} - T_s}{h_i}
+$$
+- where $k_i = 2$ W m$^{-2}$ k$^{-1}$ is the thermal conductivity of ice.
+- currently the implicit solve is implemented as one Newton iteration:
+$$
+T_s^{t+1} = T_s + \frac{F}{dF /d T_s}  = T_s^{t} + \frac{- F_{atm}^t + F_{ice}^{t+1}}{k_i/h_i^{t+1} + d F_{atm}^t / d T_s^t}   
+$$
+- where $h_i^{t+1}$ is the updated $h^i$ from the previous section, and $d F_{atm}^t / d T_s^t$ needs to be supplied from the atmosphere model (or crudely calculated in the coupler, given $T_s$, turbulent diffusivities and transfer coefficients, and atmos state). 
+- Where $T_s^{t+1} > T_{melt}$, we set $T_s^{t+1} = T_{melt}$. Where there is no ice $T_s^{t+1} = T_{ml}^{t+1}$.
+
+## 5. Update ice mask
+- `mask = 1` if ice and `mask = 0` if no ice
+
+# Q flux (optional)
+- We can add an additional flux to the RHS of the $T_{ml}$ equations in (2), which corresponds to a more realistic ocean heating, coarsely mimicking otherwise neglected ocean dynamics, such as lateral advection, convection and diffusion. This is especially needed to improve low latitude oceanic forcing.
+
+- An analytic formulation can be written as:
+$$
+Q = Q_0(1-2\phi^2/w_\phi^2) \frac{exp(- (\phi^2/w_\phi^2))}{cos(\phi)}
+$$
+- where $\phi$ is latitude in radians, $Q_0$ is the amplitude of the equatorial heating and $w_\phi$ the width of the heating in radians. 
+
+# Alternatives
+
+## [Semtner 1976](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2021MS002671) 3-layer model 
+- Semtner (1976) developed a simple model for the evolution ice (2 layers) and snow (1 layer) temperatures, which is represented by a 1D diffusve process also accounting for radiation, melting and energy release from brine poskets and accumulating snow. 
+- Main formation of sea ice - energy fluxes from vertical boundaries and heat storage in brine pockets. The vertical processes retain the central role.  
+- Semtner 1976 found that the 0-layer model is broadly comparable in terms of accuracy as the 3-layer model
+
+## Ocean-based dynamical sea-ice model
+- to be coordinated after AMIP
+
+
--- a/experiments/ClimaCore/sea-ice/slab_ice.jl
+++ b/experiments/ClimaCore/sea-ice/slab_ice.jl
@@ -1,0 +1,341 @@
+# # Heat Equation + Slab Tutorial
+
+#  - load external packages:
+import LinearAlgebra, UnPack
+import ClimaCore: Fields, Domains, Topologies, Meshes, DataLayouts, Operators, Geometry, Spaces
+
+using Base: show_supertypes
+using OrdinaryDiffEq: ODEProblem, solve, SSPRK33
+
+using Logging: global_logger
+using TerminalLoggers: TerminalLogger
+
+using RecursiveArrayTools
+
+using OrdinaryDiffEq
+
+using Statistics
+
+#src ## Setup Logging Information
+global_logger(TerminalLogger())
+const CI = !isnothing(get(ENV, "CI", nothing))
+
+# ## Define Parameters
+#  - Global Constants
+const FT = Float64;
+
+#  - Experiment-specific Parameters
+parameters = (
+    ## atmos parameters
+    zmin_atm = FT(0.0), # height of atm stack bottom [m] 
+    zmax_atm = FT(1.0), # height of atm stack top [m]
+    n = 15,  # number of elements in atm stack 
+    μ = FT(0.0001), # diffusion coefficient [m^2 / s]
+    T_top = FT(280.0), # fixed temperature at the top of the domain_atm [K]
+    ## slab ice parameters
+    ρc_ml = FT(4e6), # density times heat transfer coefficient for mixed layer [J / m2 / K ]
+    F0_base = FT(120), # ice base transfer coefficient [W / m2 / K]
+    T_base = FT(273.16), # ice base temperature [K] 
+    L_ice = FT(3e8), # latent heat coefficient for ice [J / m3]
+    h_ml = FT(1), # mixed layer depth [m] 
+    T_freeze = FT(273.16), # temperature at freezing point [K] 
+    k_ice = FT(2), # thermal conductivity of ice [W / m / K]
+    ## coupling parameters 
+    λ = FT(0.001), # coupling coefficient (J / K / s / m2)
+)
+
+# ## Define Model Functions
+
+# - Model 1 (atm) Equations
+"""
+    ∑tendencies_atm!(du, u, (parameters, T_sfc), t)
+
+Heat diffusion equation
+    dT/dt =  ∇ μ ∇ T
+    where
+        (- μ ∇ T) = 0           at z = zmax_atm
+        (- μ ∇ T) = F_sfc       at z = zmin_atm
+
+We also use this model to calculate and accumulate the downward surface fluxes, F_sfc:
+    F_sfc = - λ * (T_sfc - T1) 
+    d(F_integrated)/dt  = F_sfc
+    where
+        F_integrated is reset to 0 at the beginning of each coupling cycle
+        T1 = atm temperature near the surface (here assumed equal to the first model level)
+"""
+function ∑tendencies_atm!(du, u, p, t)
+
+    FT = eltype(u)
+    T = u.T # u.x = vector of prognostic variables from DifferentialEquations
+    F_sfc = calculate_flux(p.Ya.T_sfc[1], parent(T)[1], p.p)
+
+    ## set BCs
+
+    bcs_bottom = Operators.SetValue(Geometry.WVector(F_sfc ./ p.p.λ))
+    bcs_top = Operators.SetValue(Geometry.WVector(FT(0)))
+
+    gradc2f = Operators.GradientC2F()
+    gradf2c = Operators.DivergenceF2C(bottom = bcs_bottom, top = bcs_top) # # Neumann BC (face-to-center)
+    ## tendency calculations
+    @. du.T = gradf2c(p.p.μ * gradc2f(T)) # dT/dt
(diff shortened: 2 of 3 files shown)


experiments/ClimaCore/sea-ice/README.md
```diff
@@ -0,0 +1,71 @@
+# **Simple Sea Ice Model**
+
+# Semtner's Zero-layer Model
+
+Sea ice is approximated as a 0-layer slab. This means that the model does not occupy a particular gridpoint between the mixed layer (ocean) and the atmosphere. It is essentially a column-wise ODE, with no horizontal transport between the columns. In the absence of ice, the sea-ice model reduces to the slab ocean formulation. The ice is assumed to have a negligible heat capacity, so there is no heat storage within it. The only store of energy comes from the ice thickness.
```
kmdeck, line 5: This is a little unclear to me - "The ice is assumed to have a negligible heat capacity, so there is no heat storage within it. The only store of energy comes from the ice thickness."  I guess we mean that the temperature of the ice is fixed at the freezing point (c dT_ice /dt = 0 b/c c -> 0) but that amount of ice can change, which changes the energy stored within the ice.

experiments/ClimaCore/sea-ice/README.md
```diff
@@ -0,0 +1,71 @@
...
+$$
+- with the latent heat of fusion, $L_i=3 \times 10^8$ J / m3, and where the (upward-pointing) flux into the atmosphere is
+$$
+F_{atm} = F_{rad} + F_{SH} + F_{LH}
+$$
+- $F_{atm}$ is obtained from the atmospheric model (via the coupler). The flux at the ice base into the mixed layer is
+$$
+F_{base} = F_0(T_{ml} - T_{melt})
+$$
+- where $T_{melt} = 273.16$ K is the freezing temperature, which sets the temperature of the entire isothermal ice slab, and the basal heat coefficient $F_0 = 120$ W /m2 / K.
```
kmdeck, line 21: what sets the temp of the slab? Isnt the slab set to T_freeze?

experiments/ClimaCore/sea-ice/README.md
```diff
@@ -0,0 +1,71 @@
...
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{atm}
+$$
+- while ice-covered conditions require that:
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{base}
+$$
+
+## 3. Transition from no ice to ice
+- When the updated $T_{ml}^{t+1} < T_{melt}$, set $T_{ml}^{t+1} = T_{melt}$ and grow ice via $h_i$ due to the corresponding energy deficit.
```
kmdeck, line 34: Does this mean that when ice is present, T_ml quickly becomes T_freeze (due to F_base) and then remains there (F_base = 0)?

experiments/ClimaCore/sea-ice/README.md
```diff
@@ -0,0 +1,71 @@
...
+- where $T_{melt} = 273.16$ K is the freezing temperature, which sets the temperature of the entire isothermal ice slab, and the basal heat coefficient $F_0 = 120$ W /m2 / K. 
+
+## 2. Ocean mixed layer temperature, $T_{ml}$
+- $T_{ml}$ is the standard slab ocean formulation in ice-free conditions:
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{atm}
+$$
+- while ice-covered conditions require that:
+$$
+\rho_w c_w h_{ml}\frac{dT_{ml}}{dt} = - F_{base}
```
kmdeck, line 30: should be +F_base, right? If it is -Fbase in ice equation

experiments/ClimaCore/sea-ice/README.md
```diff
@@ -0,0 +1,71 @@
...
+$$
+
+## 3. Transition from no ice to ice
+- When the updated $T_{ml}^{t+1} < T_{melt}$, set $T_{ml}^{t+1} = T_{melt}$ and grow ice via $h_i$ due to the corresponding energy deficit.
+- When the updated $h_i^{t+1} <= 0$ from a non-zero $h_i^t$, adjust $h_i^{t+1} = 0$ and use the surplus energy to warm the mixed layer.
+
+## 4. Surface temperature ($T_s$)
```
kmdeck, line 37: If there is ice, T_s = surface temp of ice? Does that contradict the statement earlier that the ice is an isothermal slab? It seems like we have ice with surf temp = T_s, bottom temp  = T_freeze, and flux at surface =  F_i = k_i \frac{T_{melt} - T_s}{h_i}, flux at bottom = F_base, but we assume F_atm = F_i.

When the surface is ice free, is T_s = T_ml? Should mention that here?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+
+# - Model 1 (atm) Equations
+"""
+    ∑tendencies_atm!(du, u, (parameters, T_sfc), t)
+
+Heat diffusion equation
+    dT/dt =  ∇ μ ∇ T
+    where
+        T  = 280 K              at z = zmax_atm
+        dT/dt = - ∇ F_sfc       at z = zmin_atm
```
kmdeck, line 57: Is this a lower BC? I'm confused that it seems to be setting dT/dt. Do we want -\mu \nabla T = F_sfc at the bottom?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+"""
+function ∑tendencies_atm!(du, u, p, t)
+
+    FT = eltype(u)
+    T = u.T # u.x = vector of prognostic variables from DifferentialEquations
+    F_sfc = calculate_flux(p.Ya.T_sfc[1], parent(T)[1], p.p)
+
+    ## set BCs
+    bcs_bottom = Operators.SetValue(Geometry.WVector(F_sfc ./ p.p.λ))
+    bcs_top = Operators.SetValue(Geometry.WVector(FT(0)))
```
kmdeck, line 74: Does not match BC in comments (T = 280)?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+    ∑tendencies_atm!(du, u, (parameters, T_sfc), t)
+
+Heat diffusion equation
+    dT/dt =  ∇ μ ∇ T
+    where
+        T  = 280 K              at z = zmax_atm
+        dT/dt = - ∇ F_sfc       at z = zmin_atm
+
+We also use this model to calculate and accumulate the downward surface fluxes, F_sfc:
+    F_sfc = - λ * (T_sfc - T1)
```
kmdeck, line 60: Im a little confused why F_sfc is not F_rad+SHF + LHF here.

kmdeck: ah OK, I see that this is a to-do item, and that F_sfc is consistent with the ice model (but not as described in the docs for the slab ice model). Maybe you could note in the documentation that as a stand-in for F_sfc, you use this expression?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+            @. ΔT_base = p.T_freeze - T_sfc
+        end
+        # surface is ice-covered, so update T_sfc as ice surface temperature
+        @. Y.T_sfc += ΔT_base
+    else # ice-free, so update T_sfc as mixed layer temperature
+        @. Y.T_sfc = T_ml + ΔT_ml
+    end
+
+    # update state
+    @. Y.T_ml += ΔT_ml
```
kmdeck, line 155: Does this work? I thought you could only change prognostic variables via callbacks (or at least, that it was only recommended to do so). 

Would it be possibly to implement the ice model as a diff eq now, rather than implement this kind of funky "set Y by timestepping by hand, wrap inside a ODE problem which doesnt ever change dY from 0" (also, is dY *always* 0 when initialized? perhaps you should still set dY .= zero field inside the ice tendencies stub)

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+    # solve for T_sfc
+    if (h_ice[1] + Δh_ice[1] > 0) #  surface is ice-covered
+        # if ice covered, solve implicity (for now one Newton iteration: ΔT_s = - F(T_s) / dF(T_s)/dT_s )
+        @. F_ice = p.k_ice / (h_ice + Δh_ice) * (p.T_base - T_sfc)
+        @. ΔT_base = (-F_atm + F_ice) / (p.k_ice / (h_ice + Δh_ice) + ∂F_atm∂T_sfc)
+        if (T_sfc[1] + ΔT_base[1] > p.T_freeze)
+            @. ΔT_base = p.T_freeze - T_sfc
+        end
+        # surface is ice-covered, so update T_sfc as ice surface temperature
+        @. Y.T_sfc += ΔT_base
```
kmdeck, line 149: another item to keep in mind (we need to do so on the land side too) is that ice/ocean and snow/soil have different albedos and latent heats involved (when computing LHF). They may also have different roughness lengths.

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+
+    ∂F_atm∂T_sfc = get_∂F_atm∂T_sfc(p) # this will be passed from atmos/SF.jl
+
+    ΔT_ml = similar(F_atm)
+    Δh_ice = similar(F_atm)
+    F_ice = similar(F_atm)
+    ΔT_base = similar(F_atm)
+
+    # ice thickness and mixed layer temperature changes due to atmosphereic and ocean fluxes 
+    if h_ice[1] <= 0 # ice-free
```
kmdeck, line 120: Im not sure if this really matters, but Im not sure how often something is =0. You could say h_ice <eps(FT).

Do you see h_ice as negative ever?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+parameters = (
+    ## atmos parameters
+    zmin_atm = FT(0.0), # height of atm stack bottom [m] 
+    zmax_atm = FT(1.0), # height of atm stack top [m]
+    n = 15,  # number of elements in atm stack 
+    μ = FT(0.0001), # diffusion coefficient [m^2 / s]
+    T_top = FT(280.0), # fixed temperature at the top of the domain_atm [K]
+    ## slab ice parameters
+    ρc_ml = FT(4e6), # for mixed layer
+    F_basal = FT(120), # W / m2 / K
```
kmdeck, line 37: this is minor, but F_basal for the coefficient and F_base for the flux is a little confusing. maybe even a letter besides "F" for the coefficient would be helpful, since "F" denotes a flux everywhere else?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+    if h_ice[1] <= 0 # ice-free 
+        @. ΔT_ml = -(F_atm + ocean_qflux) * Δt / (p.h_ml * p.ρc_ml)
+        @. Δh_ice = 0
+    else # ice-covered
+        @. ΔT_ml = -(p.F_basal * (T_ml - p.T_base) + ocean_qflux) * Δt / (p.h_ml * p.ρc_ml)
+        @. Δh_ice = (F_atm - p.F_basal * (T_ml - p.T_base)) * Δt / p.L_ice
+    end
+
+    # adjust if transition to ice-covered 
+    if (T_ml[1] + ΔT_ml[1] < p.T_freeze)
```
kmdeck, line 129: If h_ice > 0, I assume this can still be satisfied. That might be what you want, in any case! To keep T_ml at T_freeze and creat more ice. But thought Id point it out.

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+        @. Δh_ice = Δh_ice - (T_ml + ΔT_ml - p.T_freeze) * (p.h_ml * p.ρc_ml) / p.L_ice
+        @. ΔT_ml = p.T_freeze - T_ml
+    end
+
+    # adjust if transition to ice-free
+    if ((h_ice[1] > 0) & (h_ice[1] + Δh_ice[1] <= 0))
+        @. ΔT_ml = ΔT_ml - (h_ice + Δh_ice) * p.L_ice / (p.h_ml * p.ρc_ml)
+        @. Δh_ice = -h_ice
```
kmdeck, line 137: h_ice[1]?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+    # update state
+    @. Y.T_ml += ΔT_ml
+    @. Y.h_ice += Δh_ice
+
+    @. Ya.ice_mask = h_ice[1] > 0 ? h_ice[1] * FT(1) : h_ice[1] * FT(0)
+
+    return
+end
+
+function ∑tendencies_ice_stub(du, u, p, t)
```
kmdeck, line 163: Are you *required* to use forward Euler with this model, to handle the ice <-> no ice transitions?

experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -0,0 +1,341 @@
...
+    dY = du
+    FT = eltype(dY)
+
+    solve_ice!((; u = u, p = p), p.Δt) # timestepping outside of DeffEq (which is only used for logging here)
```
kmdeck, line 167: If you cant get around this, is it preferable to not make the ice variables prognostic (needing dY...) and instead just make them aux vars that get updated?

experiments/ClimaCore/sea-ice/slab_ice.jl
kmdeck, line 149: the += here could be an issue. What if in the previous step, Y.T_sfc = mixed layer temp? In this case, would this give you the right ice surface temp? It seems safer to use Y.T_sfc = ice surface temp.

kmdeck (commented):
Hi Lenka. This generally seems fine, though the documentation is ahead of where the code is (in terms of F_sfc, and in terms of writing the ice/ocean equations as continuous ODEs rather than discretized equations). 

Right now, it's a hard-coded algorithm for updating T_sfc, T_ml, h_ice using foward Euler stepping. If it will stay this way, I think it might be better to remove these from Y and stick them in p, because the code isnt actually timestepping them with ODE.jl/making use of the `prob` and keywords a user might set...

a couple items flagged that seemed like they could lead to issues - modifying the state directly, using a += for the ice covered T_sfc, and not setting dY = 0 each step (since you are modifying them without using dY - if it is nonzero at initialization, would ODE algorithm update Y?)

experiments/ClimaCore/sea-ice/README.md
LenkaNovak: I guess this is just saying that the ice layer doesn't have a prognostic equation for T. I will rephrase.

experiments/ClimaCore/sea-ice/README.md
LenkaNovak: It does :)

experiments/ClimaCore/sea-ice/README.md
LenkaNovak: Sorry, which ice equation do you mean? All fluxes should be positive when pointing upward, following the original paper.

LenkaNovak: Ahh, you mean `h_ice`? The sign should be the same - e.g. an upward (positive) flux will decrease ocean temperature (`T_ml`) as well as decreasing the ice thickness (`h_ice`). Lmk if this is not what you meant. 🤔

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: Reworded.

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: Yeah, it works lol. I did it this way mainly to deal with the time stepping and to be able to access the ice state from previous time steps for diagnostics. We can do something more sophisticated in the next iteration, but it may not be straightforward since you need to do stuff between time integrating `h_ice` and `dT_sfc`.

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: It shouldn't be negative and the `@. Δh_ice = -h_ice` should take care of this, but since it is not an integer I left it there in case there were some FP errors. I can invert the argument and change it to `h_ice[1] > 0` for ice covered and `else` otherwise?

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: Yeah, I agree. Great point!

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: Umm, `h_ice > 0`  wouldn't indicate a transition, so it would loop over the entire ice-covered domain, rather than just where it's needed. Thanks for pointing it out though!

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: It doesn't need it here. Once we can deal with spatially varying BCs in ClimaCore, this will be point wise, so all the index brackets should go away. :)

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: I'm still using DiffEq to save the solution with time. Not sure it's easily possible otherwise.

experiments/ClimaCore/sea-ice/slab_ice.jl
LenkaNovak: This method should account for this. The` ice surface temp = Y.T_sfc +  ΔT_base` (btw the name `ΔT_base` was ambiguous so I changed it).

LenkaNovak:
Thanks so much for all the above comments, @kmdeck. The docs are a little ahead in view of the next stage of this issue (#42 ), but you're right, it is clearer to say that explicitly. I tried to note where we're using simplifications for the time being (e.g. `F_sfc`). The timestepping I agree is a little clunky, but I wasn't sure how else to implement the weird implicit time stepping where` T_sfc^{t+1} = f(h_ice^{t+1},  T_ml^{t+1})`. Since it seems to be doing the right thing, are you happy with merging this, so we have something to play around with in the GCM and then come up with something more optimal in the next stage of #42 ? And no problem, we can make sure that `dY` is always zero! :)

kmdeck:
> Thanks so much for all the above comments, @kmdeck. The docs are a little ahead in view of the next stage of this issue (#42 ), but you're right, it is clearer to say that explicitly. I tried to note where we're using simplifications for the time being (e.g. `F_sfc`). The timestepping I agree is a little clunky, but I wasn't sure how else to implement the weird implicit time stepping where` T_sfc^{t+1} = f(h_ice^{t+1}, T_ml^{t+1})`. Since it seems to be doing the right thing, are you happy with merging this, so we have something to play around with in the GCM and then come up with something more optimal in the next stage of #42 ? And no problem, we can make sure that `dY` is always zero! :)

hi @LenkaNovak! Yes, feel free to merge! It looks like it is working based on your checks. And I'm not sure how to better handle the timestepping either - we'll face the same issue with snow on land. It makes sense to do what works now, and we can address it later, as you say.

PR #50, opened 2022-05-19
Smooth MPI Simple Earth

LenkaNovak:
# MPI Simple Earth without Land Hydrology

## Co-authors
Akshay Sridhar <akshaysridhar@users.noreply.github.com>
Ben Mackay <jbmackay@caltech.edu>
Kat <kmdeck@users.noreply.github.com>
jiahe23 <jiahe23@users.noreply.github.com>
Zhaoyi Shen <szy21@users.noreply.github.com>

## Purpose
- uses two slabs with different properties (e.g. depth) and the full atmos except turbulence (i.e. 0-moment microphysics, gray radiation using `RRRMGP.jl`, diffusion, aerodynamic formulae for all variables using `SurfaceFluxes.jl`) and stabilized with hyperdiffusion
- NB: clipping of negative water was applied, with negligible effects on conservation (should be removed with limiters)

## New features
- adds features to #39 by adding
    - [x] RRTMGP usage and flux passing  
    - [x] RRTMGP T_sfc
    - [x] land sea mask
    - [x] prescribed SSTs 
    - [x] prescribed sea ice
    - [x] radiation conservation checks 
    - albedo passing (to follow ClimaAtmos) - to be addressed in #51 

## Checks for full atmos + 2 slabs
- conservation (with gray radiation) of domain-integrated energy. E.g. spinup over the first 1e4 seconds:

Screen Shot 2022-06-07 at 9 41 59 AM

long_run

long_run_error

- physical (stable 60d runs)
    - T_sfc:
 
https://user-images.githubusercontent.com/38083334/169573166-54231b68-7633-4262-b92c-db63d66a38bd.mp4

    - u at 7km

https://user-images.githubusercontent.com/38083334/169573347-4bb0b479-896c-4805-a2d7-610c2d931bed.mp4

    - but qt reduces dramatically (needs more investigation) - e.g. at 1km:

https://user-images.githubusercontent.com/38083334/169573742-4d810e10-72fc-444c-bea0-9ceb7e298593.mp4

## TODO
- [x] check  qt reduction (that was only spin-up)
- [x] check non-conservation with radiation
- [x] check sea ice quantitatively
- [ ] clean-up 

## Next PRs
- add the existing sea ice and hydrology column models
  - sea ice (building on #42 )
  - ClimaLSM bucket (building on #41 )

PR #51, opened 2022-05-30
Earth bucket

kmdeck:
This PR adds in a test case showing a coupled atmos, land, ocean, and sea ice model. The bucket model in place of the slab land model, which is how it differs from https://github.com/CliMA/ClimaCoupler.jl/pull/50.

We've also set it up so that the specific humidity at the surface uses the correct land value, and set the RRTMGP fields for surface temp and albedo correctly based on the surface values.

A future PR will update the version of ClimaAtmos we are using so that we can read the precipitation from the atmos cache, and pass it to the land model, and so that we are using topography in the atmos model.

A future PR will also implement a coupler interface as this interface is currently hard coded to some extent to the models being used here.

co-authors:
Ben Mackay @jb-mackay 
Lenka Novak @LenkaNovak 
Zhao-yi Shen @szy21


kmdeck:
Currently setting the mask= 1 everywhere to check conservation in a simpler model, but the ocean and sea ice models are still running (without error).
There may be some issues with roundoff. 

conservation

kmdeck:
Things to do:
- clean up conservation plot code, add water to it (note - \rho q in atmos is clipped at zero so we may not see conservation)
- get precipitation from atmos (change coupler driver code and atmos)
- pass albedo to radiation code (change to LSM code, radiation code, what coupler driver does)
- get realistic d_soil parameter
- figure out where FT and other parsed args are actually set, kind of confusing

experiments/AMIP/moist_mpi_earth/coupler_driver.jl
```diff
@@ -0,0 +1,272 @@
...
+
+# At this stage, the integrators all have dY(0) computed based on stuff stored in aux, Y0, etc. - dE is up to date
+# Then we need to update aux to t1, because in the next step!, Y(1) will be computed based on Y0 and dY(0), and then
+# dY(1) will be computed using aux(1), Y(1), t(1)
+# But how to update aux to t(1) without Y(1)??
+@show "Starting coupling loop"
+walltime = @elapsed for t in (tspan[1]+Δt_cpl:Δt_cpl:tspan[end])
+    @show t
+    ## Atmos
+    atmos_pull!(atmos_sim, slab_ice_sim, bucket_sim, slab_ocean_sim, mask, boundary_space, prescribed_sst, z0m_S,  z0b_S, T_S, ocean_params, SST)
```
kmdeck, line 210: thought that this line may need to be moved to *after* step!. step! does not update atmos_sim.integrator.p, atmos_pull! does. but we have already set atmos_sim.integrator.p for the first iteration in the atmos_pull!/reinit period above, before this loop starts?

this step is not computing anything new on the first iteration, I dont think.

kmdeck: on the first step, the fluxes in line 218 are unchanged from the fluxes we computed up front.

PR #52, opened 2022-06-01
Strip coupler interface to basics

jb-mackay:
This removes extraneous functionality (units, date checking) from the coupler interface beyond putting and getting data from a coupler storage object. This is the starting point for coming development.

### DIFF
--- a/src/CouplerState/coupler_state.jl
+++ b/src/CouplerState/coupler_state.jl
@@ -1,28 +1,19 @@
-using Unitful, Dates
 using PrettyTables
 
 export CouplerState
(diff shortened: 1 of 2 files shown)


jb-mackay:
@LenkaNovak will merge after formatting - next step will probably introduce the "ClimaSimulations-like" abstractions

christophernhill:
@jb-mackay @LenkaNovak it might be good to leave some of this optional. What can happen when you don't have this sort of stuff available is something gets confused and then you end up mapping the wrong fields to each other.  It can be easier tom unravel that if you have the option to tag things in various ways.

PR #53, opened 2022-06-02
Add basic simulations wrappers

jb-mackay:
Adds simulations wrappers and implements them in the sea breeze experiment.

The Simulation abstraction standardizes how models are run. This is important for two reasons:
1. Standalone models all are built and run in the same way, regardless of component (Atmos, Land, Ocean, etc...) or backend (ClimaCore, Oceananigans). This greatly boosts usability and accessibility.
2. The coupler expects standardized behavior from simulations. This is necessary when writing coupled timesteppers that are readable (using `step!`, and for writing coupling methods (`put, get, push, pull`, etc.)

### DIFF
--- a/experiments/ClimaCore/sea_breeze/atmos_rhs.jl
+++ b/experiments/ClimaCore/sea_breeze/atmos_rhs.jl
@@ -101,8 +101,7 @@ function atm_rhs!(dY, Y, params, t)
     dYc = dY.Yc
     dρw = dY.ρw
 
-    center_coords = Fields.coordinate_field(params.domain.hv_center_space)
-    face_coords = Fields.coordinate_field(params.domain.hv_face_space)
+    center_coords = Fields.coordinate_field(axes(Yc))
 
     # spectral horizontal operators
     hdiv = Operators.Divergence()
@@ -211,6 +210,27 @@ function atm_rhs!(dY, Y, params, t)
     Spaces.weighted_dss!(dYc)
     Spaces.weighted_dss!(dρw)
     return dY
+end
+
+# Atmos Simulation - later to live in ClimaAtmos
+struct AtmosSimulation <: ClimaCoupler.AbstractAtmosSimulation
+    integrator::Any
+end
+
+function AtmosSimulation(Y_init, t_start, dt, t_end, timestepper, p, saveat, callbacks = CallbackSet())
+    atm_prob = ODEProblem(atm_rhs!, Y_init, (t_start, t_end), p)
+
+    atm_integ = init(
+        atm_prob,
+        timestepper,
+        dt = dt,
+        saveat = saveat,
+        progress = true,
+        progress_message = (dt, u, params, t) -> t,
+        callback = callbacks,
+    )
+
+    return AtmosSimulation(atm_integ)
 end
 
 # init simulation
--- a/experiments/ClimaCore/sea_breeze/ocean_rhs.jl
+++ b/experiments/ClimaCore/sea_breeze/ocean_rhs.jl
@@ -27,6 +27,17 @@ function hspace_1D(xlim = (-π, π), npoly = 0, helem = 10)
     return space
 end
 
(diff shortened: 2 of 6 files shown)


src/CoupledSimulations/coupled_simulation.jl
```diff
@@ -1,11 +1,38 @@
...
-    CoupledSimulation
+    AbstractSimulation
 
-An abstract type representing a coupled simulation.
+An abstract type representing a model simulation.
 """
-abstract type CoupledSimulation end
+abstract type AbstractSimulation end
+
+abstract type AbstractAtmosSimulation  <: AbstractSimulation end
```
jb-mackay, line 10: It may not be necessary to have abstract subtypes of `AbstractSimulation`, but we'll see.

experiments/ClimaCore/sea_breeze/ocean_rhs.jl
```diff
@@ -27,6 +27,17 @@ function hspace_1D(xlim = (-π, π), npoly = 0, helem = 10)
     return space
 end
 
+# Ocean Simulation - Later to live in Oceananigans
+struct OceanSimulation <: ClimaCoupler.AbstractOceanSimulation
```
jb-mackay, line 31: When we use Oceananigans, this will become instantly more important. The simulation wrapper standardizes what the coupler sees and expects from models. A custom `step!` method for an oceananigans simulation would be required.

experiments/ClimaCore/sea_breeze/run.jl
```diff
@@ -61,115 +83,96 @@ ocn_Y_default, ocn_domain = ocn_init(xmin = -500, xmax = 0, helem = 10, npoly =
...
+lnd_Y = Fields.FieldVector(T_sfc = lnd_Y_default.T_sfc)
+lnd_p = (cpl_parameters, F_sfc = lnd_F_sfc)
+land = LandSimulation(lnd_Y, t_start, cpl_Δt / lnd_nsteps, t_end, SSPRK33(), lnd_p, saveat)
+
+# coupled simulation
+struct AOLCoupledSimulation{FT, A, O, L} <: ClimaCoupler.AbstractCoupledSimulation
+    atmos::A
+    ocean::O
+    land::L
+    Δt::FT
```
LenkaNovak, line 118: I wonder if `Δt` should live in the `CoupledSimulation`. Once we start running the AOL models at different time steps, their respective `Δt` will live in their integrator.

experiments/ClimaCore/sea_breeze/atmos_rhs.jl
```diff
@@ -213,6 +212,27 @@ function atm_rhs!(dY, Y, params, t)
     return dY
 end
 
+# Atmos Simulation - later to live in ClimaAtmos
+struct AtmosSimulation <: ClimaCoupler.AbstractAtmosSimulation
+    integrator::Any
```
LenkaNovak, line 217: I think access to `Y_init` is useful for debugging. Actually is `Y_init = integrator.sol.u[1]` with how we currently use the `step!` function?

LenkaNovak (approved):
Looks great! I just left a couple of comments.

experiments/ClimaCore/sea_breeze/run.jl
jb-mackay: This is the coupled time step; as you note, each individual simulation has its own time step that will live in the integrators. I can add a in-code comment to clarify this.

experiments/ClimaCore/sea_breeze/atmos_rhs.jl
jb-mackay: > Actually is `Y_init = integrator.sol.u[1]` with how we currently use the `step!` function?

I think so - as long as we save the solution after the init then we would have access to this. We could make some helper functions to grab the initial conditions though!

christophernhill (approved):
@jb-mackay and @LenkaNovak this is super exciting!

PR #54, opened 2022-06-02
Remove Manifest and update gitignore.

jb-mackay:


.gitignore
```diff
@@ -20,13 +21,15 @@ TAGS
...
 # Docs
-docs/Manifest.toml
 docs/build/
 docs/src/generated/
 !docs/src/assets/*.png
 !docs/src/assets/*.svg
 !docs/src/images/*.png
 
+# Experiments
+!experiments/AMIP/*/Manifest.toml
```
jb-mackay, line 31: @LenkaNovak if there are experiments that require a manifest you can add a similar line to the .gitignore.

Project.toml
```diff
@@ -1,12 +1,7 @@
 name = "ClimaCoupler"
 uuid = "4ade58fe-a8da-486c-bd89-46df092ec0c7"
-authors = ["jbmackay <jbmackay@caltech.edu>"]
+authors = ["CliMA Contributors <clima-software@caltech.edu>"]
 version = "0.1.0"
 
 [deps]
-ClimaCommsMPI = "5f86816e-8b66-43b2-912e-75384f99de49"
-Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
-OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
jb-mackay, line 9: Similarly, only packages that are required for code in `src` should live in the Project.toml. Packages like SciMLBase and OrdinaryDiffEq are just needed for running models and can live in the Project.toml files in `experiments/`

LenkaNovak: Sounds good!

Issue #44, opened 2022-04-15
Coupler-specific Interface


LenkaNovak:
# Purpose

A formal coupler-specific interface is missing. We need to re-introduce concepts from `CouplerMachine.jl` to formalize the coupling and information passing between different models for the `ClimaCore`-related codebases. 

# Final Key Result
A lightweight coupler interface that users feel comfortable using when building coupled models going forward.
- simplified, readable driver files
- automatic remapping
- mock simulations interface

# Sketch of Needed PRs

- [x] Reset current interface to basics "blank canvas" #52
- [x] Introduce simple simulations interface (wrapping integrators) #53 
- [ ] Auto-building remap operators #67
- [x] get with remapping #58 #76 
- [ ] put with remapping #65
- [x] Implement push and pull methods #83 
- [ ] Coupled timestepper standardization
- [ ] Diagnostic/utility standardization
- [ ] `coupler_utils` to src #134 
- [ ] streamline slab types

# Driver PRs
- Use sea-breeze as testbed and example of new interface

# Current Design
- `ComponentModelSet`: struct (or Tuple), listing all required model components
```
struct ComponentModelSet{M} <: AbstractComponentModelSet 
    models::M
end
```

- `CouplerSimulation`: coupler and component model simulation info. `coupler_fields ` is optional and may contain any relevant fields or their subsets (e.f. boundary fields)
```
struct CouplerSimulation{CS, CF, RW, S, L} <: AbstractSimulation
    coupler_solver::CS
    coupler_fields::CF
    regrid_weights::RW
    simulations::S
    logger::L
end
```

- `CouplerSolver`: time stepping information, unknown by the model components. Current implementation `stepping = Leapfrog()` and `mode = Sequential()`
```
struct CouplerSolver{T, TS, S, M} <: AbstractSolver
    dt::T
    timespan::TS
    stepping::S
    mode::M
end
```

- `init`: iterate over all component models' init functions, with externally customizable ICs, BCs, dt (requirement for ClimaAtmos, ClimaSLM, Oceananigans)
```
function init(cm:: ComponentModelSet)
    # iterate over all component models' init functions
        # spatial info (domain / spaces / models, etc)
        # specify BCs
        # specify ICs
        # init integrators
    # check compatibility
        # dt for each model not to exceed CFL
        # unit checks
    # regridding init (if applicable)
        # generate all regridding weight maps
    #@showinfo 
    return CouplerSimulation
end
```

[...]

PR #58, opened 2022-06-15
Add remapping to coupler puts & gets

jb-mackay:
## Purpose and Content
Adds remapping to the coupler `put!` and `get` calls. Implements these calls in the sea breeze simulation

## Benefits and Risks
Hides explicitly calling remapping operations from the user.

## Next steps
The initialization phase needs reworking and reordering in order to facilitate the auto-construction of the remap operators. 

## Linked Issues
#44 

## PR Checklist

### DIFF
--- a/experiments/ClimaCore/sea_breeze/run.jl
+++ b/experiments/ClimaCore/sea_breeze/run.jl
@@ -94,7 +94,9 @@ maps = (
 )
 
 # initialize coupling fields
-atm_T_sfc = Operators.remap(maps.ocean_to_atmos, ocn_Y_default.T_sfc) .+ Operators.remap(maps.land_to_atmos, lnd_Y_default.T_sfc) # masked arrays; regrid to atm grid
+atm_T_sfc =
+    Operators.remap(maps.ocean_to_atmos, ocn_Y_default.T_sfc) .+
+    Operators.remap(maps.land_to_atmos, lnd_Y_default.T_sfc) # masked arrays; regrid to atm grid
 atm_F_sfc = Fields.zeros(atm_boundary)
 ocn_F_sfc = Fields.zeros(ocn_domain)
 lnd_F_sfc = Fields.zeros(lnd_domain)
@@ -113,8 +115,13 @@ lnd_p = (cpl_parameters, F_sfc = lnd_F_sfc)
 land = LandSimulation(lnd_Y, t_start, cpl_Δt / lnd_nsteps, t_end, SSPRK33(), lnd_p, saveat)
 
 # coupled simulation
-struct AOLCoupledSimulation{FT, A <: AtmosSimulation, O <: OceanSimulation, L <: LandSimulation, C <: ClimaCoupler.CouplerState} <:
-       ClimaCoupler.AbstractCoupledSimulation
+struct AOLCoupledSimulation{
+    FT,
+    A <: AtmosSimulation,
+    O <: OceanSimulation,
+    L <: LandSimulation,
+    C <: ClimaCoupler.CouplerState,
+} <: ClimaCoupler.AbstractCoupledSimulation
     # Atmosphere Simulation
     atmos::A
     # Ocean Simulation
@@ -136,7 +143,7 @@ for (name, map) in pairs(maps)
     coupler_add_map!(coupler, name, map)
 end
 
-sim = AOLCoupledSimulation(atmos, ocean, land,coupler, cpl_Δt)
+sim = AOLCoupledSimulation(atmos, ocean, land, coupler, cpl_Δt)
 
 # step for sims built on OrdinaryDiffEq
 function step!(sim::ClimaCoupler.AbstractSimulation, t_stop)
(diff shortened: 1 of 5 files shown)


jb-mackay:
Currently it is not possible to map to an exchange or intermediary grid -- any field `put!` in the coupler lives on the space of the simulation that puts it there.

jb-mackay:
Currently it is not possible to map to an exchange or intermediary grid -- any field `put!` in the coupler lives on the space of the simulation that puts it there.

jb-mackay:
> Currently it is not possible to map to an exchange or intermediary grid -- any field `put!` in the coupler lives on the space of the simulation that puts it there.

This is because of the current way the mapping is set up which uses a tag (`write_sim`) to specify the source space/sim when remapping. This tag also (as the name suggests) indicates which simulation has write permissions to that coupled field. 

Perhaps the right solution is to remove the double usage of `write_sim` and add another tag like "`space_name`" which would be part of the remap operator name.

jb-mackay:
This PR sets up several follow up PRs:
1. Documentation updates:
    i. Change example used to the sea breeze as this is checked by buildkite and now a cleaner example
    ii. Add DocStringExtensions to improve doc quality
2. Add a `coupler_get!` method that accepts a field that will be modified/filled by the `coupler_get` call
3. ClimaCore: introduce `AbstractRemap` abstract type, with concrete subtypes including `LinearRemap`, `IdentityRemap`, `RLLRemap`, etc.
4. Update `coupler_put!` to include a remap operation. This creates symmetry and more flexibility for what grid the coupling takes place on.
5. Auto-build remap operators during init phase

experiments/ClimaCore/sea_breeze/run.jl
```diff
@@ -86,13 +86,17 @@ lnd_Y_default, lnd_domain = lnd_init(xmin = 0, xmax = 500, helem = 10, npoly = 0
 # Build remapping operators
 atm_boundary = Spaces.level(atm_domain.hv_face_space, PlusHalf(0))
 
-atm_to_ocn = Operators.LinearRemap(ocn_domain, atm_boundary)
-atm_to_lnd = Operators.LinearRemap(lnd_domain, atm_boundary)
-ocn_to_atm = Operators.LinearRemap(atm_boundary, ocn_domain)
-lnd_to_atm = Operators.LinearRemap(atm_boundary, lnd_domain)
+maps = (
+    atmos_to_ocean = Operators.LinearRemap(ocn_domain, atm_boundary),
```
kmdeck, line 90: minor, but we should be consistent between using "atm" and "atmos", "land" and "lnd, etc.

experiments/ClimaCore/sea_breeze/run.jl
```diff
@@ -111,18 +115,35 @@ lnd_p = (cpl_parameters, F_sfc = lnd_F_sfc)
...
+    # Coupler storage
+    coupler::C
     # The coupled time step size
     Δt::FT
 end
-sim = AOLCoupledSimulation(atmos, ocean, land, cpl_Δt)
+
+# init coupler fields and maps
+coupler = CouplerState()
+coupler_add_field!(coupler, :T_sfc_ocean, ocean.integrator.u.T_sfc; write_sim = ocean)
```
LenkaNovak, line 139: Another way of asserting that the correct models write into the right vars could be via multiple dispatch, which would allow specifying the same name to one variable in different domains (e.g. `T_sfc(::atm_sim)`, `T_sfc(::ocn_sim)` etc). We can discuss this during the interface 🍐 next week.

test/CouplerState/cplstate_interface.jl
```diff
@@ -1,37 +1,77 @@
 using Test
 using Random
 using ClimaCoupler, Dates, Unitful
+using IntervalSets
+using ClimaCore: Domains, Meshes, Geometry, Topologies, Spaces, Fields, Operators
+
+struct SimulationA{T} <: ClimaCoupler.AbstractSimulation
+    data::T
+end
+ClimaCoupler.name(::SimulationA) = :simA
```
LenkaNovak, line 10: I wonder if we could get away with avoiding the (simA, simB, ...) assignment, e.g. using `Symbol`? Again, something we can address during the next iteration. Noted down to discuss.

LenkaNovak (approved):
This looks very neat, thanks @jb-mackay ! I added a couple of comments, both of which we can discuss when we do the interface revamp. I think this can be merged.

test/CouplerState/cplstate_interface.jl
jb-mackay: Can you elaborate?

experiments/ClimaCore/sea_breeze/run.jl
jb-mackay: I'm worried that that would lead to a large number of methods for these shared variables. I think it would be better to have fewer methods that can give us flexibility. Let's discuss further.

test/CouplerState/cplstate_interface.jl
LenkaNovak: I guess it would be good for this to work for an arbitrary number of sims, each of which has it's own user-defined tag. Maybe the solution will become clearer next week, when we focus on the interface! :)

Issue #41, opened 2022-04-04
Land implement bucket model in Land 


LenkaNovak:
## Purpose
- provide a stub for the land model (including ClimaLSM interface) for the first AMIP simulation.
- enable a simplified water store in the land and compare it to the currently implemented hydrology model in a SCM. This will ensure all flux passing interfaces are implemented for the more complex test cases.

## Final Key Result
- coupled RCE SCM: 
    - atmos (full=TC,rad,microphysics[,gravity waves]), 
    - land (slab / full soil(= no canopy)), hydrology (bucket / currently implemented soil hydrology) 

## Measure of success 
(literature comparison / conservation / performance benchmark)
- [x] conservation of tracers
- [x] comparable performance to the standalone run

## Risks
- none

## Design, Features and Dependencies
- [x] approved design [here](https://www.overleaf.com/project/625c5f7677155b314c98eb75), see also  [CLSM-pr#43](https://github.com/CliMA/ClimaLSM.jl/pull/43)
- [x] enable interfaces to modify fluxes externally - ClimaAtmos [#257](https://github.com/CliMA/ClimaAtmos.jl/pull/257)
- [x] enable interfaces to modify fluxes externally - ClimaLSM
- [x] implement simple bucket in ClimaLSM in a standalone setting  [CLSM-pr#50](https://github.com/CliMA/ClimaLSM.jl/pull/50/files/d139f5c216cb21603ec7d7a8926f2932c6f0f8a3..3216f607273270d4d9e2ff3d0d751150264d12b0) replaced by https://github.com/CliMA/ClimaCoupler.jl/pull/57

## Driver PR(s)
- [x] coupled bucket case #57 
- [ ] clean-up PR
  - introduce more coupler-specific interface abstractions https://github.com/CliMA/ClimaCoupler.jl/issues/44
  - substitute the latest version of ClimaAtmos interface (TC.jl, limiters, perf improvements, etc)
  - documentation update, incl. comparison to results in Semtner
  - (look for literature benchmark - e.g., CESM RCE case?)

## People
- Lead: @kmdeck 
- Collaborators: @jb-mackay, @LenkaNovak 
- Reviewers: @yairchn 

## Ideal delivery date
- June 28 2022 (without TC.jl)

tapios:
This is looking good to me. Thanks, @kmdeck! I left some comments in the Overleaf.

kmdeck:
[...]

PR #57, opened 2022-06-14
Bucket2.0

kmdeck:
This PR adds in a test case showing a coupled atmos, land, ocean, and sea ice model. The bucket model in place of the slab land model, which is how it differs from https://github.com/CliMA/ClimaCoupler.jl/pull/50.

We've also set it up so that the specific humidity at the surface uses the correct land value, and set the RRTMGP fields for surface temp and albedo correctly based on the surface values.

A future PR will update the version of ClimaAtmos we are using so that we can read the precipitation from the atmos cache, and pass it to the land model, and so that we are using topography in the atmos model.

A future PR will also implement a coupler interface as this interface is currently hard coded to some extent to the models being used here.

co-authors:
Ben Mackay @jb-mackay
Lenka Novak @LenkaNovak
Zhao-yi Shen @szy21
Charlie @charleskawczynski

PR #69, opened 2022-06-23
Dynamical Sea Ice Model

LenkaNovak:
## Purpose and Content
- the dynamical sea ice from PR#50 needs to be tested and finalized as a stand-alone driver (addressing SDI #42 ), separate from the `moist_mpi_earth` directory, which will contain the prescribed sea ice and bucket model (PR#51 that addresses SDI #41 ).

### To do:
- [x] separate example from the bucket / prescribed (`moist_mpi_earth`) directory
- [x] energy conservation test - , neglecting the initial equilibration over the first 2 timesteps, up to 0.1% error after 27d
    - first 100 timesteps (dt  =200s)
    Screen Shot 2022-06-24 at 10 23 14 AM
    
    - 27 days
    Screen Shot 2022-06-24 at 10 22 32 AM
- [x] physical test - h_ice (m)

https://user-images.githubusercontent.com/38083334/175771964-65e1b164-ee53-4cb3-9d3a-7151d5909236.mp4

### Future PRs:
- update for the latest interface (after #51 is ready)
- add area fraction (apply to all examples)
- prescribe albedo
- make an abstraction for the sea ice model + thermal slab
- replace heavisides with smooth functions
- add regression test
- add buildkite test
- investigate if the small conservation error grows for longer integrations

## Benefits and Risks
- benefit: finalizes the sea ice model development, which will be useful when testing all available ODE surface-model setups, when appraising the AMIP configuration. 
- risk: the dynamical sea ice is strictly speaking not needed for the first AMIP simulation (however, this is already implemented and only requires some time for testing. Completing it now will save time later. )

## Linked Issues
(Provide references to any link issues. Use closes #issuenum to automatically close an open issue)
- Fixes #
- Closes #

## PR Checklist
- [x] This PR is linked to SDI #42.

### DIFF
--- a/experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/slab_ice/slab_init.jl
+++ b/experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/slab_ice/slab_init.jl
@@ -1,0 +1,213 @@
+# slab_ice
+
+# sea-ice parameters
+struct IceSlabParameters# <: CLIMAParameters.AbstractEarthParameterSet{F} 
+    h::FT # sea ice height
+    ρ::FT
+    c::FT
+    T_init::FT
+    z0m::FT
+    z0b::FT
+    ρc_ml::FT    # density times heat transfer coefficient for mixed layer [J / m2 / K ]
+    F0_base::FT  # ice base transfer coefficient [W / m2 / K]
+    T_base::FT   # ice base temperature [K] 
+    L_ice::FT    # latent heat coefficient for ice [J / m3]
+    h_ml::FT     # mixed layer depth [m] 
+    T_freeze::FT # temperature at freezing point [K] 
+    k_ice::FT   # thermal conductivity of ice [W / m / K]
+    α::FT # albedo
+    σ::FT # Stefan Boltzmann constant
+end
+
+# init simulation
+function slab_ice_space_init(::Type{FT}, space, p) where {FT}
+
+    # prognostic variable
+    Y = Fields.FieldVector(T_sfc = ones(space) .* p.T_freeze, h_ice = zeros(space), T_ml = ones(space) .* p.T_freeze)
+
+    return Y, space
+end
+
+"""
+    solve_ice!(dT_sfc, T_sfc, (parameters, F_accumulated), t)
+
+slab RHS with an implicit solve ice and explicit (forward Euler) solve for ocean
+
+"""
+function solve_ice!(integ, Δt)
+
+    Y = integ.u
+    Ya = integ.p.Ya
+    p = integ.p.Ya.params
+
+    # prognostic
+    T_ml = Y.T_ml
+    T_sfc = Y.T_sfc
+    h_ice = Y.h_ice
+
+    # auxiliary
+    F_atm = @. Ya.F_aero + Ya.F_rad
+    ∂F_atm∂T_sfc = get_∂F_atm∂T_sfc(p, T_sfc, Ya) # this will be passed from atmos/SF.jl
+    F_base = similar(T_sfc)
+
+    sm = semtner_zero_layer_model.(T_ml, T_sfc, h_ice, F_atm, ∂F_atm∂T_sfc, Ref(p), Ref(Δt))
+
+    T_ml .= sm.T_ml
+    h_ice .= sm.h_ice
+    T_sfc .= sm.T_sfc
+    F_base .= sm.F_base
+
+    # update state
+    @. Y.T_ml = T_ml
+    @. Y.h_ice = h_ice
+    @. Y.T_sfc = T_sfc
+
+    Ya.ice_mask .= get_ice_mask.(h_ice)
+    Ya.F_base .= F_base
+
+    return nothing
+end
+
+
+function semtner_zero_layer_model(T_ml, T_sfc, h_ice, F_atm, ∂F_atm∂T_sfc, p, Δt)
+
+    # local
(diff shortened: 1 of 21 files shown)


LenkaNovak:
@kmdeck this is ready to be merged as a checkpoint for the dynamic sea ice. It is now in a separate directory, so it shouldn't conflict with the Bucket PR, and we can pick it up again after AMIP. If you're ok with it, I'll wait for you to merge #57 and then rebase on that. As for documentation, the sea ice model is documented in the `experiments/ClimaCore` directory, which we can copy over when we have an abstraction for the sea ice model next week.

Manifest.toml
```diff
@@ -0,0 +1,929 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.7.2"
```
kmdeck, line 3: Why did the Manifest for ClimaCoupler change with this PR?

experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/mpi/mpi_postprocess.jl
```diff
@@ -0,0 +1,37 @@
...
+        global_Y_slab = DataLayouts.gather(comms_ctx, Fields.field_values(sol_slab.u[i].T_sfc))
+        if ClimaComms.iamroot(comms_ctx)
+            global_sol_u_atmos[i] = Fields.FieldVector(
+                c = Fields.Field(global_Y_c, global_center_space),
+                f = Fields.Field(global_Y_f, global_face_space),
+            )
+            global_sol_u_slab[i] = Fields.Field(global_Y_slab, global_h_space)
+        end
+    end
+    if ClimaComms.iamroot(comms_ctx)
```
kmdeck, line 28: This seems hardcoded to a slab+atmos. Will this work for this example? If not, should it be removed from the PR?

experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/slab_ice/slab_init.jl
```diff
@@ -0,0 +1,213 @@
...
+    tspan;
+    stepper = Euler(),
+    nelements = 6,
+    npolynomial = 4,
+    dt = 0.02,
+    saveat = 1.0e10,
+    space = nothing,
+    mask = nothing,
+    prescribed_sic_data = nothing,
+    ocean_params = (; ρ = FT(1e3), c = FT(4e3), h = FT(1)),
```
kmdeck, line 162: should these be enforced to be consistent with the slab_ocean params being used?

kmdeck (commented):
Hey Lenka - not sure of the best way to proceed, it's up to you. We can merge this as is (except for small changes noted in comments), as a copy of the sea ice model, but this code will have to be changed a lot to be used in practice (globals, unused files, etc). This is also not being tested, so it may break at some point (due to other packages changing). It might be better to bring it up to date with moist_mpi_earth changes (but still in it's own directory, so it can have its own versions of conservation checker, etc). Then we can add it to buildkite. and it will be easier to refactor, etc later for interface changes.

Manifest.toml
LenkaNovak: Ooh! Good catch! I think the formatting did this.

experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/slab_ice/slab_init.jl
```diff
@@ -0,0 +1,213 @@
...
+end
+
+get_∂F_atm∂T_sfc(p, T_sfc, Ya) = @. FT(4) * (FT(1) - p.α) * p.σ * T_sfc^3 + Ya.∂F_aero∂T_sfc
+
+function ∑tendencies_ice_stub(du, u, p, t)
+    dY = du
+    Y = u
+    FT = eltype(dY)
+
+    if p.Ya.prescribed_sic_data !== nothing # Prognostic eqn for T_sfc
```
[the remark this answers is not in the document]
LenkaNovak: I agree with this plan. I'm tempted to leave it in  for now though until we revamp the interface, as it is useful to have the prescribed_sic option as a check in case the Dynamic one is doing something funky.

experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/slab_ice/slab_init.jl
LenkaNovak: True, I made it a non-keyword argument, so there is no accidental imposition of the default ocean parameters.

experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/mpi/mpi_postprocess.jl
LenkaNovak: Good point - btw, if we wanted to use it (say in the Bucket example), we'd need to combine all of the surface fields into one, like we do in the coupling loop in the `coupler_driver.jl`. But since we don't need MPI for this example imminently, and since we'll be changing the interface for this in the bucket example (and then broadcast it to the other examples), I think deleting this here makes sense. :)

LenkaNovak:
As discussed on Slack, we will go ahead with the merge, because we be won't be working on the dynamical sea ice for a while (except for perhaps making it an abstract type), so it would be good to checkpoint it now (as a working model integration PR), and then iterate with smaller PRs (with a new interface or enhancements to the physical parameterization) when we have time.

Issue #60, opened 2022-06-17
Remove parents


kmdeck:
Remove `parent` calls from moist_mpi_earth bucket/atmos/ice/ocean coupled sim.

LenkaNovak:
to be addressed in #205

PR #80, opened 2022-07-06
Enable broadcasting with mismatched spaces.

### LINKED ISSUE #79: Enable broadcasting with mismatched spaces
We need to implement this formally in ClimaCoupler to remove our `parent` calls.

```
import ClimaCore.Fields
Fields.allow_mismatched_diagonalized_spaces() = true
```

This changes the error to a warning when the spaces are the same but different instances. I think we can simply put this in the ClimaCoupler.jl module/file so that this is automatically enabled when the Coupler is imported. We will also probably want to turn off warning logs so that the warning is not printed a million times during a coupled simulation:

```
import Logging
Logging.disable_logging(Logging.Warn)
```

Not sure if we can target this specific warning to be filtered out.

_Originally posted by @jb-mackay in https://github.com/CliMA/ClimaCore.jl/issues/640#issuecomment-1139209023_

jb-mackay:
## Purpose and Content
Turns off broadcasting errors that stem from mismatched spaces using the tools from [ClimaCore#640](https://github.com/CliMA/ClimaCore.jl/pull/640#issuecomment-1139209023).

## Benefits and Risks
This is necessary to couple components without using `parent`. Will allow #60 to be addressed. Current risks surround warning suppression - to prevent a warnings from being printed each broadcast, warnings and below are disabled in the logger. A more target approach would be more useful to only hide warnings stemming from broadcasting with mismatched spaces.

## Linked Issues
- Closes #79 

## PR Checklist

### DIFF
--- a/src/ClimaCoupler.jl
+++ b/src/ClimaCoupler.jl
@@ -5,6 +5,18 @@ Coupling module sufficient for initial atmos-ocean-land coupled simulation.
 """
 module ClimaCoupler
 
+using ClimaCore
+import ClimaCore: Fields, Operators
+import Logging
+
+# Disable errors from broadcasting with mismatched spaces
+Fields.allow_mismatched_diagonalized_spaces() = true
+Operators.allow_mismatched_fd_spaces() = true
+
+function __init__()
+    Logging.disable_logging(Logging.Warn) # disable warnings (for broadcasting with mismatched spaces)
+end
+
 include("CoupledSimulations/clock.jl")
 include("CoupledSimulations/coupled_simulation.jl")
 include("CouplerState/coupler_state.jl")
--- a/Project.toml
+++ b/Project.toml
@@ -5,5 +5,6 @@ version = "0.1.0"
 
 [deps]
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
--- a/test/general_tests.jl
+++ b/test/general_tests.jl
@@ -1,0 +1,31 @@
+using Test
+using ClimaCoupler
+import ClimaCore: Fields, Domains, Topologies, Meshes, Spaces, Geometry
+using IntervalSets
+
+function spectral_space_2D(; n1 = 1, n2 = 1, Nij = 4)
+    domain = Domains.RectangleDomain(
+        Geometry.XPoint(-1.0) .. Geometry.XPoint(1.0),
+        Geometry.YPoint(-1.0) .. Geometry.YPoint(1.0),
+        x1periodic = false,
+        x2periodic = false,
+        x1boundary = (:east, :west),
+        x2boundary = (:south, :north),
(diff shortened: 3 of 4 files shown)


src/ClimaCoupler.jl
```diff
@@ -5,6 +5,18 @@ Coupling module sufficient for initial atmos-ocean-land coupled simulation.
...
+import ClimaCore: Fields, Operators
+import Logging
+
+# Disable errors from broadcasting with mismatched spaces
+Fields.allow_mismatched_diagonalized_spaces() = true
+Operators.allow_mismatched_fd_spaces() = true
+
+function __init__()
+    Logging.disable_logging(Logging.Warn) # disable warnings (for broadcasting with mismatched spaces)
+end
```
jb-mackay, line 18: @charleskawczynski @simonbyrne is there a better way to prevent these warnings from printing? 

One option is to remove the warning all together if we think that the error should only be disabled when coupling anyways.

Another is perhaps to change the `@warn` to `@debug` so that they are at the lowest logging level and hidden by default.

Any suggestions?

src/ClimaCoupler.jl
```diff
@@ -5,6 +5,18 @@ Coupling module sufficient for initial atmos-ocean-land coupled simulation.
 """
 module ClimaCoupler
 
+using ClimaCore
+import ClimaCore: Fields, Operators
+import Logging
+
+# Disable errors from broadcasting with mismatched spaces
+Fields.allow_mismatched_diagonalized_spaces() = true
+Operators.allow_mismatched_fd_spaces() = true
```
charleskawczynski, line 14: This should be moved into the test folder, it must be done at global scope.

charleskawczynski: The rest of the changes in this file can probably be deleted

jb-mackay: Can you explain your reasoning for both of these points? I was thinking we want to disable these errors anytime we are using the coupler. Maybe that is simply too broad? The benefit is that coupler users won't have to think about these levers.

charleskawczynski: This is kind of like type piracy. We should probably avoid changing other packages default behavior under the hood. What's the issue with doing this at the global level? I'm fairly certain it'll work, it's simple, and more transparent.

jb-mackay: That makes sense to me r.e. transparency. I just wanted to allow the user to not have to think about this when coupling and these errors are something we will want to disable pretty much every time we are coupling, therefore something more automatic seemed warranted. 

Maybe a method that redefines these would be a happy medium? Something like
```julia
function allow_mismatched_spaces()
    @eval Fields.allow_mismatched_diagonalized_spaces() = true
    @eval Operators.allow_mismatched_fd_spaces() = true
end
```

I just feel that `allow_mismatched_diagonalized_spaces` and `allow_mismatched_fd_spaces` are pretty niche.

charleskawczynski: > I just wanted to allow the user to not have to think about this

This is not transparent.

My recommendation is to include this at the top level. I recommend not using `@eval`.

jb-mackay: > I recommend not using @eval. My recommendation is to include this at the top level. That is by far the simplest way to handle this situation. We can hide it later if we want to make things pretty.

Ok, I'm fine with holding off then.

PR #75, opened 2022-07-05
Sea Breeze Tutorial Docs

jb-mackay:
## Purpose and Content
This PR will turn the sea breeze test case into a tutorial to be included in the docs. This replaces tc1 which was removed in #62. 

The sea breeze driver, README, and all files included in the driver should be documented and included as tutorial pages.

## Benefits and Risks
This tutorial should clearly and concisely explain and demonstrate the different parts of the ClimaCoupler interface as well as how to set up a coupled simulation. It should be straightforward to add docs as the interface is enhanced.

## Linked Issues
- Closes #63 

## PR Checklist
- [x] This PR has a corresponding issue OR is linked to an SDI.
- [x] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [x] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [x] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [x] I linted my code on my local machine prior to submission OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code used in an integration test OR N/A.
- [x] All tests ran successfully on my local machine OR N/A.
- [x] All classes, modules, and function contain docstrings OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #83, opened 2022-07-09
Add `coupler_push!` and `coupler_pull!` to Sea Breeze

jb-mackay:
## Purpose and Content
Adds `coupler_push!` and `coupler_pull!` methods for each component in the sea breeze demo. These methods wrap how each component interacts with the coupler (sending & receiving via `coupler_put` and `coupler_get`). 

This PR also moves the coupled time-step into the `CouplerState` struct so that it is accessible in the push/pull calls. 

## Benefits and Risks
Push/pull: A major benefit is the conciseness of the main driver loop - which now consists of only `coupler_push!`, `step!` and `coupler_pull!` calls. This means we can standardize the simulation run loop and develop a set of general coupled timesteppers that take some inputted component order. This may not be a high priority as custom coupled steppers and schemes will be useful.

Coupled timestep storage: As the coupled timestepping info becomes more well defined, this may need to be refined (e.g. a CoupledTimestepper struct that could pair with the `CouplerState` struct as a full `Coupler`). This will likely be determined by how accumulation is ultimately handled.

## Linked Issues
Part of #44 

## PR Checklist

### DIFF
--- a/experiments/ClimaCore/sea_breeze/atmos_rhs.jl
+++ b/experiments/ClimaCore/sea_breeze/atmos_rhs.jl
@@ -233,6 +233,19 @@ function AtmosSimulation(Y_init, t_start, dt, t_end, timestepper, p, saveat, cal
     return AtmosSimulation(atm_integ)
 end
 
+function ClimaCoupler.coupler_push!(coupler::ClimaCoupler.CouplerState, atmos::AtmosSimulation)
+    coupler_put!(coupler, :F_sfc, atmos.integrator.u.F_sfc, atmos)
+end
+
+function ClimaCoupler.coupler_pull!(atmos::AtmosSimulation, coupler::ClimaCoupler.CouplerState)
+    # pre: reset flux accumulator
+    atmos.integrator.u.F_sfc .= 0.0 # reset surface flux to be accumulated
+
+    T_sfc_ocean = coupler_get(coupler, :T_sfc_ocean, atmos)
+    T_sfc_land = coupler_get(coupler, :T_sfc_land, atmos)
+    atmos.integrator.p.T_sfc .= T_sfc_land .+ T_sfc_ocean
+end
+
 # init simulation
 function atm_init(; xmin = -500, xmax = 500, zmin = 0, zmax = 1000, npoly = 3, helem = 20, velem = 20, bc = nothing)
 
--- a/test/CouplerState/cplstate_interface.jl
+++ b/test/CouplerState/cplstate_interface.jl
@@ -38,7 +38,7 @@ end
 
     simA = SimulationA(ones(spaceA))
     simB = SimulationB(zeros(spaceB))
-    coupler = CouplerState()
+    coupler = CouplerState(1.0)
 
     coupler_add_field!(coupler, :test1, simA.data; write_sim = simA)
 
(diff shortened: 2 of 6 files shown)


jb-mackay:
Will add more docs in #75

experiments/ClimaCore/sea_breeze/atmos_rhs.jl
```diff
@@ -233,6 +233,19 @@ function AtmosSimulation(Y_init, t_start, dt, t_end, timestepper, p, saveat, cal
...
+    coupler_put!(coupler, :F_sfc, atmos.integrator.u.F_sfc, atmos)
+end
+
+function ClimaCoupler.coupler_pull!(atmos::AtmosSimulation, coupler::ClimaCoupler.CouplerState)
+    # pre: reset flux accumulator
+    atmos.integrator.u.F_sfc .= 0.0 # reset surface flux to be accumulated
+
+    T_sfc_ocean = coupler_get(coupler, :T_sfc_ocean, atmos)
+    T_sfc_land = coupler_get(coupler, :T_sfc_land, atmos)
+    atmos.integrator.p.T_sfc .= T_sfc_land .+ T_sfc_ocean
```
jb-mackay, line 246: @LenkaNovak this is a good example of why we would want a `coupler_get` rather than `coupler_get!` - because the atmos boundary is shared by land and ocean we can't just use `.=` directly as in `coupler_get!`.

test/CouplerState/cplstate_interface.jl
```diff
@@ -38,7 +38,7 @@ end
 
     simA = SimulationA(ones(spaceA))
     simB = SimulationB(zeros(spaceB))
-    coupler = CouplerState()
+    coupler = CouplerState(1.0)
```
jb-mackay, line 41: > Coupled timestep storage: As the coupled timestepping info becomes more well defined, this may need to be refined (e.g. a CoupledTimestepper struct that could pair with the CouplerState struct as a full Coupler). This will likely be determined by how accumulation is ultimately handled.

This is one reason that having a separation here might be useful! You shouldn't really need time information when testing these `CouplerState` `put` and `get` methods. On the other hand, `push` and `pull` are more tied into the time coordination (since they would be part of a coupled step method). I still think we can punt on this.

LenkaNovak: I agree with the separation between put/get and push/pull. But I would expect `Δt_coupled` to be in the `CoupledSimulation` to mimic the model simulations. There seems to be quite a bit of cross-over between the `CoupledSimulation` and `CoupledState`. Any thoughts on combining these?

LenkaNovak: If we do this, it would be a separate PR anyway, so I'm happy to go ahead with merging this one.

LenkaNovak (approved):

test/CouplerState/cplstate_interface.jl
jb-mackay: > I agree with the separation between put/get and push/pull. But I would expect `Δt_coupled` to be in the `CoupledSimulation` to mimic the model simulations. There seems to be quite a bit of cross-over between the `CoupledSimulation` and `CoupledState`. Any thoughts on combining these?

I think there is a separation - `CoupledSimulation` should fit the Simulations interface and deals with packing up component simulations _and_ the coupler (currently just a `CouplerState`) and deals with how to execute things. More on the time stepper side as a whole.

`CouplerState` deals with the actual coupled fields and passing them around. This is the part that will deal with MPI and how things get distributed.

Issue #74, opened 2022-06-29
Simple I / O pipeline


LenkaNovak:
## Purpose
I/O pipeline coordinated by the coupler is necessary for providing restart files, BC files, model diagnostics, performance metrics and conservation checks for the AMIP run. 

- related to the [CC-i#747](https://github.com/CliMA/ClimaCore.jl/issues/747) issue on standard format

The desired outcome are the following features
- **online** 
  - model-grid dump
    - save checkpoint files containing the fully recoverable uninterpolated state by individual MPI PEs at a desired frequency
    - temporal means (extendable to temporal covariances and other similar stats on model grid)
  - BC field reader and manipulation
  - on-the-fly viz
  - other online metrics: conservation, performance
- **offline** 
  - component model diagnostics:
    - each model needs to write user-defined diagnostics interpolated on the lat/long grid on the fly at a desired frequency
    - basic diagnostics (intermediate solution)   
      - a version of this is already implemented as a callback in ClimaAtmos (code reduction possible, apply to netcdf or add a post processing utility) and needs to be implemented in ClimaLSM 
      - a more comprehensive list following the [CMIP guidelines](https://pcmdi.llnl.gov/CMIP6/Guide/modelers.html#4-model-output-fields) and the [Clima user diagnostics issue](https://github.com/CliMA/ClimateMachine.jl/issues/1596)
          - high priority output: https://github.com/CliMA/ClimateMachine.jl/issues/214 
          -  requirements: https://pcmdi.llnl.gov/mips/cmip3/variableList.html 
          - Cmip6 tables: https://github.com/PCMDI/cmip6-cmor-tables/tree/master/Tables, https://clipc-services.ceda.ac.uk/dreq/index/miptable.html
          - database: https://clipc-services.ceda.ac.uk/dreq/mipVars.html   
      - borrowing concepts of distributed diagnostics from the [DiagnosticsMachine](https://github.com/CliMA/ClimateMachine.jl/blob/9503cbc83eb098f0d92bef72745a1d214175940c/src/Diagnostics/DiagnosticsMachine/DiagnosticsMachine.jl)
    - other metrics: comparison with published benchmarks, MSE tables 

## Cost/Benefits/Risks
- benefit: enables the first AMIP showcase
- risk: diagnostics tend to become complicated very quickly, but here we focus on basic fields only to demonstrate the pipeline

## People and Personnel
- Lead: @LenkaNovak 
[...]

PR #87, opened 2022-07-13
Basic calendar functionality

LenkaNovak:
# PULL REQUEST

## Purpose and Content
Adds the usage of Dates and a `calendar_callback` macro, which allows evaluations on a particular date. This is a prerequisite PR for BC specification at different times from file. 

## Benefits and Risks
- benefit: 
  - enables AMIP BC specification
  - designed as a general callback, so will be useful for monthly diagnostics aggregation, message showing and logging
- risk: step-wise operations may slow down code - benchmarks in place

## Linked Issues
- #74 

## PR Checklist

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/calendar_timer.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/calendar_timer.jl
@@ -1,25 +1,29 @@
 # calendar timer
-#  - fascilitates calendar functions and temporal interpolations
+#  - facilitates calendar functions and temporal interpolations
 
 """
     current_date(t)
-- returns the model date
+Return the model date
 """
 current_date(t) = date0 + Dates.Second(t)
 
 """
-    strdate_to_datetime(strdate), datetime_to_strdate(datetime)
-- convert between String ("YYYYMMDD") and Date formats  
+    strdate_to_datetime(strdate)
+Convert from String ("YYYYMMDD") to Date format  
 """
 strdate_to_datetime(strdate::String) =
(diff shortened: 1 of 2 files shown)


experiments/AMIP/moist_mpi_earth/coupler_utils/calendar_timer.jl
```diff
@@ -0,0 +1,52 @@
+# calendar timer
+#  - facilitates calendar functions and temporal interpolations
+
+"""
+    current_date(t)
+Return the model date
+"""
+current_date(t) = date0 + Dates.Second(t)
```
jb-mackay, line 8: Going forward, it will be best if `date0` is not a global variable. The question then is, what is the natural place to store this? There is some code in src regarding a clock that could be expanded or changed to support this!

jb-mackay (approved):
Looks good @LenkaNovak - what are the scenarios which we will use these data callbacks? Can this be built around the time kept in the integrator?

LenkaNovak:
> Looks good @LenkaNovak - what are the scenarios which we will use these data callbacks? Can this be built around the time kept in the integrator?

Thanks for the review @jb-mackay ! 🚀  Yes, I would say that once we re-introduce the interface where the coupler has its own way of keeping track of `t`, we should def make the callback a function of that.

experiments/AMIP/moist_mpi_earth/coupler_utils/calendar_timer.jl
LenkaNovak: Yeah, there are a few places in the coming PRs where global variables are used (e.g. `landmask`). I can make an issue, so we don't forget to come back to this.

PR #94, opened 2022-07-18
update Manifest to use registered ClimaLSM

kmdeck:
# PULL REQUEST

## Purpose and Content
Upgrade Manifest to use ClimaLSM - registered version

Note that when I went to do so, there was a conflict between the IntervalSets compat in ClimaAtmos and ClimaLSM. I am broadening the compat in ClimaLSM to overlap with ClimaAtmos. That is the branch that is currently used in this PR. But, I'd like to wait to release another version of ClimaLSM until there are more changes, so this is on hold.


jb-mackay:
@kmdeck what other dependencies are requiring us to check in the Manifest for this experiment? If just LSM, we could remove the Manifest as long as the `compat` versions are specified in the Project.toml

kmdeck:
we actually can use the current registered ClimaLSM if we upgrade the version of atmos we are using (I think) - I saw a PR in ClimaAtmos that bumps the compat of CA to include 0.7 for Interval sets, which is what ClimaLSM uses.


We may not need to check in the Manifest, Im not sure. Would the above still cause an issue, though? If CA and CLSM have compat entries that do not overlap?

LenkaNovak:
Outdated and addressed in #143

Issue #67, opened 2022-06-20
Auto-building remap operations & Coupled Sim Init


jb-mackay:
Initializing a coupled simulation needs to be designed such that the actual remap operator construction can be hidden.

The init process might look something like:
1. Init all component simulations with existing standalone code
2. Init coupled simulation struct
3. Init maps using boundary spaces of each model
4. Reinit component simulations with updated coupled fields

LenkaNovak:
To be addressed in #211

PR #99, opened 2022-07-24
Add remap to `coupler_put!`

### LINKED ISSUE #65: Add remapping to `coupler_put!`
Currently it is not possible to map to an exchange or intermediary grid -- any field `put!` in the coupler lives on the space of the simulation that puts it there.

This is because of the current way the mapping is set up which uses a tag (`write_sim`) to specify the source space/sim when remapping. This tag also (as the name suggests) indicates which simulation has write permissions to that coupled field. 

Perhaps the right solution is to remove the double usage of `write_sim` and add another tag like "`space_name`" which would be part of the remap operator name.

Adding remapping to the `put!` side of things creates symmetry and more flexibility for what grids the coupling takes place on. It creates the potential that there could be a coupler-owned field in memory, rather than the coupler only registering pointers.

_Originally posted by @jb-mackay in https://github.com/CliMA/ClimaCoupler.jl/issues/58#issuecomment-1159262614_

jb-mackay:
## Purpose and Content
This PR adds remapping to the `coupler_put!` calls and changes the way that the remap operators are stored/accessed in the coupler.

## Benefits and Risks
Having remapping at both the `put!` and `get` stages of the coupling enables the use of an exchange grid, should the user want one. This is optional, though also the standard in the wider coupling community. It's also nice to make `put!` and `get` more symmetric.

The way that maps are stored is also changed from keys with `simA_to_simB` naming convention to a key that is a tuple of the target and source spaces. This has some nice effects: 
- the naming system is abandoned and not necessary for the user to learn
- it is not necessary to store exchange grid names, which would otherwise be necessary for accessing maps during coupler_put!` calls
- it is much easier to build the maps automatically (see `coupler_add_field!` changes), likely making #67 easier to tackle.

In talking to @simonbyrne, a risk here is the time to hash spaces vs. names.

It should also be noted that a formal `IdentityRemap` operator is beneficial here and further simplifies some of the remap construction and coupler calls (since you can assume _every_ call has an associated map, even if it does nothing). [ClimaCore PR#793](https://github.com/CliMA/ClimaCore.jl/pull/793) adds this and should be treated as a dependency.

## Linked Issues
- Closes #65
- #67 

## PR Checklist
- [ ] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [ ] Unit tests are included OR N/A.
- [ ] All tests ran successfully on my local machine OR N/A.
- [ ] All classes, modules, and function contain docstrings OR N/A.
- [ ] Documentation has been added/updated OR N/A.

### DIFF
--- a/src/CouplerState/coupler_state.jl
+++ b/src/CouplerState/coupler_state.jl
@@ -37,7 +37,11 @@ etc... can be embeded in the intermdediate coupling layer.
 A field is exported by one component and imported by one or more other components.
 """
 function CouplerState(Δt_coupled)
-    return CouplerState(Dict{Symbol, CplFieldInfo}(), Dict{Tuple{ClimaCore.Spaces.AbstractSpace, ClimaCore.Spaces.AbstractSpace}, Operators.LinearRemap}(), Δt_coupled)
+    return CouplerState(
+        Dict{Symbol, CplFieldInfo}(),
+        Dict{Tuple{ClimaCore.Spaces.AbstractSpace, ClimaCore.Spaces.AbstractSpace}, Operators.LinearRemap}(),
+        Δt_coupled,
+    )
 end
 
 """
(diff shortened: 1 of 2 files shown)


jb-mackay:
TODO:
Docs update
Update Sea Breeze experiment with changes
Update `coupler_put!` tests to include mapping checks.

One thing that needs to be thought out is the map construction. If there is an exchange space, we don't want to build maps by hand from one sim to another. One options is building the map during the first `coupler_get` calls, or whatever the equivalent would be during initialization of the coupler. This would build the maps for accessing fields from the coupler, while the maps for putting fields in the coupler are built during `coupler_add_field`.

The user probably doesn't need to call `coupler_add_map!` directly anymore!

juliasloan25:
this is pretty out of date and I don't think it will be merged

PR #91, opened 2022-07-13
Daily interpolation from monthly data

LenkaNovak:
# PULL REQUEST

## Purpose and Content
Adds capability to interpolate linearly between two monthly Fields. The demo uses prerequisite PRs outlined in #88 to obtain monthly data. 

## Benefits and Risks
- benefit: last piece for enabling specifying temporarily varying boundary conditions from a file.
- risk: performance slow-down - check are in place (see below)

## Linked Issues
- SDI: #74 
- Design: #88

## Dependent PRs
- no not merge before  #87 ,  #89 and #90

## Performance of the final solution 
- time loop only over added infrastructure of #88 : adding 0.24s per simulation day at `dt = 200s`

## PR Checklist
- [ ] This PR has a corresponding issue OR is linked to an SDI.
- [ ] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [ ] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [ ] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [ ] I linted my code on my local machine prior to submission OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code used in an integration test OR N/A.
- [ ] All tests ran successfully on my local machine OR N/A.
- [ ] All classes, modules, and function contain docstrings OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #90, opened 2022-07-13
Monthly updating of BCs from file

LenkaNovak:
# PULL REQUEST

## Purpose and Content
This PR adds functionality to collect BC field information in a struct `..._info`, which is then updated each month with a new file read of the regridded data. In the next PR `..._info` will contain two Fields read from a file, which will be used for the daily interpolation.

## Benefits and Risks
- benefit: prerequisite for the monthly to daily interpolation
- risk: changes the way the BCs are read, but most of it is abstracted

## Linked Issues
- SDI: #74 
- Design: #88

## Dependent PRs
- no not merge before  
  - [x] #87 
  - [x]  #89
  - [x]  #93 
 
## Coupling loop performance
- the macro is comparable to using functions (though @time larger variability). The pros and cons for this method are [here](https://github.com/CliMA/ClimaCoupler.jl/issues/95), are will be revisited post AMIP (and well as including the unit test in the same file).
- in either case the monthly update does not cause noticeable performance change from the previous commit for the full AMIP setup. 

## PR Checklist
- [x] This PR has a corresponding issue OR is linked to an SDI.
- [x] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [x] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [x] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [x] I linted my code on my local machine prior to submission OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code used in an integration test OR N/A.
- [x] All tests ran successfully on my local machine OR N/A.
- [x] All classes, modules, and function contain docstrings OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #89, opened 2022-07-13
Separate regridding from file reader

LenkaNovak:
# PULL REQUEST

## Purpose and Content

This PR re-organises the regridding `coupler_utils`, so that regridding is separate from reading the regridded output from `TempestRemap`. 

- note that mesh generation with the space-filling curve breaks the regridding order required in `TempestRemap`. This was fixed with specifying a `regrid_space` using the previous method of mesh generation. However, this should be generalized in the future.  

## Benefits and Risks
- benefit: this will enable a monthly callback for temporal interpolation 
- risk: few more lines to the interface (though this will be removed by the following PR)

## Linked Issues
- SDI: #74 
- Design: #88

## Dependent PRs
- [x] do not merge before  #87 

## PR Checklist
- [x] This PR has a corresponding issue OR is linked to an SDI.
- [x] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [x] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [x] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [x] I linted my code on my local machine prior to submission OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code used in an integration test OR N/A.
- [x] All tests ran successfully on my local machine OR N/A.
- [x] All classes, modules, and function contain docstrings OR N/A.
- [x] Documentation has been added/updated OR N/A.

Issue #88, opened 2022-07-13
Time Interpolator Design


LenkaNovak:
Provides additional info for #74 

For temporarily variable boundary conditions specified from files, it is necessary to limit file reading for performance purposes. The AMIP protocol dictates that monthly files are interpolated to daily values. It was found to be most efficient to separate the calls into 3 stages:
- once: regrid all boundary data during initialization and save sparse matrix as temporary file
- monthly: extract the current and next month's data as `Fields` needed for the linear interpolation during that month (note that we're already using pre-processed mid-monthly data, so that the monthly means of the observed values are preserved)
- each timestep: interpolate between two monthly `Fields`

This will require the following features
- [x] calendar #87 
- [x] regridding separate from file reading #89 
- [x] interface changes #92 
- [x] callback to update monthly `Fields` #90
- [x] daily interpolation #91 #105

LenkaNovak:
Complete

PR #105, opened 2022-07-28
Daily interpolation from monthly data (v2)

LenkaNovak:
# PULL REQUEST
- Update from #91

## Purpose and Content
Adds capability to interpolate linearly between two monthly `Fields`. The demo uses prerequisite PRs outlined in #88 to obtain monthly data. 

## Benefits and Risks
- benefit: last piece for enabling specifying temporarily varying boundary conditions from a file.
- risk: performance slow-down - check are in place (see below)

## Linked Issues
- SDI: #74 
- Design: #88

## Dependent PRs
- no not merge before  #87 ,  #89 and #90

## Performance of the final solution 
- time loop only over added infrastructure of #88 : adding 0.24s per simulation day at `dt = 200s`

## PR Checklist

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
@@ -17,10 +17,10 @@ The inputs are:
     monthly_fields::C               # Tuple of the two monthly fields, that will be used for the daily interpolation
     segment_length::Vector{Int}     # length of each month segment (used in the daily interpolation)
     scaling_function::O             # function that scales, offsets or transforms the raw variable
-    interpolate_daily::I            # switch to trigger daily interpolation
-    land_mask::M                     # mask with 1 = land, 0 = ocean / sea-ice
+    interpolate_daily::Bool         # switch to trigger daily interpolation
+    land_mask::M                    # mask with 1 = land, 0 = ocean / sea-ice
 """
-struct BCFileInfo{F, S, V, D, C, O, I, M}
+struct BCFileInfo{F, S, V, D, C, O, M}
     FT::F
     datafile_cgll::S
     varname::V
@@ -31,7 +31,7 @@ struct BCFileInfo{F, S, V, D, C, O, I, M}
     monthly_fields::C
     segment_length::Vector{Int}
     scaling_function::O
-    interpolate_daily::I
+    interpolate_daily::Bool
     land_mask::M
 end
 
@@ -83,12 +83,12 @@ function bcfile_info_init(
         varname,
         weightfile,
         data_dates,
-        segment_idx0 .- Int(1),
+        deepcopy(segment_idx0),
         segment_idx0,
         current_fields,
         segment_length,
         scaling_function,
-        [interpolate_daily],
+        interpolate_daily,
         land_mask,
     )
 
@@ -105,7 +105,6 @@ Extracts boundary condition data from regridded (to model grid) NetCDF files (wh
 function update_midmonth_data!(date, bcf_info)
 
     # monthly count
-    bcf_info.segment_idx[1] += Int(1)
 
     all_dates = bcf_info.all_dates
     midmonth_idx = bcf_info.segment_idx[1]
@@ -133,7 +132,10 @@ function update_midmonth_data!(date, bcf_info)
                 ),
             Tuple(1:length(monthly_fields)),
         )
+        bcf_info.segment_length .= FT(0)
     elseif (midmonth_idx == midmonth_idx0) && (Dates.days(date - all_dates[midmonth_idx]) < 0) # for init
+        midmonth_idx = bcf_info.segment_idx[1] -= Int(1)
+        midmonth_idx = midmonth_idx < Int(1) ? midmonth_idx + Int(1) : midmonth_idx
         @warn "this time period is before BC data - using file from $(all_dates[midmonth_idx0])"
         map(
             x ->
@@ -149,6 +151,7 @@ function update_midmonth_data!(date, bcf_info)
                 ),
             Tuple(1:length(monthly_fields)),
         )
+        bcf_info.segment_length .= FT(0)
     elseif Dates.days(date - all_dates[end - 1]) > 0 # for fini
         @warn "this time period is after BC data - using file from $(all_dates[end - 1])"
         map(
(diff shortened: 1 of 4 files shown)


experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -133,7 +132,10 @@ function update_midmonth_data!(date, bcf_info)
                 ),
             Tuple(1:length(monthly_fields)),
         )
+        bcf_info.segment_length .= FT(0)
     elseif (midmonth_idx == midmonth_idx0) && (Dates.days(date - all_dates[midmonth_idx]) < 0) # for init
+        midmonth_idx = bcf_info.segment_idx[1] -= Int(1)
```
kmdeck, line 137: It looks like segment_idx and segment_idx0 are vectors, but the first element is always used. Can these be ints instead of Vectors?

The description of segment_idx0 also makes it sound like it should be a single element.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -192,3 +197,40 @@ function update_midmonth_data!(date, bcf_info)
...
+
+# IO - daily
+"""
+    interpolate_midmonth_to_daily(date, bcf_info)
+
+Interpolates linearly between two `Fields` in the `bcf_info` struct, or returns the first Field if interpolation is switched off. 
+"""
+function interpolate_midmonth_to_daily(date, bcf_info)
+    @show bcf_info.segment_length[1]
+    if bcf_info.interpolate_daily[1] && bcf_info.segment_length[1] > FT(0)
```
kmdeck, line 209: Isnt interpolate_daily a Boolean? why the indexing?

(another comment - in the struct defining BCInfo, we could make interpolate_daily::Bool right away, instead of making it interpolate_daily::I with I a parametric type.)

kmdeck: indexing on segment_length also confusing - a length is a single #?

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -192,3 +197,40 @@ function update_midmonth_data!(date, bcf_info)
...
+        segment_length = bcf_info.segment_length
+        segment_idx = bcf_info.segment_idx
+        all_dates = bcf_info.all_dates
+        monthly_fields = bcf_info.monthly_fields
+
+        return intepol.(
```
kmdeck, line 215: intepol -> interpol

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -192,3 +197,40 @@ function update_midmonth_data!(date, bcf_info)
...
+        )
+    else
+        return bcf_info.monthly_fields[1]
+    end
+end
+
+"""
+    intepol(f_t1, f_t2, t, Δt, FT)
+
+Performs linear interpolation within a segment `Δt` = (t1 - t2), of fields `f_t1` and `f_t2`, with `t`(`f_t1`) = t1.
```
kmdeck, line 230: with f(t_1) = f_1, right? Not t(f_1) = t_1.

kmdeck: maybe also clearer to have t_2 > t_1, and dt  = t2-t1?

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -192,3 +197,40 @@ function update_midmonth_data!(date, bcf_info)
...
+"""
+function intepol(f1, f2, t, Δt, FT)
```
kmdeck, line 232: you can instead do "interpol(f1::FT, f2::FT, t::FT, dt::FT) where {FT}" and not pass in the FT as well.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -192,3 +197,40 @@ function update_midmonth_data!(date, bcf_info)
...
+    interp_fraction = FT(t) / Δt
+    @assert abs(interp_fraction) <= FT(1) "time interpolation weights must be <= 1, but `interp_fraction` = $interp_fraction"
```
kmdeck, line 234: I dont think this formula is correct. 
For example:
If t = t1, we get f1*t1/Delta +f2*(Delta-t1)/Delta = 
f1*t1/Delta - f2 *t2/Delta. It should be f1.

I think it should be f = (f2-f1)/(t2-t1)(t-t1) + f1. Then f(t1) = f1, f(t2) = f2, and f((t1+t2)/2) = (f1+f2)/2.

(Also, it should not matter if t is negative, the formula should work still.)

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
LenkaNovak: Yeah, I wanted to make it a single element, but it needs to be mutable (at least `segment_idx` needs to be updated every month). Is there a better way of doing this? 🤔 I've heard mutable structs are not great for performance...

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
LenkaNovak: Actually, this one we can change. Originally I had it set up to be mutable (so that after init it switched to true), but in this version it can be static.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
kmdeck: good question! Maybe @jb-mackay knows? 

I think maybe the answer is that if it is mutating, it should be not stored in in the struct, but instead passed in as an argument to functions that need it.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
LenkaNovak: Ahh, the notation is a little confusing. 🤯  `t` here actually does refer to difference between the current date and `t1`, which I believe makes it equivalent to your suggestion. I failed at trying to be less verbose, haha! I will call it `\Delta_tt1` and `\Delta_t1t2`.  This is why reviews are super helpful! :) As for the assert, this is just catching cases where the dates are not passed in properly, so the current date is outside of range [t1,t2].

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
LenkaNovak: Hmmm, not sure exactly how you mean in this context. Would you mind giving an example?

LenkaNovak:
All issues above have been addressed, except whether to use functions instead of the mutable vectors in static structs. At the last sync meeting it was decided that since this is not a breaking issue, we will merge this and address this as part of a future PR.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
kmdeck: hey @LenkaNovak ! Happy to chat about this more. This is what I was thinking...

instead of this:
```
function foo(bc_info::BCDataInfo)
    # do something with bc_info.segment_idx[1]
end

# in main driver couper loop:
 bc_info.segment_idx[1] += 1 # or however it is updated
foo(bc_info)
```
we could instead remove `segment_idx` from the struct, and do something like this
```
function foo(bc_info::BCDataInfo, segment_idx::Int)
    # does something with segment_idx
end

# in main driver
segment_idx = 1 # initialize somehow
# in coupler loop
segment_idx += 1 # update
foo(bc_info, segment_idx)
```

kmdeck: The thought was that if we have an integer valued object that is changing value every step, why store it in the struct? why not update it every step of the main coupler loop?

.Another thought along those lines - Maybe cleaner conceptualy if the BCDataInfo struct holds only static information about the BC file and how it is read and handled.

PR #132, opened 2022-10-11
Bring fluxes up to date with ClimaAtmos0.5.0

LenkaNovak:
Co-authored-by: Zhaoyi Shen <szy21@users.noreply.github.com>

## Purpose 

Update for new ClimaAtmos0.5.0 surface flux calculation interface, which enables the Coupler to use the same adapter functions to calculate fluxes using SurfaceFluxes.jl. It reduces code duplication. 

## Content

- This requires revamping the flux calculation in the coupler and use the new cash variables. 
- We also add frictional dissipation to the energy calculation. Energy conservation checks are passing.
- We added a more realistic initial temperature to the bucket model.
- 
## Checks between ClimaAtmos0.5.0 and ClimaAtmos0.4.0 
- slabplanet energy conservation
Screen Shot 2022-10-11 at 11 02 21 AM

- AMIP without radiation (snapshot of zonal wind after 10 days)

Screen Shot 2022-10-11 at 11 02 54 AM

- AMIP with radiation 
  - ~~this is numerically unstable due to strong tropical fluxes lead to negative moisture. This can be addressed by adding a spin up and tuning the forcing parameters for the spin up. However, @szy21 and I decided to do such tuning for the target resolution, and submit the current Buildkite drivers without radiation. The reason why this did not crash for ClimaAtmos0.4.0 because `T_sfc` was not communicated into RRTMGP. All other initial conditions were the same. This is now fixed.~~
  - after adjusting the bucket T_init to be more realistic, a stable run is obtained. Below is the monthly-mean output after the first 32 days. Note the `T_sfc` over land is closer to expected compared to the current main (though a thorough validation will be performed as part of the long runs planned for the next PR).  
 
Screen Shot 2022-10-11 at 11 57 50 AM

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/conservation_checker.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/conservation_checker.jl
@@ -9,6 +9,7 @@ struct OnlineConservationCheck{A} <: AbstractCheck
     ρe_tot_seaice::A
     toa_net_source::A
     ice_base_source::A
+    friction_sink::A
 end
 
 function array2field(array, space) # reversing the RRTMGP function field2array (TODO: this now exists in ClimaAtmos)
@@ -53,6 +54,7 @@ function check_conservation(cc::OnlineConservationCheck, coupler_sim)
     # global sums
     atmos_e = sum(u_atm)
 
+    # save radiation source
     if radiation != nothing
         face_space = axes(atmos_sim.integrator.u.f)
         z = parent(Fields.coordinate_field(face_space).z)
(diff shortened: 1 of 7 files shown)


experiments/AMIP/moist_mpi_earth/coupler_utils/conservation_checker.jl
```diff
@@ -129,16 +144,20 @@ function plot_global_energy(cc, coupler_sim, figname1 = "total_energy.png", fign
...
     Plots.plot!(times_days, diff_ρe_tot_slab[1:length(times_days)], label = "land")
     Plots.plot!(times_days, diff_ρe_tot_slab_seaice[1:length(times_days)], label = "seaice")
     Plots.plot!(times_days, diff_toa_net_source[1:length(times_days)], label = "toa")
+    Plots.plot!(times_days, diff_friction_sink[1:length(times_days)], label = "friction")
     Plots.plot!(times_days, diff_ρe_tot_slab_ocean[1:length(times_days)], label = "ocean")
 
-    tot = cc.ρe_tot_atmos .+ cc.ρe_tot_ocean .+ cc.ρe_tot_land .+ cc.ρe_tot_seaice .+ cc.toa_net_source
+    tot =
+        cc.ρe_tot_atmos .+ cc.ρe_tot_ocean .+ cc.ρe_tot_land .+ cc.ρe_tot_seaice .+ cc.toa_net_source .+
+        cc.friction_sink
```
szy21, line 160: I wonder if we actually don't need to take into account the friction_sink here, as at least in the atmosphere the surface drag of momentum doesn't enter the tendency of ρe_tot. (I guess kinetic energy flux is already taken into account when calculating surface flux of total energy.)

szy21 (approved):
Looks good! I just have one comment on the energy conservation.

experiments/AMIP/moist_mpi_earth/coupler_utils/conservation_checker.jl
LenkaNovak: Discussed offline - the effects of this will be investigated in the future. For now we'll ignore KE dissipation here, so as to be consintent with the total energy tendency calculation.

Issue #134, opened 2022-10-17
ClimaCoupler interface


juliasloan25:
## Purpose
We want to refactor the ClimaCoupler to make a more modular interface.

## Benefits
This will remove global-scope objects, making performance tracking faster and increasing usability of the interface. Code will be more generalized and therefore able to be used by different modules more easily.

## Risks
We will develop an interface that will conflict with the current main stream of the model. To minimize this risk, we are developing this interface in parallel with development of the rest of the model. Eventually we will change over the rest of the code to align with this new interface design.

## People and Personnel
- Lead: @juliasloan25 
- Collaborators: @LenkaNovak
- Reviewers: @charleskawczynski @dennisYatunin @valeriabarra 

## Components
1. Modularize utilities used by the coupler driver
  a. Check unit test coverage
2. Move functions from the coupler driver to an external module
  a. Check unit test coverage
3. Optimize and generalize modules used by driver
  a. Re-evaluate scope of global variables and make them local if possible
  b. Add coupler-specific abstractions which are consistent with other CliMA model abstractions
  c. Optimize, generalize existing coupler functions

### Module Conversion Task Breakdown
For each module: 
PHASE 1 - moving and documentation
1. Copy file(s) to module in `src/`
2. Add module to `src/ClimaCoupler.jl`
3. Docstring for each function
  a. Add input/output types in docstring if not type annotated
  b. Add docstring info to docs/src, docs/make.jl
4. Add exports for externally-used functions
5. Flag potentially-removable parts of code

PHASE 2 - testing
1. Create test file in `test/`, add to `test/runtests.jl`
2. Unit test for each exported function in module
  a. Test simplest functions first
3. Integration tests as needed
4. Add any tests that use MPI in `test/mpi_tests` and in buildkite pipeline
  a. Add file to `run_mpi_tests.jl`

PHASE 3 - cleanup
1. Remove flagged parts of code
  a. Test after each one
2. Check all imports are necessary
3. Formatting
4. Read through comments

## Results and deliverables
- One module for each utility used by the driver
  - Regridding (masker.jl, regridder.jl)
  - Conservation checker (conservation_checker.jl)
  - Input file reader (bcfile_reader.jl)
  - Callback manager + calendar (calendar_timer.jl)
  - General utilities (general_helper.jl, flux_calculator.jl)
  - Diagnostics/visualization/output file generator (viz_explorer.jl, output_dumper.jl)
[...]

PR #144, opened 2022-10-26
Regridder interface

juliasloan25:
## Purpose 
Move functions from `coupler_utils/regridder.jl` to `Regridder` module in `src/`, and ensure that they are thoroughly commented and tested.
Additionally, create a `Utilities` module in `src/` containing functions commonly used by other coupler modules.

Part of SDI #134

## To-do

PHASE 2 - testing
- [x] Create test file in `test/`
  - [x] Test simplest functions first
- [x] Integration tests as needed
- [x] Create test file in `test/mpi_tests` and add any tests that use MPI

PHASE 3 - cleanup
  - [x] Test after each one
- [x] Check all imports are necessary
- [x] Formatting
- [x] Read through comments


juliasloan25:
bors try

experiments/AMIP/moist_mpi_earth/coupler_driver.jl
```diff
@@ -64,6 +64,16 @@ using ClimaCore: InputOutput
...
+# modify parsed args for fast testing from REPL #hide
+# parsed_args["coupled"] =  true #hide
+# parsed_args["moist"]= "equil" #hide
+# parsed_args["vert_diff"] =  true #hide
+# parsed_args["rad"] =  "gray" #hide
+# parsed_args["microphy"] =  "0M" #hide
+# parsed_args["energy_check"] = true
+# parsed_args["precip_model"] =  "0M" #hide
+# parsed_args["mode_name"] =  "slabplanet" #hide
+
```
LenkaNovak, line 76: I've just found `isinteractive()`, which we can use here as a conditional for running from the REPL, so we can leave this uncommented.

test/TestHelper.jl
```diff
@@ -0,0 +1,84 @@
...
+Create an NCDataset with lat/lon dimensions containing the value `val` for the
+variable `varname`, and store it at `path`.
+
+# Arguments
+- `FT`: [DataType] Float type.
+- `path`: [String] location to store output datafile.
+- `varname`: [Symbol] variable name.
+- `val`: [FT] value to store as `varname` at all indices.
+"""
+function gen_ncdata!(FT, path, varname, val)
```
LenkaNovak, line 58: The bang is not consistent with the docstring. AFAIK the convention is only to use bangs when arguments are mutated, so I would remove it here.

LenkaNovak (approved):
Hi @juliasloan25 , looks great! I've only left a couple of small comments, after resolving which feel free to merge. Thanks!

PR #146, opened 2022-11-01
enable infrequent coupling

LenkaNovak:
## Purpose 

enables `dt_atmos != dt_cpl`

## Content

----


LenkaNovak:
@szy21 the Buildkites are passing as before, so I'm going to merge this. I also added a couple of tests in the slab planet setting so we can keep track of the various options we'd previously considered.

Issue #135, opened 2022-10-21
AMIP improvements


LenkaNovak:
- WIP

High priority
- [x] long Buildkite runs with current packages
  - [x] #143 
  - [x] Artifacts race condition #145 
- [ ] update for latest features when they become available (incl. topog, EDMF etc)
- [ ] address the unrealistic circulation features (@szy21 tagging you here for more pointers as we proceed)
  - [ ] revisit surface flux formulation
  - [ ] effect of sponge on qt in the stratosphere 
- [ ] performance
  - [ ] unnecessary allocs
  - [ ] re-intro flux accumulation for infrequent coupling
    - [ ] need radiative flux accumulation at each timestep in atmos OR the stepping needs to coincide with coupling dt, so that rad fluxes are constant throughout coupling timestep  #146 
- [ ] check that radiation is correct in conservation checker (small divergence, also check mono remapping) - links to [CA#1079](https://github.com/CliMA/ClimaAtmos.jl/issues/1079)
- [ ] demonstrate restart in buildkite 
- [ ] abstract artifacts
- [ ] use `accumulate_fluxes!` from CA when 0.7.1 is available.

Medium priority
- [ ] ozone 
- [ ] use land snow

Low priority
- [ ] bring back dynamical sea ice

Other minor issues:
- [ ] cs needs to be global for the monthly macro - explore other solutions
- [ ] SST_init is modified with SST_info even with deepcopy
- [ ] replace `if mode_name` statements with dispatches, write better abstractions
- [ ] add spin up for AMIP
- [ ] do nothing for the parts of the domain that are masked out - aware of MPI and fracitonal surfaces  
- [ ] add an energy check for MPI runs
- [ ] revamp reinit! (and extra step, which is necessary to set values to the unitialized cache. In `ODE.jl` this is done as part of `reinit!`)
- [ ] add mse tables
- [ ] add perf tracking tools (e.g. flamegraphs etc)

szy21:
Another thing I forgot to mention during the meeting: It would be good to add or bring back restart capabilities.

LenkaNovak:
to be addressed in #215

PR #167, opened 2022-11-21
Conservation checker interface

LenkaNovak:
## Purpose 

Move existing energy check to `src` and extend to water conservation. 

## Content

- [x] `ConservationChecker` module
- [x] move energy checker code to `src`
- [x] extend energy checker to water mass
- [x] add unit tests for accumulator
- [x] add unit tests for visualization
- [x] demonstrate implementation in a new driver (`coupler-driver_modular.jl`)
- [x] add buildkite hook
- [x] add documentation

## Notes
- the energy check plots in Buildkite are identical to those produced using the non-modularized version of the conservation checker, called by `coupler_driver.jl`. 
- conservation is not to the required precision. Investigating this is already a TODO in #135 
- generalizing the contained functions further requires a re-design of our simulation objects. This will be deferred to the upcoming SDI on optimization and performance.

----


src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+)
+    @unpack model_sims, surface_masks = coupler_sim
+    @unpack atmos_sim, land_sim, ocean_sim, ice_sim = model_sims
+    radiation = atmos_sim.integrator.p.radiation_model # TODO: take out of global scope in ClimaAtmos
+    boundary_space = coupler_sim.boundary_space # thin shell approx (boundary_space[z=0] = boundary_space[z_top])
+
+    FT = eltype(coupler_sim.surface_masks.land)
+
+    # save radiation source
+    if radiation != nothing
```
szy21, line 84: We have [this](https://github.com/CliMA/ClimaAtmos.jl/blob/a881905404ee6c50babc14c70ee57a0ece055587/examples/hybrid/callbacks.jl#L172-L187) in ClimaAtmos which integrates fluxes at the boundaries. It might be useful when you clean up the code.

docs/src/conservation.md
```diff
@@ -0,0 +1,16 @@
+# Conservation Checks
+
+If the model is a physically closed system (e.g., in the `slabplanet` configuration with free slip conditions), it should conserve mass (including water), energy and momentum. The conservation checker logs global conservation.
+
+Only energy and water are currently implemented. 
+
+Note that kinetic energy is not included in the calculation of the global energy, reflecting the formulation on `ClimaAtmos`, which assumes that kinetic energy is negligible in comparison with the moist static energy components.
```
szy21, line 7: I think this is a bit confusing because e.g. when you calculate the total energy of atmosphere using `e_tot`, it includes kinetic energy. Maybe rephrase it as something like `Kinetic energy transfer between atmosphere, land, and ocean is not included`?

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+"""
+    surface_water_gain_from_rates(cs)
+
+Determines the total water content gain/loss of a surface from the begining of the simulation based on evaporation and precipitation rates. 
+"""
+function surface_water_gain_from_rates(cs::CoupledSimulation)
+    evaporation = cs.fields.F_E # kg/m^2/s / layer depth
+    precipitation_l = cs.fields.P_liq
+    precipitation_s = cs.fields.P_snow
+    @. (evaporation + precipitation_l + precipitation_s) * cs.Δt_cpl # in kg
```
szy21, line 236:
```suggestion
    @. (evaporation + precipitation_l + precipitation_s) * cs.Δt_cpl # in kg /m^2
```
I think that's the unit? I'm not sure though.

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+    # save sea ice
+    coupler_sim.fields.P_net .-= swap_space!(surface_water_gain_from_rates(coupler_sim), boundary_space) # accumulated surface water gain 
+    if ice_sim !== nothing
+        push!(cc.ρq_tot_seaice, sum(coupler_sim.fields.P_net .* surface_masks.ice))
+    else
+        push!(cc.ρq_tot_seaice, FT(0))
+    end
+
+    # save ocean
+    if ice_sim !== nothing
```
szy21, line 217:
```suggestion
    if ocean_sim !== nothing
```
?

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+    FT = eltype(coupler_sim.surface_masks.land)
+
+    # save atmos
+    push!(cc.ρq_tot_atmos, sum(atmos_sim.integrator.u.c.ρq_tot))
+
+    # save land
+    if land_sim !== nothing
+        water_content =
+            @. (land_sim.integrator.u.bucket.σS + land_sim.integrator.u.bucket.W + land_sim.integrator.u.bucket.Ws) # m^3 water / land area / layer height
+        parent(water_content) .= parent(water_content .* surface_masks.land) * FT(1000)
```
szy21, line 202: It's a bit hard for me to figure out the unit here (without knowing details about the land model). Maybe add some more comments?

szy21 (commented):
Thanks Lenka! I only quickly went through it because it's a large PR:) I left a few comments. Also, I wonder if it makes sense to add mass conservation as well?

szy21 (approved):

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
+#=
+    ConservationChecker
+
+This module contains functions that check global conservation of energy and water (and momentum - TODO).  
+=#
+module ConservationChecker
+
+using ClimaCoupler.Utilities
+using ClimaCore: ClimaCore, Geometry, Meshes, Domains, Topologies, Spaces, Fields, InputOutput
+using ClimaCore.Utilities: half
```
juliasloan25, line 10: I think this is redundant since you already have using `ClimaCoupler.Utilities`

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+    ρq_tot_atmos::A
+    ρq_tot_ocean::A
+    ρq_tot_land::A
+    ρq_tot_seaice::A
+end
+
+"""
+    check_conservation!(coupler_sim::CoupledSimulation)
+
+itertes over all specified conservation checks.
```
juliasloan25, line 54: docstring doesn't match function signature

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+"""
+check_conservation!(coupler_sim::CoupledSimulation, get_slab_energy, get_land_energy) =
+    map(x -> check_conservation!(x, coupler_sim, get_slab_energy, get_land_energy), coupler_sim.conservation_checks)
+
+"""
+        check_conservation!(
```
juliasloan25, line 60: docstring doesn't match function signature

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+        @. cc.ρe_tot_atmos + cc.ρe_tot_ocean + cc.ρe_tot_land + cc.ρe_tot_seaice + cc.toa_net_source + cc.friction_sink
+    # @assert abs((tot[end] - tot[1]) / tot[end]) < 1e-4
+end
+
+"""
+    ke_dissipation(sim)
+
+Calculate the kinetic energy diffipation of a simulation. 
+"""
+function ke_dissipation(sim)
```
juliasloan25, line 161: could be helpful to annotate the input and return types (at least in the docstring)

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+            Geometry.Covariant12Vector.(
+                sim.integrator.p.dif_flux_uₕ_bc.components.data.:1,
+                sim.integrator.p.dif_flux_uₕ_bc.components.data.:2,
+            )
+        )
+    dot.(Geometry.UVVector.(level(sim.integrator.u.c.uₕ, 1)), drag_uv) .* level(sim.integrator.u.c.ρ, 1)
+end
+
+"""
+        check_conservation(
```
juliasloan25, line 173: docstring doesn't match function signature

src/ConservationChecker.jl
```diff
@@ -0,0 +1,331 @@
...
+"""
+function surface_water_gain_from_rates(cs::CoupledSimulation)
+    evaporation = cs.fields.F_E # kg/m^2/s / layer depth
+    precipitation_l = cs.fields.P_liq
+    precipitation_s = cs.fields.P_snow
+    @. (evaporation + precipitation_l + precipitation_s) * cs.Δt_cpl # in kg 
+end
+
+# https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
+ENV["GKSwstype"] = "nul"
```
juliasloan25, line 240: a short comment explaining why this is here could be helpful, then readers can follow the link for more detail

juliasloan25 (approved):

artifacts/Artifacts.toml
```diff
@@ -0,0 +1,8 @@
+[land_mask]
```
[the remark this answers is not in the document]
LenkaNovak: I would have thought it would be good to keep it for tracking changes in the files (each time the file is replaced with a new one of the same name, the `git-tree-sha1 ` changes). I just found [this](https://github.com/JuliaLang/Pkg.jl/issues/1914) issue, which may require using `Tar.tree_hash` to be general to all platforms, but apparently this should have been fixed in Julia 1.6. @charleskawczynski do you have any thoughts? I noticed Atmos doesn't check these in?

docs/src/conservation.md
LenkaNovak: That was rather misleading, thanks for pointing it out! It was meant to refer to the KE contribution being neglected in all diffusive fluxes in the atmos interior, so this is consistent with these BCs. In any case, since we won't be needing this function (except perhaps for diagnostics?) in the foreseeable, I've decided to remove it for now. :P

src/ConservationChecker.jl
LenkaNovak: Actually the half needs to be specifically imported, but line 8 can be removed.

src/ConservationChecker.jl
LenkaNovak: That's great! We can update to this upon the next ClimaAtmos release.

src/ConservationChecker.jl
LenkaNovak: This function is now removed.

src/ConservationChecker.jl
LenkaNovak: Good catch! I missed this after the last rebase! 😳

src/ConservationChecker.jl
LenkaNovak: Sure!

src/ConservationChecker.jl
LenkaNovak: Sure! I can comment more lines. Also, instead of the FT(1000), I'm adding  `ClimaLSM.LSMP.ρ_cloud_liq` which clarifies the units further. The final values are in kg, thanks to ClimaCore's `sum` function doing the volume integral for us (rather than a simple sum).

src/ConservationChecker.jl
LenkaNovak: see above. Now clarified

LenkaNovak:
@szy21 and @juliasloan25 , thank you both for the review! 🚀

LenkaNovak:
> 

Sure, we can add this as an new feature in the next PR. As you say this PR is already long, and its main purpose is to make existing code modular.

artifacts/Artifacts.toml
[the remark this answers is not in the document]
charleskawczynski: Yes, I think the artifacts toml is the intended file to be tracked, so that there's a direct mapping between these hashes and the artifacts used.

We don't in ClimaAtmos, but we probably should.

Also, I haven't seen that issue before, it could be that we somehow need to update ArtifactWrappers to use this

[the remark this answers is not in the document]
LenkaNovak: Thank you, Charlie!

LenkaNovak:
> 

It's really no problem to add it in here temporarily, if it's useful. But in the long term we should probably have check for coupled and standalone runs. I'll put it on the AMIP improvements list and maybe we can reassess in Jan?

PR #175, opened 2022-11-30
Boundary Condition Reader interface

juliasloan25:
For reviewers: This is a large PR, but the majority of the changes are in `src/BCReader.jl`, `test/bcreader_tests.jl`, and `test/mpi_tests/bcreader_mpi_tests.jl`. Also, I created the `src/CallbackManager.jl` file in this PR to simplify compatibility with BCReader, but that module will be documented and tested in a separate PR.

Move functions from `coupler_utils/bcfile_reader.jl` to `src/BCReader.jl` module, and ensure that they are thoroughly commented and tested.

#134 

## To-do

PHASE 2 - testing
- [x] Create test file in `test/`, add to `test/runtests.jl`
- [x] Integration tests as needed
- [x] Add any tests that use MPI in `test/mpi_tests` and in buildkite pipeline

PHASE 3 - cleanup
- [x] Check all imports are necessary
- [x] Formatting
- [x] Read through comments

### DIFF
--- a/docs/src/bcreader.md
+++ b/docs/src/bcreader.md
@@ -1,0 +1,23 @@
+# BCReader
+
+This module coordinates reading of boundary conditions from NetCDF files,
+as well as regridding calls and temporal interpolations from
+monthly to daily intervals.
+
+## BCReader API
+
+```@docs
+ClimaCoupler.BCReader.BCFileInfo
+ClimaCoupler.BCReader.bcfile_info_init
+ClimaCoupler.BCReader.update_midmonth_data!
+ClimaCoupler.BCReader.next_date_in_file
+ClimaCoupler.BCReader.interpolate_midmonth_to_daily
+```
+
+
+## BCReader Internal Functions
+
+```@docs
+ClimaCoupler.BCReader.no_scaling
+ClimaCoupler.BCReader.interpol
+```
--- a/src/BCReader.jl
+++ b/src/BCReader.jl
@@ -1,0 +1,312 @@
+#=
+    BCReader
+
+This module coordinates reading of boundary conditions from NetCDF files,
+as well as regridding calls and temporal interpolations from
+monthly to daily intervals.
+=#
+module BCReader
+
+using ..Utilities, ..Regridder, ..CallbackManager
+using ClimaCore: Fields
+using ClimaComms
+using Dates
+using JLD2
+
+export BCFileInfo, bcfile_info_init, update_midmonth_data!, next_date_in_file, interpolate_midmonth_to_daily
+
+
+"""
+    BCFileInfo
+
+Stores information specific to each boundary condition from a file and each variable.
+
+# Inputs
+    bcfile_dir::b                   # directory of the BC file
+    comms_ctx::X                    # communication context used for MPI
+    hd_outfile_root::S              # filename root for regridded data 
+    varname::V                      # name of the variable
+    all_dates::D                    # vector of all dates contained in the original data file
+    monthly_fields::C               # tuple of the two monthly fields, that will be used for the daily interpolation
+    scaling_function::O             # function that scales, offsets or transforms the raw variable
+    land_mask::M                    # mask with 1 = land, 0 = ocean / sea-ice
+    segment_idx::Vector{Int}        # index of the monthly data in the file
+    segment_idx0::Vector{Int}       # `segment_idx` of the file data that is closest to date0
+    segment_length::Vector{Int}     # length of each month segment (used in the daily interpolation)
+    interpolate_daily::Bool         # switch to trigger daily interpolation
+"""
+struct BCFileInfo{FT, B, X, S, V, D, C, O, M}
+    bcfile_dir::B
+    comms_ctx::X
+    hd_outfile_root::S
+    varname::V
+    all_dates::D
+    monthly_fields::C
+    scaling_function::O
+    land_mask::M
+    segment_idx::Vector{Int}
+    segment_idx0::Vector{Int}
+    segment_length::Vector{Int}
+    interpolate_daily::Bool
+end
+
+BCFileInfo{FT}(args...) where {FT} = BCFileInfo{FT, typeof.(args[1:8])...}(args...)
+
+float_type(::BCFileInfo{FT}) where {FT} = FT
+
+"""
+    no_scaling(field, bcf_info)
+
+Remap the values of a field onto the space of the
+BCFileInfo's land_mask without scaling.
+
+# Arguments
+- `field`: [Fields.Field] contains the values to be remapped.
+- `bcf_info`: [BCFileInfo] contains a land_mask to remap onto the space of.
+"""
+no_scaling(field::Fields.Field, bcf_info::BCFileInfo) = Utilities.swap_space!(field, axes(bcf_info.land_mask))
+
+"""
+    bcfile_info_init(
+        FT,
+        bcfile_dir,
+        datafile_rll,
+        varname,
+        boundary_space,
+        comms_ctx;
+        interpolate_daily = false,
+        segment_idx0 = nothing,
+        scaling_function = no_scaling,
+        land_mask = nothing,
+        date0 = nothing,
+        mono = true,
+    )
+
+Regrids from lat-lon grid to cgll grid, saving the output in a new file,
+and returns the info packaged in a single struct.
+
+# Arguments
+- `FT`: [DataType] Float type.
+- `bcfile_dir`: [String] directory the BC file is stored in.
+- `datafile_rll`: [String] file containing data to regrid.
+- `varname`: [String] name of the variable to be regridded.
+- `boundary_space`: [Spaces.AbstractSpace] the space to which we are mapping.
+- `comms_ctx`: [ClimaComms.AbstractCommsContext] context used for this operation.
+- `interpolate_daily`: [Bool] switch to trigger daily interpolation.
+- `segment_idx0`: [Vector{Int}] index of the file data that is closest to date0.
+- `scaling function`: [Function] scales, offsets or transforms `varname`.
+- `land_mask`: [Fields.field] mask with 1 = land, 0 = ocean / sea-ice.
+- `date0`: [Dates.DateTime] start date of the file data.
+- `mono`: [Bool] flag for monotone remapping of `datafile_rll`.
+
+# Returns
+- BCFileInfo
+"""
+function bcfile_info_init(
+    FT,
+    bcfile_dir,
+    datafile_rll,
+    varname,
+    boundary_space,
+    comms_ctx;
+    interpolate_daily = false,
+    segment_idx0 = nothing,
+    scaling_function = no_scaling,
+    land_mask = nothing,
+    date0 = nothing,
+    mono = true,
+)
+
+    # regrid all times and save to hdf5 files
+    hd_outfile_root = varname * "_cgll"
+    if ClimaComms.iamroot(comms_ctx)
+        Regridder.hdwrite_regridfile_rll_to_cgll(
+            FT,
+            bcfile_dir,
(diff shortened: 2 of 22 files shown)


src/BCReader.jl
```diff
@@ -0,0 +1,310 @@
+#=
+    BCReader
+
+This module coordinates reading of boundary conditions from NetCDF files,
+as well as regridding calls and temporal interpolations from
+monthly to daily intervals.
+=#
+module BCReader
+
+using ClimaCoupler: ClimaCoupler, Utilities, Regridder, CallbackManager
```
charleskawczynski, line 10:
```suggestion
using ..: Utilities, Regridder, CallbackManager
```
This is probably safer (`using ClimaCoupler` can result in using multiple versions of the package, and can cause very confusing errors)

src/BCReader.jl
```diff
@@ -0,0 +1,310 @@
...
+- `date`: [Dates.DateTime] start date for data.
+- `bcf_info`: [BCFileInfo] containing boundary condition data.
+"""
+function update_midmonth_data!(date, bcf_info::BCFileInfo)
+    # monthly count
+    all_dates = bcf_info.all_dates
+    midmonth_idx = bcf_info.segment_idx[1]
+    midmonth_idx0 = bcf_info.segment_idx0[1]
+    outfile_root = bcf_info.hd_outfile_root
+    scaling_function = bcf_info.scaling_function
```
charleskawczynski, line 184:
```suggestion
    (; hd_outfile_root, scaling_function) = bcf_info
```
Note that unpacking can be done like this, too

src/CallbackManager.jl
```diff
@@ -0,0 +1,43 @@
+#=
+    CallbackManager - TODO WIP
+
+This module facilitates calendar functions and temporal interpolations
+of data.
+=#
+
+module CallbackManager
+
+using ClimaCoupler: ClimaCoupler, Utilities
```
charleskawczynski, line 10:
```suggestion
using ..: Utilities
```

test/bcreader_tests.jl
```diff
@@ -0,0 +1,257 @@
...
+            sst_data = AW.get_data_folder(
+                AW.ArtifactWrapper(
+                    @__DIR__,
+                    "sst",
+                    AW.ArtifactFile[AW.ArtifactFile(
+                        url = "https://caltech.box.com/shared/static/8gd3wjq2dbrnzuv8pd5ww3a54h0kkcz8.nc",
+                        filename = "sst.nc",
+                    ),],
+                ),
+            )
```
charleskawczynski, line 217: Note that this can lead to a race condition (if this same code is exercised in different buildkite jobs), since artifacts are saved to a common path. Please see [this PR](https://github.com/CliMA/ClimaAtmos.jl/pull/981) for an example of how this race condition can be avoided.

test/bcreader_tests.jl
```diff
@@ -0,0 +1,308 @@
...
+            # construct weightfile name to test values
+            hd_outfile_root = varname * "_cgll"
+            outfile = hd_outfile_root * ".nc"
+            outfile_root = mono ? outfile[1:(end - 3)] * "_mono" : outfile[1:(end - 3)]
+            weightfile = joinpath(regrid_dir, outfile_root * "_remap_weights.nc")
+            weights = NCDataset(weightfile)["S"]
+
+            # test monotone remapping (all weights in [0, 1])
+            max_weight = maximum(weights)
+            min_weight = minimum(weights)
```
charleskawczynski, line 300:
```suggestion
           nt = NCDataset(weightfile)["S"] do weights

                # test monotone remapping (all weights in [0, 1])
                max_weight = maximum(weights)
                min_weight = minimum(weights)
                (; max_weight, min_weight)
            end
            (; max_weight, min_weight) = nt
```

`NCDataset(filename::AbstractString)` opens a file. I recommend using the `do` syntax to ensure that it's always closed, otherwise it's a memory leak

test/bcreader_tests.jl
```diff
@@ -0,0 +1,308 @@
...
+            monthly_fields,                     # monthly_fields
+            segment_length,                     # segment_length
+            nothing,                            # scaling_function
+            interpolate_daily,                  # interpolate_daily
+            nothing,                            # land_mask
+        )
+        @test BCReader.interpolate_midmonth_to_daily(FT, date0, bcf_info_no_interp) == monthly_fields[1]
+    end
+
+    # Add tests which use TempestRemap here -
```
LenkaNovak, line 118: Don't the tests below use TR?

test/bcreader_tests.jl
```diff
@@ -0,0 +1,308 @@
...
+            topology = Topologies.Topology2D(mesh)
+            quad = Spaces.Quadratures.GLL{Nq}()
+            boundary_space_t = Spaces.SpectralElementSpace2D(topology, quad)
+            current_fields = Fields.zeros(FT, boundary_space_t), Fields.zeros(FT, boundary_space_t)
+
+            land_mask_t = Fields.zeros(boundary_space_t)
+            dummy_data = (; test_data = zeros(axes(land_mask_t)))
+
+            datafile_rll = sst_data
+            varname = "SST"
```
LenkaNovak, line 142: is this using the original file with all of the SST dates? This would make the test quite slow. I wonder if we use a more lightweight version (e.g. save only a couple of the dates in a separate file in the Box as [here](https://github.com/CliMA/ClimaCoupler.jl/wiki/ClimaCoupler-Lessons-Learned#Working-with-external-data-files)).

LenkaNovak (commented):
Looks good! I had a couple of comments and agree with Charlie's suggestions. Also, it would be good to implement this in the `coupler_driver_modular.jl` file once the #167 is merged.

test/bcreader_tests.jl
juliasloan25: This is a great idea. Do you know how we can select a subset of the dates in the file when or after we download it?

test/bcreader_tests.jl
juliasloan25: The two tests below this both call `bcfile_info_init`, which uses TR so it can't be run on windows

test/bcreader_tests.jl
charleskawczynski: cc @juliasloan25

charleskawczynski (commented):
This is a pretty big PR, could it maybe be broken into smaller parts?

Also, I suggest we start using SafeTestsets.jl in the test suite to avoid leaky tests-- this can cause some really unintuitive errors

src/BCReader.jl
```diff
@@ -0,0 +1,309 @@
+#=
+    BCReader
+
+This module coordinates reading of boundary conditions from NetCDF files,
+as well as regridding calls and temporal interpolations from
+monthly to daily intervals.
+=#
```
charleskawczynski, line 7: Just FYI, this can be a doc string (`#= this is a multi-line comment =#`

src/BCReader.jl
```diff
@@ -0,0 +1,309 @@
...
+    varname::V
+    all_dates::D
+    segment_idx::Vector{Int}
+    segment_idx0::Vector{Int}
+    monthly_fields::C
+    segment_length::Vector{Int}
+    scaling_function::O
+    interpolate_daily::Bool
+    land_mask::M
+end
```
charleskawczynski, line 53:
```suggestion
struct BCFileInfo{FT, B, X, S, V, D, C, O, M}
    bcfile_dir::B
    comms_ctx::X
    hd_outfile_root::S
    varname::V
    all_dates::D
    segment_idx::Vector{Int}
    segment_idx0::Vector{Int}
    monthly_fields::C
    segment_length::Vector{Int}
    scaling_function::O
    interpolate_daily::Bool
    land_mask::M
end
float_type(::BCFileInfo{FT}) where {FT} = FT
```
Storing types in struct fields is actually a performance issue-- the `typeof(Float64)` is a `DataType`, and the compiler cannot optimize fields of this type.

By chance, was this pattern adopted from somewhere else?

juliasloan25: It was - I'll open a little PR to address all the places in the code this style is used

juliasloan25: It seems like with the new style, no default constructors are created for the struct. I created a MWE 
`struct Foo{FT, S}
       val::S
       end`, which then gives `methods(Foo) 0 methods for type constructor`.
Is there an easy way around this? I could write explicit constructors for all the structs, but they're pretty large.

src/BCReader.jl
```diff
@@ -0,0 +1,309 @@
...
+export BCFileInfo, bcfile_info_init, update_midmonth_data!, next_date_in_file, interpolate_midmonth_to_daily
+
+
+"""
+    BCFileInfo
+
+Stores information specific to each boundary condition from a file and each variable.
+
+# Inputs
+    FT::F                           # float type
```
valeriabarra, line 25: If you use Charlie's suggestion, remember to update the docstring :)

src/CallbackManager.jl
```diff
@@ -0,0 +1,43 @@
...
+=#
+
+module CallbackManager
+
+using ..Utilities
+using Dates
+
+export current_date, strdate_to_datetime, datetime_to_strdate, calendar_callback
+
+# TODO remove calls to calendar_callback - can be replaced with `<`
```
valeriabarra, line 15: Is this TODO for a later PR or the comment is outdated? I don't see calls to `calendar_callback` in this file

juliasloan25: This `TODO` is for a later PR when I go through the `CallbackManager` module in more detail. I just put the reminder here since it's where `calendar_callback` is defined.

src/BCReader.jl
juliasloan25: Will do!

src/BCReader.jl
charleskawczynski: There are a few options, one is to make a generic-like outer constructor:
```
julia> struct Foo{FT, S} val::S end

julia> Foo{FT}(args...) where {FT} = Foo{FT, typeof.(args)...}(args...)

julia> Foo{Float64}(1)
Foo{Float64, Int64}(1)
```
Another option is basically that with the use of the `Base.@kwdef` for adding a keyword constructor

docs/src/bcreader.md
```diff
@@ -0,0 +1,23 @@
...
+    ClimaCoupler.BCReader.interpolate_midmonth_to_daily
+```
+
+
+## BCReader Internal Functions
+
+```@docs
+    ClimaCoupler.BCReader.no_scaling
+    ClimaCoupler.BCReader.interpol
+```
```
charleskawczynski, line 23:
```suggestion
```@docs
ClimaCoupler.BCReader.BCFileInfo
ClimaCoupler.BCReader.bcfile_info_init
ClimaCoupler.BCReader.update_midmonth_data!
ClimaCoupler.BCReader.next_date_in_file
ClimaCoupler.BCReader.interpolate_midmonth_to_daily
```


## BCReader Internal Functions

```@docs
ClimaCoupler.BCReader.no_scaling
ClimaCoupler.BCReader.interpol
```
```

Comment (charleskawczynski) [reply docs/src/bcreader.md:23]: Oh, not sure my suggestion worked, basically I was suggesting to remove prepending spaces

File: src/BCReader.jl [hunk 13]
```diff
@@ -0,0 +1,309 @@
...
+# Returns
+- BCFileInfo
+"""
+function bcfile_info_init(
+    FT,
+    bcfile_dir,
+    datafile_rll,
+    varname,
+    boundary_space;
+    comms_ctx = ClimaComms.SingletonCommsContext(),
```
charleskawczynski, line 108:
```suggestion
    comms_ctx,
```
I think that creating separate `comms_ctx` may lead to strange mpi errors, so I'd suggesting making this a required argument

PR #191, opened 2022-12-13
NCDataset memory leaks fixed

juliasloan25:
## Purpose 
Remove memory leaks caused by opening NCDatasets without closing them.

Update: accomplished by #195 - this PR is no longer necessary

----

### DIFF
--- a/experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/coupler_utils/regridder.jl
+++ b/experiments/AMIP/moist_mpi_earth_dynamical_sea_ice/coupler_utils/regridder.jl
@@ -45,9 +45,9 @@ function ncreader_rll_to_cgll(
 )
     isdir(REGRID_DIR) ? nothing : mkpath(REGRID_DIR)
 
-    ds = NCDataset(datafile_rll)
-    nlat = ds.dim["lat"]
-    nlon = ds.dim["lon"]
+    nlat, nlon = NCDataset(datafile_rll) do ds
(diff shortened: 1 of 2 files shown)


charleskawczynski (approved):

.buildkite/pipeline.yml
```diff
@@ -144,3 +144,4 @@ steps:
           CLIMACORE_DISTRIBUTED: "MPI"
         agents:
           slurm_ntasks: 32
+          slurm_mem: 80GB
```
charleskawczynski, line 147: Should this be done separately?

juliasloan25: Yeah, probably. I just didn't want to merge with main before all the CI tests were passing, but I guess `MPI AMIP FINE` is already failing in main too.

charleskawczynski: Maybe just do this in a separate PR to fix CI (see the [latest build](https://buildkite.com/clima/climacoupler-ci/builds/586#0184fa24-d737-4ab9-8a92-4e053d5d4f70)). It's a bit striking that we need this much memory, cc @simonbyrne

charleskawczynski: Yeah, I think we should first focus on fixing CI, then these other changes should be able to merge in smoothly

simonbyrne: I'm not sure what the issue is, but do you open the netcdf file on every MPI process?

PR #197, opened 2022-12-19
Use land albedo from file, update packages

juliasloan25:
## Purpose 
ClimaLSM has been updated to read in surface albedo from a data file, rather than passing in a function for it as before (see ClimaLSM [PR #125](https://github.com/CliMA/ClimaLSM.jl/pull/125)). The coupler is updated to have a command line option `albedo_from_file` to specify whether we want to read in the albedo from a data file, with the default being `true`. The code relating to the bucket model in the coupler is also updated to reflect the changes in ClimaLSM.

This PR also updates ClimaCoupler to use the most recent package versions, notably of ClimaAtmos, SurfaceFluxes, and ClimaLSM.

## A Note on Buildkite Changes
After making these changes, one buildkite test fails to converge (`target_params_in_slab`). When the same test is set up with prescribed albedo, as was previously the default, it runs successfully. Because of this, we think the failure is due to the more realistic albedo from the input file causing too sharp of a contrast between land and ocean. For now, we will address this by changing this buildkite run to use prescribed albedo rather than albedo from a file.
See relevant plots [here](https://docs.google.com/document/d/17spTazJ0z45Qak8yCpVYaBUxc5Xz7nHxiOwM-T3d9NI/edit?usp=sharing).

## Albedo Plots
Here, you can see the surface albedo stored in the atmos simulation when albedo is read in from a file vs. when albedo is prescribed via a function. The difference between these plots shows us that the atmosphere is actually getting updated when we read in albedo from a file. Note that in each case, the "direct" and "diffuse" albedos were identical in the atmosphere, so we just show direct albedo here.

direct_albedo-from_file
direct_albedo-from_function

----

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/Manifest.toml
+++ b/experiments/AMIP/moist_mpi_earth/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.8.3"
 manifest_format = "2.0"
-project_hash = "f2fc4f446f0d9872ae3e86d9183faa9636191e76"
+project_hash = "ec523edebb9d4184491441617975d3204cb731e3"
 
 [[deps.AMD]]
 deps = ["Libdl", "LinearAlgebra", "SparseArrays", "Test"]
@@ -17,15 +17,15 @@ uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 version = "1.2.1"
 
 [[deps.AbstractTrees]]
-git-tree-sha1 = "52b3b436f8f73133d7bc3a6c71ee7ed6ab2ab754"
+git-tree-sha1 = "faa260e4cb5aba097a73fab382dd4b5819d8ec8c"
 uuid = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
-version = "0.4.3"
+version = "0.4.4"
 
 [[deps.Adapt]]
 deps = ["LinearAlgebra"]
-git-tree-sha1 = "195c5505521008abea5aee4f96930717958eac6f"
+git-tree-sha1 = "0310e08cb19f5da31d08341c6120c047598f5b9c"
 uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
-version = "3.4.0"
+version = "3.5.0"
 
 [[deps.ArgParse]]
 deps = ["Logging", "TextWrap"]
@@ -44,16 +44,16 @@ uuid = "ec485272-7323-5ecc-a04f-4719b315124d"
 version = "0.2.0"
 
 [[deps.ArrayInterface]]
-deps = ["ArrayInterfaceCore", "Compat", "IfElse", "LinearAlgebra", "Static"]
-git-tree-sha1 = "d6173480145eb632d6571c148d94b9d3d773820e"
+deps = ["ArrayInterfaceCore", "Compat", "IfElse", "LinearAlgebra", "SnoopPrecompile", "Static"]
+git-tree-sha1 = "dedc16cbdd1d32bead4617d27572f582216ccf23"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "6.0.23"
+version = "6.0.25"
 
 [[deps.ArrayInterfaceCore]]
-deps = ["LinearAlgebra", "SparseArrays", "SuiteSparse"]
-git-tree-sha1 = "c46fb7dd1d8ca1d213ba25848a5ec4e47a1a1b08"
+deps = ["LinearAlgebra", "SnoopPrecompile", "SparseArrays", "SuiteSparse"]
+git-tree-sha1 = "e5f08b5689b1aad068e01751889f2f615c7db36d"
 uuid = "30b0a656-2188-435a-8636-2ec0e6a096e2"
-version = "0.1.26"
+version = "0.1.29"
 
 [[deps.ArrayInterfaceGPUArrays]]
 deps = ["Adapt", "ArrayInterfaceCore", "GPUArraysCore", "LinearAlgebra"]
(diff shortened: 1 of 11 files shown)


experiments/AMIP/moist_mpi_earth/Project.toml
```diff
@@ -36,6 +36,7 @@ Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 RRTMGP = "a01a1ee8-cea4-48fc-987c-fc7878d79da1"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
```
kmdeck, line 39: if you ended up not using this, could remove

experiments/AMIP/moist_mpi_earth/Manifest.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.2"
+julia_version = "1.8.1"
```
kmdeck, line 3: not sure if you wanted to revert to 1.8.1?

experiments/AMIP/moist_mpi_earth/bucket/bucket_init.jl
```diff
@@ -4,17 +4,20 @@ using ClimaLSM
...
     liquid_precipitation,
-    BulkAlbedo,
+    BulkAlbedoMap,
+    BulkAlbedoFunction,
     surface_albedo,
     snow_precipitation
 
-using ClimaLSM: make_ode_function, initialize, obtain_surface_space
+using ClimaLSM: make_ode_function, initialize, obtain_surface_space, make_set_initial_aux_state
+import ClimaLSM.Regridder: MapInfo
```
kmdeck, line 20: I think this is not used now, with the default

experiments/AMIP/moist_mpi_earth/coupler_driver.jl
```diff
@@ -93,8 +100,23 @@ import ClimaCoupler
 pkg_dir = pkgdir(ClimaCoupler)
 COUPLER_OUTPUT_DIR = joinpath(pkg_dir, "experiments/AMIP/moist_mpi_earth/output", joinpath(mode_name, run_name))
 !isdir(COUPLER_OUTPUT_DIR) && mkpath(COUPLER_OUTPUT_DIR)
-REGRID_DIR = joinpath(COUPLER_OUTPUT_DIR, "regrid_tmp/")
+
+# create unique REGRID_DIR for each buildkite job
+if "BUILDKITE" in keys(ENV)
+    if ClimaComms.iamroot(comms_ctx)
+        REGRID_DIR = joinpath(COUPLER_OUTPUT_DIR, "regrid_tmp/", ENV["BUILDKITE_STEP_KEY"])
+    else
```
kmdeck, line 108: do you need this `else`? if it isnt the root process, is REGRID_DIR = "" used/should it be used?

not sure I follow this part.

kmdeck (approved):
looks good! small comments only

experiments/AMIP/moist_mpi_earth/coupler_driver.jl
juliasloan25: If it isn't the root process, it will receive the `REGRID_DIR` defined on the root process. I added this `else` to set `REGRID_DIR=""` just so the variable is initialized in both branches, but the empty string will be overwritten by the broadcast on line 113.

experiments/AMIP/moist_mpi_earth/push_pull.jl
```diff
@@ -154,16 +149,38 @@ function atmos_pull!(cs)
...
-        parent(atmos_sim.integrator.p.z0b) .= parent(z0b_cpl)
-        parent(atmos_sim.integrator.p.z0m) .= parent(z0m_cpl)
+    if :z0b in propertynames(atmos_sim.integrator.p.surface_scheme)
+        parent(atmos_sim.integrator.p.sfc_inputs.z0b) .= parent(z0b_cpl)
+        parent(atmos_sim.integrator.p.sfc_inputs.z0m) .= parent(z0m_cpl)
     end
 
-    calculate_surface_fluxes_atmos_grid!(atmos_sim.integrator, info_sfc)
+    Fields.bycolumn(axes(atmos_sim.integrator.p.ts_sfc)) do colidx
+        ClimaAtmos.set_surface_thermo_state!(
```
LenkaNovak, line 159: This is assuming a saturated surface. We will need to:
- 1. re-enable setting `ρ_sfc` from the coupler (@szy21 was there a plan to do this? We can also extend the `surface_thermo_state` function if not. But even for the stand-alone runs you may want to not be bound to surface saturation?). Both land and ice surfaces will rarely be saturated. 
- 2. cap evaporation over ice, as was previously done in the `correct_e_over_ice` function in the coupler. 
To avoid scope creep, let's merge this but address the surface saturation ASAP. #237

LenkaNovak (approved):
Looks good, Julia! Let's take a look at the surface saturation issue together as part of the AMIP physics refinement / reporting during the next sprint cycle.

experiments/AMIP/moist_mpi_earth/push_pull.jl
szy21: Do you mean setting `q_sfc` from the coupler? In any case, I don't think there is a plan to do this. We used to have to option to set both `ρ_sfc` and `q_sfc`, but then it was refactored to set `ts_sfc`. In the latest release of ClimaAtmos we have `set_surface_thermo_state!(::Coupled, args...) = nothing`, so I would assume it will use the `ts_sfc` set in the coupler.

LenkaNovak: Yes, and ok, no probs. We'll make our own `set_surface_thermo_state!` extension. Thanks!

experiments/AMIP/moist_mpi_earth/bucket/bucket_init.jl
```diff
@@ -144,13 +149,18 @@ function bucket_init(
...
-        (; lat, long) = coordinate_point
-        return typeof(lat)(0.4)
-    end
 
     α_snow = FT(0.8) # snow albedo
-    albedo = BulkAlbedo{FT}(α_snow, α_soil)
+    if albedo_from_file # Read in albedo from data file (default)
+        albedo = BulkAlbedoMap{FT}(regrid_dirpath, α_snow = α_snow, comms = comms_ctx)
+    else # Use spatially-varying function for surface albedo
+        function α_sfc(coordinate_point)
```
kmdeck, line 157: one question! If the prescribed albedo function is constant, why is it varying so much in the plot? Are the non 0.4 values due to the ocean?

With snow on the surface, we do a linear interpolation between the albedo of ground and albedo of snow, which could cause variation, but I believe there is no snow being modeled yet (P_snow =0), and this would only increase the albedo from the base value of 0.4.

kmdeck: Instead of plotting SW albedo vs index, perhaps we could plot it in terms of lat/lon?

PR #204, opened 2023-01-03
update modular driver to use BCReader module

juliasloan25:
## Purpose 
The BCReader module was implemented by #175. Now the modular driver `coupler_driver_modular.jl` needs to be updated to use this instead of including `coupler_utils/bcfile_reader.jl`.

## To-do
- [x] Replace `include` line with import of BCReader module
- [x] Update calls of functions from BCReader

----

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
@@ -36,7 +36,7 @@ end
 
 BCFileInfo{FT}(args...) where {FT} = BCFileInfo{FT, typeof.(args[1:7])...}(args...)
 
-float_type(::BCFileInfo{FT}) where {FT} = FT
+float_type_bcf(::BCFileInfo{FT}) where {FT} = FT
 
 """
     bcfile_info_init(FT, comms_ctx, datafile_rll, varname, boundary_space; interpolate_daily = false, segment_idx0 = [Int(1)], scaling_function = false)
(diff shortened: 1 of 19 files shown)


LenkaNovak (approved):
Looks good, thanks @juliasloan25 !

LenkaNovak:
Actually, I've just noticed that the Buildkite test is only running this driver for the `slabplanet` mode, so the BC functions are not being called. It would be good to add another Buildkite run of this driver in the `AMIP` mode.

experiments/AMIP/moist_mpi_earth/coupler_utils/bcfile_reader.jl
```diff
@@ -36,7 +36,7 @@ end
 
 BCFileInfo{FT}(args...) where {FT} = BCFileInfo{FT, typeof.(args[1:7])...}(args...)
 
-float_type(::BCFileInfo{FT}) where {FT} = FT
+float_type_bcf(::BCFileInfo{FT}) where {FT} = FT
```
LenkaNovak, line 39: Nice!

LenkaNovak (approved):
We now have more stringent criteria for codecov, so the test is failing in `Utilities'jl`. I would delete the `heaviside` function. Looks like we don't need it anymore. Hopefully that'll help it pass.

juliasloan25:
> We now have more stringent criteria for codecov, so the test is failing in `Utilities'jl`. I would delete the `heaviside` function. Looks like we don't need it anymore. Hopefully that'll help it pass.

@LenkaNovak even after removing the `heaviside` function from `Utilities.jl`, codecov still fails. Should I add tests for the functions in `Utilities.jl`, or just merge it as-is?

PR #209, opened 2023-01-05
TimeManager interface module

juliasloan25:
## Purpose 
Move functions from coupler_utils/calendar_timer.jl to src/CallbackManager.jl module, and ensure that they are thoroughly commented and tested.
Part of SDI #134 

## To-do

PHASE 2 - testing
- [x] Create test file in `test/`, add to `test/runtests.jl`

PHASE 3 - cleanup
- [x] Formatting
- [x] Read through comments

----


src/TimeManager.jl
```diff
@@ -0,0 +1,52 @@
...
+"""
+    current_date(cs, t)
+
+Return the model date at the current timestep
+
+# Arguments
+- `cs`: [CoupledSimulation] or [CouplerSimulation] containing info about the simulation
+- `t`: [Integer] number of seconds since simulation began
+"""
+current_date(cs, t) = cs.dates.date0[1] + Dates.Second(t)
```
LenkaNovak, line 25: Please specify the argument types here.

src/TimeManager.jl
```diff
@@ -0,0 +1,52 @@
...
+required by the official AMIP input files
+
+# Arguments
+- `strdate`: [String] to be converted to Date type 
+"""
+strdate_to_datetime(strdate::String) =
+    Dates.DateTime(parse(Int, strdate[1:4]), parse(Int, strdate[5:6]), parse(Int, strdate[7:8]))
+
+"""
+    datetime_to_strdate(datetime)
```
LenkaNovak, line 40: Could add the `::DateTime` here too

test/time_manager_tests.jl
```diff
@@ -0,0 +1,48 @@
...
+        (;), # monthly_3d_diags
+        (;), # monthly_2d_diags
+    )
+
+    for t in ((tspan[1] + Δt_cpl):Δt_cpl:tspan[end])
+        @test TimeManager.current_date(cs, t) == date0 + Dates.Second(t)
+    end
+end
+
+@testset "test strdate_to_datetime" begin
```
LenkaNovak, line 38: What's the advantage of testing multiple cases vs just one?

LenkaNovak (approved):
Nice job! Feel free to merge after addressing the minor comments.

test/time_manager_tests.jl
juliasloan25: I thought it would be good to test the min and max values the string/date could have. At least for the case where the year doesn't have 4 digits, it tests that the year is correctly padded with 0s. I think we could remove the third case for the testsets without losing any information though.

src/TimeManager.jl
juliasloan25: Is there a better way I should specify the types than `[CoupledSimulation] or [CouplerSimulation]` and `[Integer]`? I wasn't sure how to be more specific since there are multiple valid types.

test/time_manager_tests.jl
LenkaNovak: Ok, sounds good. Good thinking on testing case margins in general though!

PR #236, opened 2023-01-25
Add flamegraphs

LenkaNovak:
## Purpose 

Addresses the first component of #229

## To-do

- [x] transfer/modify infrastructure from ClimaAtmos/ClimaCore (by @charleskawczynski )
- [x] setup benchmarks for allocs test
- [x] implement in Buildkite - [here](https://buildkite.com/clima/climacoupler-ci/builds/812#0185f592-2b5f-4c90-ab53-b1719b348709)
- [x] add docs to flame.jl and to Slurm-Buildkite Wiki - [here](https://github.com/CliMA/slurm-buildkite/wiki/Flame-Graphs)

----


.buildkite/pipeline.yml
```diff
@@ -158,10 +162,29 @@ steps:
...
+        command: "mpiexec julia --color=yes --project=experiments/AMIP/moist_mpi_earth/ experiments/AMIP/moist_mpi_earth/coupler_driver.jl --run_name target_amip_n32_shortrun --coupled true --surface_scheme monin_obukhov --moist equil --vert_diff true --rad allskywithclear --z_elem 50 --dz_top 3000 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --rayleigh_sponge true --alpha_rayleigh_uh 0 --alpha_rayleigh_w 10 --dt_cpl 150 --dt 150secs --dt_rad 1hours --idealized_insolation false --FLOAT_TYPE Float64 --energy_check false --mode_name amip --t_end 5days --dt_save_to_sol 10days --mono_surface false --precip_model 0M"
         artifact_paths: "experiments/AMIP/moist_mpi_earth/output/amip/target_amip_n32_shortrun_artifacts/*"
         env:
           CLIMACORE_DISTRIBUTED: "MPI"
         agents:
           slurm_ntasks: 32
           slurm_mem_per_cpu: 16G
+
+      # performance tests
+      - label: ":rocket: performance: flame graph and allocation tests"
```
juliasloan25, line 174: it might be helpful each of the performance tests has a unique `label` and `artifact_paths` so we can tell them apart more easily

experiments/AMIP/moist_mpi_earth/coupler_driver_modular.jl
```diff
@@ -320,7 +324,7 @@ cs = CoupledSimulation{FT}(
     coupler_fields,
     parsed_args,
     conservation_checks,
-    tspan,
+    [tspan[1], tspan[2]],
```
juliasloan25, line 327: would it be simpler to initialize `tspan` as an array of length 2 instead? i.e. `tspan = [Int(0), t_end]`, then we can just pass in `tspan` here.

juliasloan25 (approved):
looks good to me!

experiments/AMIP/moist_mpi_earth/coupler_driver_modular.jl
LenkaNovak: Doing it this way allows specifying tspan as a Tuple (like in Atmos's global arguments which overwrites our tspan) or as an array. I know... and I agree that we should unify this, once we don't have synonymous global variables lingering about in the background. :)

PR #241, opened 2023-01-31
Add compats for Plots, CCore, CAtmos, CComms

juliasloan25:
## Purpose 
Address CompatHelper PRs #186, #226, #233, #239

Note: ClimaCore has released v0.10.25, which changes the DSS interface. This PR restricts the ClimaCore version to v0.10.24, but the compat should be changed after ClimaCoupler and ClimaLSM are updated to use the new DSS functions.

## Content
Add compat entries for Plots, ClimaCore, ClimaAtmos, and ClimaComms. Keep any existing compat entries. 

----


LenkaNovak (commented):
Hi @juliasloan25, it looks like there are still some inconsistencies with `dss_interior_faces!` in the Buildkite runs?

LenkaNovak (approved):
Looks like this passes, and awesome that you already tagged an issue for the ClimaCore0.10.25 update. Thanks, Julia!

PR #249, opened 2023-02-06
Make `swap_spaces` not allocating and more GPU compatible

### LINKED ISSUE #248: Make `swap_space` as not allocating and more GPU compatible
This is part of #205 

## Contents
We want to not use `parent` in `swap_spaces`. By doing so we have the following benefits:
- It will be not allocating
- It will be more GPU compatible

valeriabarra:
## Purpose 
The purspose of this PR is to improve `swap_space` so that it is not allocating and to make it GPU compatible.  

Closes #248 

## To-do
- [x] ~~Remove internal call to ClimaCore `parent`~~ This cannot be done internally in the function (see error: `ERROR: LoadError: syntax: Global method definition around .... needs to be placed at the top level, or use "eval".` because of the global method table). So it needs to be worked around wherever it's used.
- [x] Make it `@inline`?

## Content
We remove the internal usage of ClimaCore `parent` by using ClimaCore's `Fields.allow_mismatched_diagonalized_spaces()` function as a temporary workaround, in the absence of a better ClimaCore's interface 

Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

In the Content, I have included 
- relevant unit tests, and integration tests, 
- appropriate docstrings on all functions, structs, and modules, and included relevant documentation.

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/flux_calculator.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/flux_calculator.jl
@@ -5,7 +5,7 @@ import ClimaAtmos: get_surface_fluxes!
 """
     set_ρ_sfc!(ρ_sfc, T_S, integrator)
 
-sets the value of the ρ_sfc field based on the temperature of the surface, 
+sets the value of the ρ_sfc field based on the temperature of the surface,
 the temperature of the atmosphere at the lowest level, and the heigh
 of the lowest level.
 """
@@ -13,7 +13,7 @@ function set_ρ_sfc!(ρ_sfc, T_S, integrator)
     ts = integrator.p.ᶜts
     thermo_params = CAP.thermodynamics_params(integrator.p.params)
     ts_int = Spaces.level(ts, 1)
-    parent(ρ_sfc) .= parent(ρ_sfc_at_point.(thermo_params, ts_int, swap_space!(T_S, axes(ts_int))))
+    parent(ρ_sfc) .= parent(ρ_sfc_at_point.(thermo_params, ts_int, swap_space!(zeros(axes(ts_int)), T_S)))
 end
 
 """
@@ -23,7 +23,7 @@ Computes the surface density at a point given the atmospheric state
 at the lowest level, the surface temperature, and the assumption of
 an ideal gas and hydrostatic balance.
 
-Required because the surface models do not compute air density as a 
+Required because the surface models do not compute air density as a
 variable.
 """
 function ρ_sfc_at_point(params, ts_int, T_sfc)
@@ -37,9 +37,9 @@ end
 """
     calculate_surface_fluxes_atmos_grid!(integrator)
 
-Calculates surface fluxes using adapter function `get_surface_fluxes!` 
-from ClimaAtmos that calls `SurfaceFluxes.jl`. The coupler updates in 
-atmos model cache fluxes at each coupling timestep. 
+Calculates surface fluxes using adapter function `get_surface_fluxes!`
+from ClimaAtmos that calls `SurfaceFluxes.jl`. The coupler updates in
+atmos model cache fluxes at each coupling timestep.
 
 - TODO: generalize interface for regridding and take land state out of atmos's integrator.p
 """
--- a/experiments/AMIP/moist_mpi_aquaplanet/coupler_driver.jl
+++ b/experiments/AMIP/moist_mpi_aquaplanet/coupler_driver.jl
@@ -14,7 +14,7 @@ include("coupler_utils/conservation_checker.jl")
 include("coupler_utils/regridder.jl")
 
 # initiate spatial and temporal info
-t_end = 2592000 # 100e2 
+t_end = 2592000 # 100e2
 tspan = (0, t_end)
 Δt_cpl = 2e2
 saveat = Δt_cpl * 1000
@@ -51,7 +51,7 @@ walltime = @elapsed for t in (tspan[1]:Δt_cpl:tspan[end])
     parent(atmos_sim.integrator.p.dif_flux_energy) .= FT(0)
     calculate_surface_fluxes_atmos_grid!(atmos_sim.integrator, T_S)
 
-    # run 
+    # run
     step!(atmos_sim.integrator, t - atmos_sim.integrator.t, true) # NOTE: use (t - integ_atm.t) here instead of Δt_cpl to avoid accumulating roundoff error in our timestepping.
 
     ## Slab
--- a/experiments/AMIP/moist_mpi_earth/coupler_driver.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_driver.jl
@@ -5,16 +5,16 @@ include("mpi/mpi_init.jl") # setup MPI context for distributed runs #hide
 #=
 ## Overview
 
-AMIP is a standard experimental protocol of the Program for Climate Model Diagnosis & Intercomparison (PCMDI). 
+AMIP is a standard experimental protocol of the Program for Climate Model Diagnosis & Intercomparison (PCMDI).
 It is used as a model benchmark for the atmospheric and land model components, while sea-surface temperatures (SST) and sea-ice concentration (SIC)
 are prescribed using time-interpolations between monthly observed data. We use standard data files with original sources:
-- SST and SIC: https://gdex.ucar.edu/dataset/158_asphilli.html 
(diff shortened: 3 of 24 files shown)


charleskawczynski:
Hi @valeriabarra,

To clarify, `parent` on ClimaCore Fields is not allocating, `parent` returns a reference to the underlying array. Ideally, we would still like to avoid using `parent` so that users do not write code that depends on the underlying data layout, because doing so may make such code incompatible if they were to change the underlying Space.

Calling `allow_mismatched_diagonalized_spaces` is really a kluge, and we shouldn't use it like this here (overloading the method table at runtime like this is generally not a good idea since it will trigger recompilation). If we can't simply remove the call to `parent`, then I suggest we revert the `parent`-related changes from the PR (the SDI says that parent related changes are potential future improvements anyway). We already have an issue in ClimaCore for this problem: https://github.com/CliMA/ClimaCore.jl/issues/1120, so no need to open one. I assume that simply applying these changes didn't work, in which case, let's revert the parent-related changes and apply the allocation-related changes needed.

`swap_space!` is allocating because of this line: https://github.com/CliMA/ClimaCoupler.jl/blob/b50046f8727925381b16783e590536f1dca6da93/src/Utilities.jl#L73, which allocates a field. It turns out that this is actually calling `zeros` defined on a ClimaCore space, so it may already be (or can be made on the ClimaCore-side) GPU-compatible-- but it's still allocating. I suggest you try to make `swap_space!` a 3-argument function, and pass `field_out` as an argument (and then make required changes to use this new 3-argument function). Doing so will allow us to push `field_out` outside of any loops and avoid repeated allocations.

valeriabarra:
Thank you so much for the thorough suggestion. It was previously suggested to me to use this `allow_mismatched_diagonalized_spaces` work-around and eliminate the internal call to `parent`, but yes, I recognize that this was a clumsy work-around and we should proceed the way you suggested, which is much better. I will apply your suggestions as soon as possible.

valeriabarra:
Hey @charleskawczynski I applied the suggestions. I made `swap_space` a 2-argument function (I really think 3 were not needed, as the space was passed just to create the zero field inside). This way, we just move the allocation from inside the `swap_space` to outside of it. There is no allocation gain, and only marginal flame graphs improvement. There are not explicit for loops where  `swap_space` is called, so that the output field can be allocated only once out of the loop. But it is implicitly called by the coupling time stepping. Unless we want to pass the `field_out` all the way from the time-stepping routine, I am not sure how the proposed changes would improve things. 

Can you please confirm that what I implemented aligns with what you had in mind? Or clarify otherwise? Thanks!

experiments/AMIP/moist_mpi_earth/coupler_utils/flux_calculator.jl
```diff
@@ -5,15 +5,15 @@ import ClimaAtmos: get_surface_fluxes!
...
+sets the value of the ρ_sfc field based on the temperature of the surface,
 the temperature of the atmosphere at the lowest level, and the heigh
 of the lowest level.
 """
 function set_ρ_sfc!(ρ_sfc, T_S, integrator)
     ts = integrator.p.ᶜts
     thermo_params = CAP.thermodynamics_params(integrator.p.params)
     ts_int = Spaces.level(ts, 1)
-    parent(ρ_sfc) .= parent(ρ_sfc_at_point.(thermo_params, ts_int, swap_space!(T_S, axes(ts_int))))
+    parent(ρ_sfc) .= parent(ρ_sfc_at_point.(thermo_params, ts_int, swap_space!(zeros(axes(ts_int)), T_S)))
```
LenkaNovak, line 16: Instead of creating a new field here, can we create it as part of the `CouplerSimulation` struct? Similar to how we treat `boundary_space`. @charleskawczynski is this what you meant?

LenkaNovak (changes requested):
Thanks @valeriabarra . I left a comment on passing a `boundary_field` like we do with `boundary_space` right now. Hopefully that'll help with allocations. It would be good to demonstrate on a minimal working example (MWE) the approaches you tried to tackle this problem (as far as I know there is 1. the flag turning off the instance error, 2. creating a field dynamically as the swap_space is being called, and 3. creating a dummy field similar to the `boundary_space` as mentioned above). Since we are kind of blocked by this long-standing issue of mismatching ClimaCore spaces, we may need to defer this issue (hopefully it will be solved soon, now that it's tracked by https://github.com/CliMA/ClimaCore.jl/issues/1120). That said, it would be good to justify this with the MWE.

experiments/AMIP/moist_mpi_earth/coupler_utils/flux_calculator.jl
valeriabarra: It actually seems like Case 2 in the gist is the best option for our purposes, so I will leave the current version of the code. I am going to squash/rebase and you're good to merge. Thanks for the review!

valeriabarra: Output of the gist at first run:
```julia
[ Info: Case 1:
alloc_u_ocn = 19493429
u_ocn = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
alloc_u_ice = 0
u_ice = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
[ Info: Case 2:
alloc_u_ocn = 12800
u_ocn = Float64-valued Field:
  [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
alloc_u_ice = 12800
u_ice = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
[ Info: Case 3:
alloc_u_ocn = 33705
u_ocn = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
alloc_u_ice = 1008
u_ice = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
[ Info: Case 4:
alloc_u_ocn = 465631
u_ocn = Float64-valued Field:
  [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
alloc_u_ice = 1008
u_ice = Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
Float64-valued Field:
  [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0  …  2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
```

valeriabarra:
bors try

LenkaNovak (approved):
Thanks, Valeria. After adjusting the alloc limit, feel free to merge.

valeriabarra:
> Thanks, Valeria. After adjusting the alloc limit, feel free to merge.

Thank you! Just to clarify, the alloc limit adjustment is not due to this PR but to upstream changes. Thanks, will do.

Issue #205, opened 2023-01-04
Some improvements to interfaces and performance


valeriabarra:
## Purpose
This SDI proposes some clean-up of some interfaces in the current ClimaCoupler API so to improve on some of its aspects, such as type-safety and overall compiler optimization and performance.

## Cost/Benefits/Risks
**Costs** 
The costs could be shared across multiple co-authors that can concurrently work on separate files following the same improvement patterns or on the same pattern across different files. 

**Benefits**
Benefits include: 
-  an easier to read and follow interface
-  type-safety
-  reduced allocations
-  performance speed-up
-  dependencies stability
-  identification of missing interfaces in upstream packages (e.g., ClimaCore)

**Risks** 
None identified at the moment.

## People and Personnel
- Lead or Co-Lead of some parts of this SDI: @valeriabarra 
- Potential Collaborators: @LenkaNovak , @juliasloan25 ?
- Reviewers: @charleskawczynski 

## Components
Examples of components (not an exhaustive list) include:

In `src/Utilities.jl`:
- [x] #224
- [x] #220
- [x] #250
- [x] Make `swap_space` not allocate. (Inline it and avoid parent). #248 

In `Regridder.jl`:
- [x] #222
- [x] #252
- [x] #254 
- [x] #256

Specifically, an example of how to avoid abstract types in structs in `src/Utilities.jl` is, instead of having

```julia
struct CoupledSimulation{FT, B, ...}
a::NamedTuple
...
end
```
we would have:

```julia
struct CoupledSimulation{FT <: Real, NT <: NamedTuple, B, ...}
a::NT
...
end
```

Also, it's useful to add helpers for extracting type information for types that are commonly needed in functions (e.g., the float type), for example:
```julia
float_type(::CoupledSimulation{FT}) where {FT} = FT
FT = float_type(coupler_sim)
```

## Inputs
These inputs were kindly provided by @charleskawczynski and can be refined/modified further as we identify more points of improvement.

## Results and deliverables
Better, type-safe, and more optimized interfaces in ClimaCoupler. 
**Update**: So far the improvements have led to about 60% less allocations. See these CI [perf jobs log](https://buildkite.com/clima/climacoupler-ci/builds/898#01862cf7-a2bc-4773-8314-6ca17eaacf4c/140).

## Task Breakdown And Tentative Due Date
(A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results)
[...]

PR #255, opened 2023-02-06
Throw warnings when zeroing NaNs

### LINKED ISSUE #254: `nan_to_zero` should throw a `@warn` to warn the user that `NaN`s were detected

valeriabarra:
## Purpose 
The purpose of this PR is to improve the `nans_to_zero` function so that it throws a warning to the user. 

Closes #254 

## Content
- [x] Added docstring
- [x] Added a warning for the user
- [x] Removed a duplicate definition in `experiments/AMIP/moist_mpi_earth/coupler_utils/masker.jl`

Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

In the Content, I have included 
- relevant unit tests, and integration tests, 
- appropriate docstrings on all functions, structs, and modules, and included relevant documentation.

### DIFF
--- a/experiments/AMIP/moist_mpi_earth/coupler_utils/masker.jl
+++ b/experiments/AMIP/moist_mpi_earth/coupler_utils/masker.jl
@@ -30,7 +30,7 @@ end
 """
     update_masks!(cs)
 
-Updates dynamically changing masks. 
+Updates dynamically changing masks.
 """
 function update_masks!(cs)
 
@@ -52,7 +52,7 @@ end
 """
     binary_mask(var::FT; threshold = 0.5)
 
-Converts a number to 1 or 0 of the same type, based on a threashold. 
+Converts a number to 1 or 0 of the same type, based on a threashold.
 """
 
 binary_mask(var::FT; threshold = 0.5) where {FT} = (var - FT(threshold)) > FT(0) ? FT(1) : FT(0)
@@ -64,18 +64,23 @@ Sums Field objects in `fields` weighted by the respective masks.
 """
 function combine_surfaces!(combined_field::Fields.Field, masks::NamedTuple, fields::NamedTuple)
     combined_field .= eltype(combined_field)(0)
+    warn_nans = false
     for surface_name in propertynames(fields) # could use dot here?
-        field_no_nans = nans_to_zero.(getproperty(fields, surface_name))  # TODO: performance analysis / alternatives
-        combined_field .+= getproperty(masks, surface_name) .* field_no_nans
+        if any(x -> isnan(x), getproperty(fields, surface_name))
+            warn_nans = true
+        end
+        combined_field .+= getproperty(masks, surface_name) .* nans_to_zero.(getproperty(fields, surface_name))
     end
-
+    warn_nans && @warn "NaNs were detected and converted to zeros."
 end
-nans_to_zero(v) = isnan(v) ? FT(0) : v
+
+#= Converts NaNs to zeros of the same type. =#
+nans_to_zero(v) = isnan(v) ? typeof(v)(0) : v
 
 """
     update_masks!(cs)
 
-Updates dynamically changing masks. 
+Updates dynamically changing masks.
 """
 function update_masks!(cs)
 
@@ -97,7 +102,7 @@ end
 
 """
     time_slice_ncfile(sic_data, time_idx = 1)
-- slices a dataset at time index `time_idx` and saves it under `sic_data_slice`. Used for more efficient regridding of mask, SST and SIC files. 
+- slices a dataset at time index `time_idx` and saves it under `sic_data_slice`. Used for more efficient regridding of mask, SST and SIC files.
 """
 function time_slice_ncfile(sic_data, time_idx = 1)
     sic_data_slice = sic_data[1:(end - 3)] * "_one_time.nc"
--- a/src/Regridder.jl
+++ b/src/Regridder.jl
@@ -17,8 +17,17 @@ using Dates
 using JLD2
 
 export write_to_hdf5,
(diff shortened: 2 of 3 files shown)


src/Regridder.jl
```diff
@@ -17,9 +17,27 @@ using Dates
...
+
+"""
+    nans_to_zero(v)
+
+Converts all NaNs to zero of the same type and throws a warning.
+
+"""
+function nans_to_zero(v)
+    v = isnan(v) ? typeof(v)(0) : v
+    @warn "NaNs were detected and converted to zeros."
```
valeriabarra, line 38: @charleskawczynski , @LenkaNovak I applied the suggestion in SDI #205 to throw a warning but perhaps it was not a good idea, since it is now flooding the CI logs (see this [build](https://buildkite.com/clima/climacoupler-ci/builds/901)) and we can't even see the entire log as they get truncated because too long.

experiments/AMIP/moist_mpi_earth/coupler_utils/masker.jl
```diff
@@ -70,12 +70,21 @@ function combine_surfaces!(combined_field::Fields.Field, masks::NamedTuple, fiel
...
 end
-nans_to_zero(v) = isnan(v) ? FT(0) : v
+
+"""
+    nans_to_zero(v)
+Converts all NaNs to zero of the same type and throws a warning.
+"""
+function nans_to_zero(v)
+    v = isnan(v) ? typeof(v)(0) : v
+    @warn "NaNs were detected and converted to zeros."
```
charleskawczynski, line 80: Based on the way that this function is used, I would recommend not putting a warning here.

valeriabarra: Ok, thanks. So you would put it only in the `Regridder.jl` module

src/Regridder.jl
charleskawczynski: I suggest inlining/deleting this function and adding a warning if any elements are zero.

valeriabarra: What do you mean by inlining/deleting this function? I think those are two different things

experiments/AMIP/moist_mpi_earth/coupler_utils/masker.jl
valeriabarra: @charleskawczynski , I have used a global flag. This way, the warning is printed only once at the end of the for loop where this function is called. This is the best solution that I could think of. (Removing the function altogether would have been more difficult because of its vectorized usage). 

It passes CI and is ready for review. Thanks!

charleskawczynski (commented):
How about we do something like this:

```julia
#= Converts NaNs to zeros of the same type. =#
nans_to_zero(v) = isnan(v) ? typeof(v)(0) : v

function combine_surfaces!(combined_field::Fields.Field, masks::NamedTuple, fields::NamedTuple)
    combined_field .= eltype(combined_field)(0)
    warn_nans = false
    for surface_name in propertynames(fields) # could use dot here?
        if any(x->isnan, getproperty(fields, surface_name))
            warn_nans = true
        end
        @. combined_field += getproperty(masks, surface_name) * nans_to_zero.(getproperty(fields, surface_name))
    end
    warn_nans && @warn "NaNs were detected and converted to zeros."
end
```

I've converted the doc string to a comment, since this is not really a user-facing function.

valeriabarra:
> A module variable is not going to be good w.r.t. inference, so we should nix that.
> 
> How about we do something like this:
> 
> ```julia
> #= Converts NaNs to zeros of the same type. =#
> nans_to_zero(v) = isnan(v) ? typeof(v)(0) : v
> 
> function combine_surfaces!(combined_field::Fields.Field, masks::NamedTuple, fields::NamedTuple)
>     combined_field .= eltype(combined_field)(0)
>     warn_nans = false
>     for surface_name in propertynames(fields) # could use dot here?
>         if any(x->isnan, getproperty(fields, surface_name))
>             warn_nans = true
>         end
>         @. combined_field += getproperty(masks, surface_name) * nans_to_zero.(getproperty(fields, surface_name))
>     end
>     warn_nans && @warn "NaNs were detected and converted to zeros."
> end
> ```
> 
> I've converted the doc string to a comment, since this is not really a user-facing function.

Thank you. Just to let you know, it doesn't quite work the way it is because of `TypeError: non-boolean (typeof(isnan)) used in boolean context`.

charleskawczynski:
Right, need `any(x->isnan(x), getproperty(fields, surface_name))`

charleskawczynski:
Instead of
```julia
@. combined_field += getproperty(masks, surface_name) * nans_to_zero.(getproperty(fields, surface_name))
```
you need
```julia
combined_field .+= getproperty(masks, surface_name) .* nans_to_zero.(getproperty(fields, surface_name))
```

charleskawczynski (approved):
Nice, it looks like this also reduces allocations from 2725088 to 2685440.

If you want, we could also address the TODO in `perf/flame.jl`:

`[ Info: TODO: lower allocs_limit[perf_default_modular] to: 2725088`

I'm fine with that going in a separate PR, though.

valeriabarra:
> 2685440

Thanks Charlie! I have updated the allocation limits for all of the 3: `perf_default_modular`, `perf_coarse_single_modular`, `perf_target_amip_n32_shortrun`

charleskawczynski:
Feel free to merge / squash & merge

PR #264, opened 2023-02-25
Add self count to FlameGraphDiff

LenkaNovak:
## Purpose 

This allows counting and displaying the allocation changes of individual functions (independent of their child nodes). The need for this was identified whilst attempting to quantify local performance in #205. In the future, this will allow us to isolate allocation changes within our software from changes in the upstream models.

This is tagged as part of #229 

## To-do

- [x] separate allocation calculation into child and current function contributions. 
- [x] implement in FlameGraphDiff plots
- [x] add a ci test 
- [x] add in docs

## QA
- [x] check buildkite plots are created
- [x] check ci tests pass

## Note
- this is still in a temporary format/location, until we test/determine how this would be useful for the rest of the CliMA team. 

----

### DIFF
--- a/perf/ProfileCanvasDiff.jl
+++ b/perf/ProfileCanvasDiff.jl
@@ -21,7 +21,7 @@ mutable struct ProfileFrame
     flags::UInt8 # any or all of ProfileFrameFlag
     taskId::Union{Missing, UInt}
     children::Vector{ProfileFrame}
-    count_change::Float64 # fractional change in count: (new - old) / old 
+    count_change::Float64 # change in count or self_count: (new - old)
 end
 
 struct ProfileDisplay <: Base.Multimedia.AbstractDisplay end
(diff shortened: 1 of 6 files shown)


valeriabarra (approved):

perf/flame_test.jl
```diff
@@ -0,0 +1,116 @@
...
+
+    test_func_name = "get_y.flame_test.jl.28"
+
+    # test flame diff
+    tracked_list["$test_func_name"] = 100
+    tracked_list["self_count_$test_func_name"] = 100
+    profile_data, new_tracked_list =
+        ProfileCanvasDiff.view(Profile.fetch(), tracked_list = tracked_list, self_count = false)
+    node = find_child(profile_data.data["all"], test_func_name, self_count = false)
+    @test node.count_change == node.count - 100
```
juliasloan25, line 99: Why do you subtract 100 from `node.count` here? (And 50 from `node.count - child_sum` below?)

juliasloan25 (approved):
Looks good to me!

perf/ProfileCanvasDiff.jl
```diff
@@ -206,17 +307,20 @@ function status(node::Profile.StackFrameTree, C::Bool)
     return st
 end
 
-function add_child(graph::ProfileFrame, node, C::Bool; tracked_list = Dict{String, Int}(;))
+function add_child(graph::ProfileFrame, node, C::Bool; tracked_list = Dict{String, Int}(;), self_count = false)
```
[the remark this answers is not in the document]
LenkaNovak: Well spotted! 👀  This was left from a previous implementation.

perf/flame_test.jl
LenkaNovak: In the first case the reference value (`tracked_list["$test_func_name"]`) is set to 100 counts and the `view()` function returns (amongst other things) the change between the reference file and the count. In the case where `self_count = true` the `view()` returns the difference between the reference independent count (here set to 50) and the current reference count (i.e., node.count - child_sum). I'll clarify this in a comment here.

PR #268, opened 2023-03-01
Add `beta` factor for evaporation

### LINKED ISSUE #199: Add `beta` calculation to coupler
SurfaceFluxes and ClimaLSM both now include a beta factor to more realistically represent evaporation (see https://github.com/CliMA/SurfaceFluxes.jl/pull/75 and https://github.com/CliMA/ClimaLSM.jl/pull/124).

We should update the coupler to calculate beta so that it can be accessed by both the atmos and land. We can do this similarly to how the coupler calculates albedo, using ClimaLSM's `β` instead of `surface_albedo`.
(see https://github.com/CliMA/ClimaCoupler.jl/blob/a80650a7bb16313476e1f9d7fdb3712859eb8066/experiments/AMIP/moist_mpi_earth/push_pull.jl#L141
https://github.com/CliMA/ClimaCoupler.jl/blob/a80650a7bb16313476e1f9d7fdb3712859eb8066/experiments/AMIP/moist_mpi_earth/bucket/bucket_utils.jl#L53
https://github.com/CliMA/ClimaLSM.jl/blob/23fd80b7006ec39bc1b9a9875ae7b6d1b7d753a8/src/Bucket/bucket_parameterizations.jl#L65 ). Note that over ocean and ice, `beta` should be set to 1 - to do this, initialize atmos beta factor to 1 everywhere, then use land mask and land_beta to update over land.

To do this, we will need ClimaAtmos to add a `beta` term to their cache (see https://github.com/CliMA/ClimaAtmos.jl/issues/1214 - DONE), and to add compatibility with SurfaceFluxes v0.5.0 (see #201 - DONE).

This also requires compatibility with the most recent version of ClimaLSM, and therefore with ClimaCore v0.10.25 (see https://github.com/CliMA/ClimaCoupler.jl/issues/242 - DONE). Some of the ClimaLSM functions used by the coupler have changed interfaces and require passing in the land model (e.g. [ClimaLSM.surface_albedo](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/Bucket/bucket_parameterizations.jl#L12), used by [ClimaCoupler.land_albedo](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/moist_mpi_earth/bucket/bucket_utils.jl#L53)), so we will need to store the model in the `BucketSimulation` to have access to it.

With this change, we can also potentially remove [this correction](https://github.com/CliMA/ClimaCoupler.jl/blob/a80650a7bb16313476e1f9d7fdb3712859eb8066/experiments/AMIP/moist_mpi_earth/push_pull.jl#L133) for surface specific humidity.

See https://github.com/CliMA/ClimaLSM.jl/pull/124 for more information

Expected delivery date: 14 March 2023

juliasloan25:
## Purpose 
SurfaceFluxes and ClimaLSM both now include a beta factor to more realistically represent evaporation (see https://github.com/CliMA/SurfaceFluxes.jl/pull/75 and https://github.com/CliMA/ClimaLSM.jl/pull/124).

See #199 for more information.

Closes #199 

## To-do
- [x] Add `land_beta` function, which uses ClimaLSM's `beta_factor` function to get the beta value over land (analogous to `land_albedo`.
- [x] Add `beta` as a coupler exchange field.
- [x] Make sure that `beta` is always 1 over snow and ice - currently done in `push_pull.jl`.

----

### DIFF
--- a/experiments/AMIP/modular/components/land/bucket_init.jl
+++ b/experiments/AMIP/modular/components/land/bucket_init.jl
@@ -16,14 +16,16 @@ import ClimaLSM.Bucket:
     surface_albedo,
     snow_precipitation
 
-using ClimaLSM: make_ode_function, initialize, obtain_surface_space, make_set_initial_aux_state
+using ClimaLSM:
+    make_ode_function, initialize, obtain_surface_space, make_set_initial_aux_state, surface_evaporative_scaling
 
 """
-    BucketSimulation{P, Y, D, I}
+    BucketSimulation{M, P, Y, D, I}
 
 The bucket model simulation object.
 """
-struct BucketSimulation{P, Y, D, I}
+struct BucketSimulation{M, P, Y, D, I}
+    model::M
     params::P
     Y_init::Y
     domain::D
@@ -49,77 +51,99 @@ multiple dispatch on `surface_fluxes`.
 struct CoupledAtmosphere{FT} <: AbstractAtmosphericDrivers{FT} end
 
 """
-    surface_fluxes(Y, p, 
-                    t::FT,
-                    parameters::P,
-                    atmos::PA,
-                    ) where {FT <: AbstractFloat, P <: BucketModelParameters{FT},  PA <: CoupledAtmosphere{FT}}
+    surface_fluxes(atmos::CoupledAtmosphere{FT},
+                    model::BucketModel{FT},
+                    Y,
+                    p,
+                    ) where {FT <: AbstractFloat}
 
 Computes the turbulent surface fluxes terms at the ground for a coupled simulation.
+Note that `Ch` is not used with the current implementation of the bucket model,
+but will be used once the canopy is incorporated.
+
+The turbulent energy flux is currently not split up between latent and sensible
+heat fluxes. This will be fixed once `lhf` and `shf` are added to the bucket's
+cache.
 """
 function ClimaLSM.Bucket.surface_fluxes(
-    Y::ClimaCore.Fields.FieldVector,
-    p::ClimaCore.Fields.FieldVector,
-    t::FT,
-    parameters::BucketModelParameters{FT, P},
     atmos::CoupledAtmosphere{FT},
-) where {FT <: AbstractFloat, P}
-    # coupler has done its thing behind the scenes already
-    return (turbulent_energy_flux = p.bucket.turbulent_energy_flux, evaporation = p.bucket.evaporation)
+    model::BucketModel{FT},
+    Y,
+    p,
+    _...,
+) where {FT <: AbstractFloat}
+    space = model.domain.surface.space
+    return (
+        lhf = ClimaCore.Fields.zeros(space),
+        shf = p.bucket.turbulent_energy_flux,
+        vapor_flux = p.bucket.evaporation,
+        Ch = ClimaCore.Fields.similar(p.bucket.evaporation),
+    )
 end
 
 
 
 """
-    net_radiation(Y, p, 
-                    t::FT,
-                    parameters::P,
-                    radiation::PR, p,
-                    ) where {FT <: AbstractFloat, P <: BucketModelParameters{FT}, PR <: CoupledRadiativeFluxes{FT}}
+    net_radiation(radiation::CoupledRadiativeFluxes{FT},
+                    model::BucketModel{FT},
+                    Y,
+                    p,
+                    _...,
+                    ) where {FT <: AbstractFloat}
 
 Computes the net radiative flux at the ground for a coupled simulation.
 """
 function ClimaLSM.Bucket.net_radiation(
+    radiation::CoupledRadiativeFluxes{FT},
+    model::BucketModel{FT},
     Y::ClimaCore.Fields.FieldVector,
     p::ClimaCore.Fields.FieldVector,
-    t::FT,
-    parameters::BucketModelParameters{FT, P},
-    radiation::CoupledRadiativeFluxes{FT},
-) where {FT <: AbstractFloat, P}
+    _...,
+) where {FT <: AbstractFloat}
     # coupler has done its thing behind the scenes already
     return p.bucket.R_n
 end
 
 
 """
-    ClimaLSM.Bucket.surface_air_density(p, atmos::CoupledAtmosphere)
-
-an extension of the bucket model method which returns the surface air 
+    ClimaLSM.surface_air_density(
+                    atmos::CoupledAtmosphere{FT},
+                    model::BucketModel{FT},
+                    Y,
+                    p,
+                    _...,
+                )
+an extension of the bucket model method which returns the surface air
 density in the case of a coupled simulation.
 """
-function ClimaLSM.Bucket.surface_air_density(p, atmos::CoupledAtmosphere)
-    # coupler has filled this in
+function surface_air_density(
+    atmos::CoupledAtmosphere{FT},
+    model::BucketModel{FT},
+    Y,
+    p,
+    _...,
+) where {FT <: AbstractFloat}
     return p.bucket.ρ_sfc
 end
 
 """
-    ClimaLSM.Bucket.liquid_precipitation(p, atmos::CoupledAtmosphere, t)
+    ClimaLSM.Bucket.liquid_precipitation(atmos::CoupledAtmosphere, p, t)
 
 an extension of the bucket model method which returns the precipitation
 (m/s) in the case of a coupled simulation.
 """
-function liquid_precipitation(p, atmos::CoupledAtmosphere, t)
+function liquid_precipitation(atmos::CoupledAtmosphere, p, t)
     # coupler has filled this in
     return p.bucket.P_liq
 end
 
 """
-    ClimaLSM.Bucket.snow_precipitation(p, atmos::CoupledAtmosphere, t)
+    ClimaLSM.Bucket.snow_precipitation(atmos::CoupledAtmosphere, p, t)
 
 an extension of the bucket model method which returns the precipitation
 (m/s) in the case of a coupled simulation.
 """
-function snow_precipitation(p, atmos::CoupledAtmosphere, t)
+function snow_precipitation(atmos::CoupledAtmosphere, p, t)
     # coupler has filled this in
     return p.bucket.P_snow
 end
@@ -168,7 +192,7 @@ function bucket_init(
     z_0b = FT(1e-3)
     κ_soil = FT(0.7)
     ρc_soil = FT(2e6)
-    t_crit = dt # This is the timescale on which snow exponentially damps to zero, in the case where all 
+    t_crit = dt # This is the timescale on which snow exponentially damps to zero, in the case where all
     # the snow would melt in time t_crit. It prevents us from having to specially time step in cases where
     # all the snow melts in a single timestep.
     params = BucketModelParameters(κ_soil, ρc_soil, albedo, σS_c, W_f, z_0m, z_0b, t_crit, earth_param_set)
@@ -206,12 +230,11 @@ function bucket_init(
     Y.bucket.W .= 0.5#0.14
     Y.bucket.Ws .= 0.0
     Y.bucket.σS .= 0.0
-    ρ_sfc = zeros(axes(Y.bucket.W)) .+ FT(1.1)
     P_liq = zeros(axes(Y.bucket.W)) .+ FT(0.0)
     P_snow = zeros(axes(Y.bucket.W)) .+ FT(0.0)
-    variable_names = (propertynames(p.bucket)..., :ρ_sfc, :P_liq, :P_snow)
+    variable_names = (propertynames(p.bucket)..., :P_liq, :P_snow)
     orig_fields = map(x -> getproperty(p.bucket, x), propertynames(p.bucket))
-    fields = (orig_fields..., ρ_sfc, P_liq, P_snow)
+    fields = (orig_fields..., P_liq, P_snow)
     p_new = ClimaCore.Fields.FieldVector(; :bucket => (; zip(variable_names, fields)...))
 
     # Set initial aux variable values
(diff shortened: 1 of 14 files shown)


experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -49,77 +51,87 @@ multiple dispatch on `surface_fluxes`.
...
-    parameters::BucketModelParameters{FT, P},
     atmos::CoupledAtmosphere{FT},
-) where {FT <: AbstractFloat, P}
-    # coupler has done its thing behind the scenes already
+    model::BucketModel{FT},
+    Y,
+    p,
+    _...,
+) where {FT <: AbstractFloat}
     return (turbulent_energy_flux = p.bucket.turbulent_energy_flux, evaporation = p.bucket.evaporation)
```
kmdeck, line 69: It ends up being helpful for us to split the energy flux into LHF and SHF. however, I know the atmosphere only computes the sum. let's discuss a bit more!

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -16,14 +16,16 @@ import ClimaLSM.Bucket:
...
 """
-    BucketSimulation{P, Y, D, I}
+    BucketSimulation{M, P, Y, D, I}
 
 The bucket model simulation object.
 """
-struct BucketSimulation{P, Y, D, I}
+struct BucketSimulation{M, P, Y, D, I}
+    model::M
     params::P
```
kmdeck, line 29: if you dont want to store params, that is fine - it should be stored in `model.parameters`

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -1,8 +1,8 @@
 """
     get_land_temp(slab_sim::BucketSimulation)
 
-Returns the surface temperature of the earth; 
-a method for the bucket model 
+Returns the surface temperature of the earth;
+a method for the bucket model
 when used as the land model.
 """
 function get_land_temp(slab_sim::BucketSimulation)
```
kmdeck, line 8: If you want, you could also use this: ClimaLSM.surface_temperature
https://github.com/CliMA/ClimaLSM.jl/blob/main/src/Bucket/bucket_parameterizations.jl#L36

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -46,27 +46,33 @@ end
...
 """
     get_land_q(slab_sim::Bucketimulation, _...)
 
-Returns the surface specific humidity of the earth; 
-a method for the bucket 
+Returns the surface specific humidity of the earth;
+a method for the bucket
 when used as the land model.
 """
 function get_land_q(slab_sim::BucketSimulation, _...)
```
kmdeck, line 78: If you would like, you can also use this
https://github.com/CliMA/ClimaLSM.jl/blob/main/src/Bucket/bucket_parameterizations.jl#L46

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -49,77 +51,87 @@ multiple dispatch on `surface_fluxes`.
...
-function ClimaLSM.Bucket.surface_air_density(p, atmos::CoupledAtmosphere)
-    # coupler has filled this in
+function surface_air_density(
+    atmos::CoupledAtmosphere{FT},
+    model::BucketModel{FT},
+    Y,
+    p,
+    _...,
+) where {FT <: AbstractFloat}
     return p.bucket.ρ_sfc
```
kmdeck, line 114: The coupler has still set this value, correct? On our side, we compute it ourselves (PrescribedAtmos). But for anything dispatching off of CoupledAtmos, the coupler has to set that field

experiments/AMIP/modular/components/land/bucket_utils.jl
kmdeck: We also have an emissivity function which just returns 1 for the bucket - if that would be useful already to start using

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -46,27 +46,33 @@ end
...
 end
 
+"""
+   land_beta(slab_sim::BucketSimulation)
+
+Returns the beta factor of the earth;
+a method for the bucket model
+when used as the land model.
+"""
+function land_beta(slab_sim::BucketSimulation)
```
LenkaNovak, line 65: It could be good to clarify that this is referring to the `beta_factor` variable in the [land model](https://github.com/CliMA/ClimaLSM.jl/blob/507ec51e4032593b1cbe3ff10e916d8ee9f06e39/src/Bucket/bucket_parameterizations.jl#L87). Not the `β` variable

LenkaNovak (approved):
Looks great, Julia!

experiments/AMIP/modular/components/land/bucket_utils.jl
LenkaNovak: This would be great! I think we would need to make it a cached parameter in the Atmos interface, so I'd probably open a new issue for it. :)

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: Do you mean ClimaLSM splits the energy flux into LHF and SHF?

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: I'm planning to make this change in a separate PR so this one doesn't become too large - see issue [here](https://github.com/CliMA/ClimaCoupler.jl/issues/273)

experiments/AMIP/modular/components/land/bucket_utils.jl
juliasloan25: Would we add a new `land_emissivity` function, or use the bucket emissivity function in an existing coupler function?

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: We have a function `set_ρ_sfc!` [here](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/modular/components/flux_calculator.jl#L12), but I don't see it actually get called. @LenkaNovak can you confirm if we set `ρ_sfc`?

experiments/AMIP/modular/components/land/bucket_utils.jl
juliasloan25: Added a comment clarifying this!

experiments/AMIP/modular/components/land/bucket_init.jl
kmdeck: perfect!

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -17,26 +17,14 @@ Returns the surface temperature of the earth, computed
...
-        bottom = ClimaCore.Operators.Extrapolate(),
-    )
-    surface_space = land_sim.domain.domain.surface.space
-    return ClimaCore.Fields.Field(
-        ClimaCore.Fields.field_values(
-            ClimaCore.Fields.level(interp_c2f.(u.bucket.T), ClimaCore.Utilities.PlusHalf(N - 1)),
-        ),
-        surface_space,
-    )
+    return ClimaLSM.surface_temperature(land_sim.model, u, land_sim.integrator.p)
```
kmdeck, line 20: I think we may need to also pass an empty fourth argument to access this method? Im not sure. 

https://github.com/CliMA/ClimaLSM.jl/blob/507ec51e4032593b1cbe3ff10e916d8ee9f06e39/src/Bucket/bucket_parameterizations.jl#L36

experiments/AMIP/modular/components/land/bucket_init.jl
kmdeck: Hm, yes - so when we compute it ourselves, we call surface fluxes, which returns the LHF and SHF separately. then we add them together outside of `surface_fluxes()`. 
https://github.com/CliMA/ClimaLSM.jl/blob/507ec51e4032593b1cbe3ff10e916d8ee9f06e39/src/Bucket/Bucket.jl#L473-L475

here is the prescribedAtmos version:
https://github.com/CliMA/ClimaLSM.jl/blob/507ec51e4032593b1cbe3ff10e916d8ee9f06e39/src/SharedUtilities/drivers.jl#L225

This is hacky, but we could do something like
`return (lhf = FT(0), shf = p.bucket.turbulent_energy_flux, vapor_flux = p.bucket.evaporation, Ch = FT(0)` here?

[We ended up returning LHF and SHF separately, as well as passing back Ch, because we are trying to adjust for canopy resistance/stomatal conductance in a new PR, but in an offline fashion, since SF.jl doesnt account for it yet.

juliasloan25: @LenkaNovak what do you think?

experiments/AMIP/modular/components/land/bucket_utils.jl
juliasloan25: Just opened a grab bag issue: https://github.com/CliMA/ClimaCoupler.jl/issues/274

experiments/AMIP/modular/components/land/bucket_utils.jl
juliasloan25: Good catch!

experiments/AMIP/modular/components/land/bucket_init.jl
LenkaNovak: Hmm, @kmdeck am I interpreting this correctly in that this function output is not being used by the simple bucket? I.e., the fluxes calculated by the coupler are only feeding into atmos right now, while this is just a placeholder for the canopy? If so, @juliasloan25 I would comment on this in the code, and do what Kat suggests (though perhaps setting all the output fields with unitialized arrays - using `similar()`).

juliasloan25: What is `Ch` here?

LenkaNovak: That's the transfer coefficient (comes to out SF.jl)

experiments/AMIP/modular/components/land/bucket_init.jl
LenkaNovak: Yes, at the moment we're using Atmos's `set_surface_thermo_state`, which calculates `ρ_sfc` internally, so in fact this variable doesn't need to exist in the coupler anymore. I was going to remove it as part of the generalized interface SDI but I won't object to deleting it here.

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: @kmdeck would it be possible to add `lhf` and `shf` to the bucket's cache so that we can access each component individually?

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: Discussed offline - we're going to remove `ρ_sfc` in the generalized interface SDI to avoid scope creep here.

kmdeck: sounds good, thank you for checking!

experiments/AMIP/modular/components/land/bucket_init.jl
kmdeck: We could do that, but as I recall the atmosphere only returns the sum of them, so Im not sure it would materially change things? let's discuss when we chat at 10

kmdeck: > Hmm, @kmdeck am I interpreting this correctly in that this function output is not being used by the simple bucket? I.e., the fluxes calculated by the coupler are only feeding into atmos right now, while this is just a placeholder for the canopy? If so, @juliasloan25 I would comment on this in the code, and do what Kat suggests (though perhaps setting all the output fields with unitialized arrays - using `similar()`).

Hey Lenka,

in the coupled sim, the fluxes computed by the coupler or the atmosphere are needed by the bucket model. But the bucket model needs only the net energy flux, not split into LHF and SHF.

In the long run, Im not sure if the canopy model really needs it split up either (though Im sure it would be useful for diagnostics and testing, so not bad to have). But, since SurfaceFluxes.jl does not take into account surface conductance (stomatal conductance or soil conductance), we returned them seperately so that as best as we could, we could account for them in standalone runs. 

That is shown here
https://github.com/CliMA/ClimaLSM.jl/blob/0426df82cca3aaba0c8e8d9034b7183a685f6886/src/Vegetation/Canopy.jl#L250-L268

But, E also appears in the SHF, so this still isnt really correct. really just a stand-in on our end.

happy to chat more!

LenkaNovak: Hmm, the atmos interface for this has changed quite a bit recently. If I'm interpreting this correctly, they now store `surface_conditions` and then set them [here](https://github.com/CliMA/ClimaAtmos.jl/blob/248e0f95bc27573494704da0bf670289c15fce08/src/tendencies/vertical_diffusion_boundary_layer.jl#L307), with `lhf` and `shf` as separate components. We use this function (called at a different time) in the coupler. If it's easier for the Bucket to keep the sum then that's fine too. We can write an adapter for it. Let us know, Kat, what you decide to do and we'll adapt.

juliasloan25: Kat and I discussed this offline. We decided it will be helpful in the future for the Bucket to have `lhf` and `shf` as separate components. Kat will work on making this change in ClimaLSM and let us know when it's done so we can store `lhf` and `shf` instead of setting one to 0 and the other storing the sum. We agreed that this should be a separate PR :)

juliasloan25:
bors try

Issue #232, opened 2023-01-18
Restructure experiments directory


LenkaNovak:
This is to separate new modular code from the old monolith. We want to retain the latter for now to assist with performance tests (Tier 2). 

This is waiting on #205.

PR #270, opened 2023-03-06
Refactor AMIP experiments

LenkaNovak:
## Purpose 

Refactors the experiments directory to contain only the currently maintained code, and ordered similarly to the CIME setup

- addresses #232 

## To-do

- [x] new components dir with model setup files
- [x] test the same results are obtained as before this PR - Buildkite output checked for both slabplanet and AMIP
- [x] clean up

----


juliasloan25 (approved):
LGTM! This setup is much more organized :)

Issue #271, opened 2023-03-07
Constrain surface mask area calculation for Float32


LenkaNovak:
Problem: 

<img width="500" alt="Screen Shot 2023-03-06 at 2 02 22 PM" src="https://user-images.githubusercontent.com/38083334/223295145-e29973b3-17b4-4fd9-8fba-dd6ab9338d53.png">
<img width="593" alt="Screen Shot 2023-03-06 at 2 57 16 PM" src="https://user-images.githubusercontent.com/38083334/223295152-ad412c17-905a-4fd4-9764-6481bd990da5.png">

and `eps(Float32)` is `1.1920929f-7`

Heisenbug, not always reproducible locally.

LenkaNovak:
for now implemented a more explicit enforcement of max(x,0) in #272

PR #272, opened 2023-03-07
Constrain surface area calculation for Float32 

LenkaNovak:
Closes issue #271 

----

### DIFF
--- a/src/Regridder.jl
+++ b/src/Regridder.jl
@@ -409,13 +409,15 @@ function update_masks!(cs::CoupledSimulation)
     # static mask
     land_s = cs.surface_masks.land
 
-    cs.surface_masks.ice .= min.(ice_d .+ land_s, FT(1)) .- land_s
-    cs.surface_masks.ocean .= (FT(1) .- cs.surface_masks.ice .- land_s)
+    # max needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
+    cs.surface_masks.ice .= max.(min.(ice_d, FT(1) .- land_s), FT(0))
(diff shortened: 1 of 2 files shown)


src/Regridder.jl
```diff
@@ -409,14 +409,15 @@ function update_masks!(cs::CoupledSimulation)
...
     land_s = cs.surface_masks.land
 
-    cs.surface_masks.ice .= min.(ice_d .+ land_s, FT(1)) .- land_s
-    cs.surface_masks.ocean .= (FT(1) .- cs.surface_masks.ice .- land_s)
+    cs.surface_masks.ice .= min.(ice_d, FT(1) .- land_s)
+    cs.surface_masks.ocean .= max.(FT(1) .- (cs.surface_masks.ice .+ land_s), FT(0)) # max needed to avoid Float32 errors
 
     @assert minimum(cs.surface_masks.ice) >= FT(0)
     @assert minimum(cs.surface_masks.land) >= FT(0)
     @assert minimum(cs.surface_masks.ocean) >= FT(0)
```
akshaysridhar, line 417: LGTM - maybe include a snippet of the code to show what the problem was initially (generated `-2.98e-8` values for the `cs.surface_masks.ocean` variable while `eps(Float32)` is `1.1920929f-7`, unless this is already tracked elsewhere in an issue). The assertion could be relaxed with `eps(FloatType)` criterion but it looks like we want strict >=0 FT(0) checks here.

akshaysridhar: Linked Issue #271

LenkaNovak: Sounds good, thank you for the review! I already referred to #271 in the buildkite pipeline file, but it's a great idea to comment in the `Regridder.jl` too.

PR #278, opened 2023-04-04
remove BucketSim params

### LINKED ISSUE #273: Remove params from BucketSimulation
As of #268 , the `BucketSimulation` struct contains the land model. The model itself contains the model parameters, so these no longer need to be stored as an individual field in `BucketSimulation`. We should remove `params` from `BucketSimulation` and update everywhere in the code this field is accessed, e.g. replacing `land_sim.params` with `land_sim.model.params`. This should be a very short PR.

Addressed by #278

juliasloan25:
## Purpose 
Remove `params` field from `BucketSimulation`, use `model.parameters` instead.

Closes #273

----


test/conservation_checker_tests.jl
```diff
@@ -112,6 +113,88 @@ end
     rm(tmp_dir, recursive = true)
 end
 
+@testset "test check_conservation for field update" begin
```
LenkaNovak, line 116: How about "land field update" to be more explicit?

LenkaNovak (approved):
Looks good, Julia. Only a tiny suggestion to change the test name. Thank you!

PR #279, opened 2023-04-10
Address buildkite memory issues

### LINKED ISSUE #192: MPI AMIP FINE memory issues
Recently, the `MPI AMIP FINE` test in CI has begun failing nondeterministically, initially [here](https://buildkite.com/clima/climacoupler-ci/builds/586#0184fa24-d737-4ab9-8a92-4e053d5d4f70). The error suggests that this is due to a lack of memory.

For now, this issue is being addressed by increasing the memory allocation for this test (see PR #195), but we should find and fix the root cause of this problem.

One task that may help enforce reproducibility is adding `compat` entries in the `Project.toml` files in this repo.

[ClimaCore #1113](https://github.com/CliMA/ClimaCore.jl/pull/1113) has resolved the memory usage issues. #279 removes the extra memory allocations we had added to the buildkite driver as a workaround.

Fixed by #279.

juliasloan25:
## Purpose 
Previously, we saw memory issues when running our `MPI AMIP FINE` test on buildkite (see #192). These issues have been resolved by memory usage changes addressed in ClimaCore PR [#1113](https://github.com/CliMA/ClimaCore.jl/pull/1113). Here, we remove the memory specifications we had added to work around this problem and replace them with the same memory limits enforced by [ClimaAtmos](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/pipeline.yml). In general, we should ensure that our buildkite driver is similar to that of ClimaAtmos.

Closes #192

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -131,7 +131,6 @@ steps:
           BUILD_HISTORY_HANDLE: ""
         agents:
           slurm_ntasks: 1
-          slurm_mem_per_cpu: 16G
 
       - label: "Moist earth with slab surface (unthreaded) - notmono + modular: bulk gray no_sponge idealinsol freq_dt_cpl"
         key: "modular_slabplanet_unthreaded"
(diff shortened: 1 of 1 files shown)


LenkaNovak (approved):
LGTM

PR #283, opened 2023-04-25
Sea Ice Energy Balance Clean-up

LenkaNovak:
## Purpose 

This PR brings back the sea ice concentration energy balance model to the [Holloway and Manabe 1971](https://journals.ametsoc.org/view/journals/mwre/99/5/1520-0493_1971_099_0335_socbag_2_3_co_2.xml?tab_body=pdf) formulation, which is the AMIP requirement. 

## To-do

- [x] reapply binary_mask correctly (accounting for curbing T to freezing point, masking)
- [x] rename ice_mask -> ice_fraction (clarifying functionality); NB: this will be done for the other models in a separate PR. 
- [x] apply_mask (only defined in sea ice init file) -> use the more general Regridder.binary_mask instead
- [x] add sea ice tests

## QA 
(from [Buildkite runs](https://buildkite.com/clima/climacoupler-ci/builds/1119#0187bbca-670a-4c1d-b885-92e719d750ab))
- AMIP: yields physically reasonable results
- Performance: not significantly affected
- (slabplanet: retains conservation - should be unaffected by this PR)

----
- [x] I have read and checked the items on the review checklist.

PR #288, opened 2023-04-27
Rename masks to fractions

LenkaNovak:
## Purpose 

- rename (surface) masks to (area) fractions for objects which are not necessarily comprised of ones and zeros, to clarify their function. 

## To-do

- [x] rename where appropriate (in `AMIP/modular`, `tests`, `docs`)
- [x] add the missing mask to the slab ocean rhs

## QA
- [x] checked that no changes in behaviour of Buildkite drivers (compared to the merging #283 PR) 

----

### DIFF
--- a/docs/src/regridder.md
+++ b/docs/src/regridder.md
@@ -15,8 +15,8 @@ ClimaCoupler.Regridder.write_to_hdf5
 ClimaCoupler.Regridder.read_from_hdf5
 ClimaCoupler.Regridder.dummmy_remap!
 ClimaCoupler.Regridder.remap_field_cgll_to_rll
-ClimaCoupler.Regridder.land_sea_mask
-ClimaCoupler.Regridder.update_masks!
+ClimaCoupler.Regridder.land_fraction
+ClimaCoupler.Regridder.update_surface_fractions!
 ClimaCoupler.Regridder.combine_surfaces!
 ```
 
--- a/experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
+++ b/experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
@@ -38,19 +38,18 @@ end
 
 # ode
 function slab_ocean_rhs!(dY, Y, cache, t)
-    p, F_aero, F_rad, ocean_mask = cache
+    p, F_aero, F_rad, ocean_fraction = cache
     FT = eltype(Y.T_sfc)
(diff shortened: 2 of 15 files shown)


docs/src/regridder.md
```diff
@@ -15,8 +15,8 @@ ClimaCoupler.Regridder.write_to_hdf5
 ClimaCoupler.Regridder.read_from_hdf5
 ClimaCoupler.Regridder.dummmy_remap!
 ClimaCoupler.Regridder.remap_field_cgll_to_rll
-ClimaCoupler.Regridder.land_sea_mask
-ClimaCoupler.Regridder.update_masks!
+ClimaCoupler.Regridder.land_fraction
```
valeriabarra, line 18: I know it might become longer and you might have favored brevity, but wouldn't it be more consistent and clear to have `land_surface_fraction` and `ocean_surface_fraction` throughout? Just an idea

LenkaNovak: I think it's quite clear since the land and ocean are just surface models in this Coupler version, but happy to change it if it is unclear to a fresh eye. @juliasloan25, shall we put this to a vote? We can also call these `area_fractions` if we want to be more explicit.

valeriabarra: Yeah, to me `land_surface_fraction` or `land_area_fraction` and similar for ocean would sound more clear. But up to you guys!

juliasloan25: I think `land_fraction` and `ocean_fraction` make sense since the coupler is using surface models, as Lenka said.

test/Project.toml
```diff
@@ -12,6 +12,7 @@ MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
juliasloan25, line 15: Why do we need to add this package?

juliasloan25 (approved):
Looks good!

test/Project.toml
LenkaNovak: Ahh well spotted, sneaked in! 👀

PR #289, opened 2023-04-27
Add buildkites for mono / binary masks to the final release list

LenkaNovak:
## Purpose 

- start a list of Buildkite drivers that we want to keep for the release.

## To-do

- move / add tests for 
  - [x] slabplanet default (stating the default options)
  - [x] slabplanet - testing non-monotonous remapping of the land-sea
  - [x] slabplanet - performance tests

- AMIP will be added later, once we decide on the final AMIP default. 

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -110,6 +110,62 @@ steps:
   - group: "Integration Tests"
     steps:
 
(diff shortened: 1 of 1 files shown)


juliasloan25 (commented):
Looks good!

PR #304, opened 2023-05-10
Update CI to use exp/AMIP/modular/

### LINKED ISSUE #299: Remove experiments/AMIP/moist_mpi_earth env from buildkite CI
As of SDI #281 , the `experiments/AMIP/modular` environment is added to our buildkite pipeline, which adds 10 minutes to the initialization time of our buildkite CI. To counteract this extra time, we should try to minimize initializing any environments which are no longer needed or won't be maintained in the future.

Since `experiments/AMIP/moist_mpi_earth` is replaced by `experiments/AMIP/modular`, we should remove `moist_mpi_earth` from our pipeline and update any tests relying on it to use the `modular` folder/environment instead.

Addressed by #304

juliasloan25:
## Purpose 
Our buildkite CI pipeline currently initialized both the `experiments/AMIP/moist_mpi_earth` and `experiments/AMIP/modular` environments, and runs tests using the drivers in both. The `modular` folder is meant to replace `moist_mpi_earth`, so we can remove the `moist_mpi_earth` initialization and replace all tests using it with the `modular` environment instead. 

This change makes the initialization stage of buildkite CI about 5 minutes faster.

Closes #299 

## Content
- [x] remove `moist_mpi_earth` environment initialization
- [x] replace `moist_mpi_earth` tests with `modular`

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -38,11 +38,6 @@ steps:
       - "julia --project=experiments/ClimaCore/sea_breeze -e 'using Pkg; Pkg.precompile()'"
       - "julia --project=experiments/ClimaCore/sea_breeze -e 'using Pkg; Pkg.status()'"
 
-      - echo "--- Instantiate amip env"
-      - "julia --project=experiments/AMIP/moist_mpi_earth/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project=experiments/AMIP/moist_mpi_earth/ -e 'using Pkg; Pkg.precompile()'"
-      - "julia --project=experiments/AMIP/moist_mpi_earth/ -e 'using Pkg; Pkg.status()'"
-
       - echo "--- Instantiate amip modular"
       - "julia --project=experiments/AMIP/modular/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
       - "julia --project=experiments/AMIP/modular/ -e 'using Pkg; Pkg.precompile()'"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -177,48 +172,42 @@ steps:
...
+      #   artifact_paths: "experiments/AMIP/modular/output/slabplanet/target_params_in_slab1_artifacts/total_energy*.png"
 
       - label: "AMIP - modular"
         key: "modular_amip"
-        command: "julia --color=yes --project=experiments/AMIP/modular/ experiments/AMIP/modular/coupler_driver_modular.jl --run_name coarse_single_modular --enable_threading true --coupled true  --surface_scheme bulk --moist equil --vert_diff true --rad gray --energy_check false --mode_name amip --anim true --t_end 32days --dt_save_to_sol 1days --dt_cpl 400 --dt 400secs --mono_surface false --h_elem 6 --dt_save_restart 10days --precip_model 0M --job_id coarse_single_modular"
+        command: "julia --color=yes --project=experiments/AMIP/modular/ experiments/AMIP/modular/coupler_driver_modular.jl --run_name coarse_single_modular --enable_threading true --coupled true  --surface_scheme monin_obukhov --moist equil --vert_diff true --rad gray --energy_check false --mode_name amip --anim true --t_end 32days --dt_save_to_sol 1days --dt_cpl 400 --dt 400secs --mono_surface false --h_elem 6 --dt_save_restart 10days --precip_model 0M --job_id coarse_single_modular"
```
valeriabarra, line 210: was this meant to change also the `surface_scheme` from `bulk` to `monin_obukhov`?

juliasloan25: The `AMIP` test before was using `monin_obukhov`, and `AMIP - modular` was using `bulk`. I changed this one to `monin_obukhov` because that's what most of the other tests were using, but I'm not really sure which we want here.

valeriabarra: Okay, it looks like we would still have at least one job that tests the `bulk` version, that is "Moist earth with slab surface - test: bulk allsky sponge realinsol infreq_dt_cpl" (which is probably good to exercise different parts of the code). That's fine then. But let's also double check with Lenka when she's back

valeriabarra (approved):

.buildkite/pipeline.yml
juliasloan25: I just discussed with Zhaoyi, who explained that `monin_obukhov` is a more realistic surface scheme that we want to use for AMIP simulations. From that, I think it makes sense for this test to use `monin_obukhov` instead of `bulk`. We can check with Lenka next week too :)

PR #296, opened 2023-04-28
Clean up surfce fluxes

LenkaNovak:
## Purpose 

- cleans up surface fluxes

## To-do

- [x] remove redundant code, move the flux calculation out of push_pull!
  - [X] ensure the same behaviour as for the last merge
- [X] add surface flux calclulation
  - [x] ensure the same behaviour as for the last merge
- [x] add tests to all new functions
- [x] transform into FluxCalculator module
  - [x] ensure the same behaviour as for the last merge
- [x] update for the latest Atmos release
  - [x] ensure the same behaviour as for the last merge

----
- [ ] I have read and checked the items on the review checklist.

PR #316, opened 2023-06-03
Update `combine_surfaces!`

### LINKED ISSUE #317: Update `combine_surfaces!`
Before running the functionality that calculates fluxes over multiple arbitrary surfaces, as specified in #311 , we need to update `combine_surfaces!` and `update_surface_fractions`. This includes a new `Interfacer` module that prescribes the component model interfaces. This is a prequel to the `ClimaEarth.jl` design, and a step towards having a source code that has no dependencies on the data structures of the underlying component models.

## Components
- [x] update code with interface changes
- [x] add unit tests for `Interfacer`
- [x] update slabpanet
- [x] update AMIP

## QA
- [x] the same behavior compared to current main recovered 
  - [x] conservation
  - [x] AMIP paperplots
- [x] comparable performance to the current main 
- [x] CI unit tests pass

## Note on the design
- we're using a function `get_field(sim::ComponentModelSimulation, ::Val(:{field_of_interest}))` for now, which allows model-agnostic field specification, as well as definition of wrapper methods, such as `get_field(sim::ComponentModelSimulation, ::Val, colidx)`. While we could define variable-specific functions, e.g. `get_termperature`, and use anonymous functions for the wrapper functions, the wrapper functions would be less transparent and may cause conflicts with methods of the same signature. Alternatively, we could define each variable as a struct object upon which we dispatch, but this adds a bit more code and less flexibility when developing. The current solution seems like a good temporary compromise.

LenkaNovak:
## Purpose 

We need to generalize `combine_surfaces!` and `update_area_fractions` before introducing the new surface fluxes outlined in #296 . 

Closes #317 (also see for more details)

----

### DIFF
--- a/docs/src/interfacer.md
+++ b/docs/src/interfacer.md
@@ -1,0 +1,55 @@
+# Interfacer
+
+This module contains functions for defining the interface for coupling component models, as well as stub objects that contain prescribed fields.
+
+## Coupled Simulation
+- `CoupledSimulation` (`cs`, currently in Utilities - TODO) stores info for ESM run. We require that each `cs` contains four (`atmos_sim`, `land_sim`, `ocean_sim` and `ice_sim`) components. While this requirement will not be eventually needed, for the time being, if a simulation surface type is not needed for a given run, it should be initialized with `SurfaceStub` with a zero `area_fracion`. The `atmos_sim` should always be specified.
+
+## Component model simulations
+- all Simulations that are not the `CoupledSimulation` fall under `ComponentModelSimulation`
+- the current version requires that there is:
+    - one `AtmosModelSimulation`
+    - one or more `SurfaceModelSimulation`s, which require the following adapter functions:
+        ```
+        get_field(sim::SurfaceModelSimulation, ::Val{:area_fraction}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:surface_temperature}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:albedo}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:roughness_momentum}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:roughness_buoyancy}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:beta}) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:area_fraction}, field::Fields.Field) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:surface_temperature}, field::Fields.Field) = ...
+        ```
+        - these adapter functions, to be defined in the component models' init files (preferably in their own repositories), allow the coupler to operate without having to assume particular data structures of the underlying component models. This allows easy swapping of model components, as well as a stable source code with coupler-specific unit tests.
+
+## Prescribed conditions
+- `SurfaceStub` is a `SurfaceModelSimulation`, but it only contains required data in `<surface_stub>.cache`, e.g., for the calculation of surface fluxes through a prescribed surface state.  The above adapter functions are already predefined for `SurfaceStub`, with the cache variables specified as:
+```
+get_field(sim::SurfaceStub, ::Val{:area_fraction}) = sim.cache.area_fraction
+get_field(sim::SurfaceStub, ::Val{:surface_temperature}) = sim.cache.T_sfc
+get_field(sim::SurfaceStub, ::Val{:albedo}) = sim.cache.α
+get_field(sim::SurfaceStub, ::Val{:roughness_momentum}) = sim.cache.z0m
+get_field(sim::SurfaceStub, ::Val{:roughness_buoyancy}) = sim.cache.z0b
+get_field(sim::SurfaceStub, ::Val{:beta}) = sim.cache.beta
+```
+with the corresponding `update_field!` functions
+```
+function update_field!(sim::SurfaceStub, ::Val{:area_fraction}, field::Fields.Field)
+    sim.cache.area_fraction .= field
+end
+function update_field!(sim::SurfaceStub, ::Val{:surface_temperature}, field::Fields.Field)
+    sim.cache.T_sfc .= field
+end
+```
+
+## Interfacer API
+
+```@docs
+    ClimaCoupler.Interfacer.ComponentModelSimulation
+    ClimaCoupler.Interfacer.AtmosModelSimulation
+    ClimaCoupler.Interfacer.SurfaceModelSimulation
+    ClimaCoupler.Interfacer.SurfaceStub
+    ClimaCoupler.Interfacer.name
+    ClimaCoupler.Interfacer.get_field
+    ClimaCoupler.Interfacer.update_field!
+```
--- a/experiments/AMIP/modular/components/land/bucket_init.jl
+++ b/experiments/AMIP/modular/components/land/bucket_init.jl
@@ -24,11 +24,12 @@ using ClimaLSM:
 
 The bucket model simulation object.
 """
-struct BucketSimulation{M, Y, D, I}
+struct BucketSimulation{M, Y, D, I, A} <: SurfaceModelSimulation
     model::M
     Y_init::Y
     domain::D
     integrator::I
+    area_fraction::A
 end
 
 include("./bucket_utils.jl")
@@ -162,6 +163,7 @@ function bucket_init(
     space,
     dt::FT,
     saveat::FT,
+    area_fraction,
     stepper = Euler(),
 ) where {FT}
     if config != "sphere"
@@ -245,5 +247,5 @@ function bucket_init(
     prob = ODEProblem(ode_function!, Y, tspan, p_new)
     integrator = init(prob, stepper; dt = dt, saveat = saveat)
 
-    BucketSimulation(model, Y, (; domain = domain, soil_depth = d_soil), integrator)
+    BucketSimulation(model, Y, (; domain = domain, soil_depth = d_soil), integrator, area_fraction)
 end
--- a/experiments/AMIP/modular/components/land/bucket_utils.jl
+++ b/experiments/AMIP/modular/components/land/bucket_utils.jl
@@ -142,3 +142,13 @@ function make_lsm_domain(
         surface_domain,
     )
 end
+
+# required by Interfacer
+get_field(sim::BucketSimulation, ::Val{:surface_temperature}) = sim.integrator.p.bucket.T_sfc
+get_field(sim::BucketSimulation, ::Val{:roughness_momentum}) = sim.model.parameters.z_0m
+get_field(sim::BucketSimulation, ::Val{:roughness_buoyancy}) = sim.model.parameters.z_0b
+get_field(sim::BucketSimulation, ::Val{:beta}) =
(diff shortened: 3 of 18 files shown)


docs/src/interfacer.md
```diff
@@ -0,0 +1,56 @@
...
+        get_field(sim::SurfaceModelSimulation, ::Val{:area_fraction}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:air_temperature}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:albedo}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0m}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0b}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:beta}) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:area_fraction}, field::Fields.Field) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:air_temperature}, field::Fields.Field) = ...
+        ```
+        - these adapter function, to be defined in the component models' init files (preferably in their own repositories), allow the coupler to operate without having to assume particular data structures of the underlying component models. This allows easy swapping of model components, as well as a stable source code with coupler-specific unit tests.
```
valeriabarra, line 24:
```suggestion
        - these adapter functions, to be defined in the component models' init files (preferably in their own repositories), allow the coupler to operate without having to assume particular data structures of the underlying component models. This allows easy swapping of model components, as well as a stable source code with coupler-specific unit tests.
```

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -162,6 +163,7 @@ function bucket_init(
     space,
     dt::FT,
     saveat::FT,
+    land_fraction,
```
valeriabarra, line 166: The named field in the struct was generic `area_fraction` (as for the other `SurfaceModelSimulation` types). Is this discrepancy here desired/intentional?

src/Interfacer.jl
```diff
@@ -0,0 +1,75 @@
+"""
+    Interfacer
+
+...
+"""
```
valeriabarra, line 5: There are some missing docstring warnings in the docs
image

src/Regridder.jl
```diff
@@ -431,7 +439,29 @@ Converts a number `var` to 1, if `var` is greater or equal than a given `thresho
 binary_mask(var::FT; threshold = FT(0.5)) where {FT} = var < FT(threshold) ? FT(0) : FT(1)
 
 """
-    combine_surfaces!(combined_field::Fields.Field, fractions::NamedTuple, fields::NamedTuple)
+    combine_surfaces_from_sol!(combined_field::Fields.Field, sims, field_name::Val)
+
+Sums Field objects in `fields` weighted by the respective area fractions, and updates
+these values in `combined_field`.
+NamedTuples `fields` and `fractions` must have matching field names.
```
valeriabarra: Can we give an example here of the call with matching field names?

experiments/AMIP/modular/components/land/bucket_init.jl
LenkaNovak: Ahh, this was left from an old version. Well spotted!

experiments/AMIP/modular/components/ocean/slab_seaice_init.jl
```diff
@@ -82,3 +96,15 @@ clean_sic(SIC, _info) = swap_space!(zeros(axes(_info.land_fraction)), SIC) ./ fl
...
+
+# required by Interfacer
+get_field(sim::PrescribedIceSimulation, ::Val{:air_temperature}) = sim.integrator.u.T_sfc
+get_field(sim::PrescribedIceSimulation, ::Val{:z0m}) = sim.integrator.p.params.z0m
+get_field(sim::PrescribedIceSimulation, ::Val{:z0b}) = sim.integrator.p.params.z0b
+get_field(sim::PrescribedIceSimulation, ::Val{:beta}) = convert(eltype(sim.integrator.u), 1.0)
+get_field(sim::PrescribedIceSimulation, ::Val{:albedo}) = sim.integrator.p.params.α
+get_field(sim::PrescribedIceSimulation, ::Val{:area_fraction}) = sim.integrator.p.area_fraction
+
+function update_field!(sim::PrescribedIceSimulation, ::Val{:area_fraction}, field::Fields.Field)
```
juliasloan25, line 108: Do we need an `update_field!` method for `air_temperature`?

docs/src/interfacer.md
```diff
@@ -0,0 +1,56 @@
...
+- all Simulations that are not the `CoupledSimulation` fall under `ComponentModelSimulation`
+- the current version requires that there is:
+    - one `AtmosModelSimulation`
+    - one or more `SurfaceModelSimulation`s, which require the following adapter functions:
+        ```
+        get_field(sim::SurfaceModelSimulation, ::Val{:area_fraction}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:air_temperature}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:albedo}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0m}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0b}) = ...
```
juliasloan25, line 18: What are `z0m` and `z0b`? I think we should add this as a comment or docstring since it's not immediately clear from the names

juliasloan25 (commented):
I left a few comments, mostly suggestions for how to reduce code duplication, and clarifying questions regarding the docs. Overall I like the interface design, I think this is a good step!

docs/src/interfacer.md
```diff
@@ -0,0 +1,56 @@
+# Interfacer
+
+This module contains functions for defining the interface for coupling component models, as well as stub objects that contain prescribed fields.
+
+## Coupled Simulation
+- TODO: move `CoupledSimulation` from Utilities in a separate PR
+- `CoupledSimulation` (`cs`) stores info for ESM run. We require that each `cs` contains an `atmos_sim`, `ocean_sim` and `ice_sim` component. While this requirement will not be eventually needed, for the time being, if a simulation type is not needed, it should be initialized with `SurfaceStub` with a zero `area_fracion`.
```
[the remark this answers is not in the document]
LenkaNovak: Ahh `land_sim` is indeed missing from that list. Thanks! But aside from these 4 required simulation types, we can't add additional arbitrary simulations. That said, this design is a step towards that, so when we need that functionality, we can add it relatively easily. As for the `atmos_sim`, we're assuming it's always needed (until we have a requirement to support a prescribed atmosphere from the coupler). I've clarified the wording to this point.

docs/src/interfacer.md
```diff
@@ -0,0 +1,56 @@
...
+    - one or more `SurfaceModelSimulation`s, which require the following adapter functions:
+        ```
+        get_field(sim::SurfaceModelSimulation, ::Val{:area_fraction}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:air_temperature}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:albedo}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0m}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:z0b}) = ...
+        get_field(sim::SurfaceModelSimulation, ::Val{:beta}) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:area_fraction}, field::Fields.Field) = ...
+        update_field!(sim::SurfaceModelSimulation, ::Val{:air_temperature}, field::Fields.Field) = ...
```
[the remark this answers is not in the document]
LenkaNovak: Once we move the init functions to the component model repos, then yes, this will be a requirement. This doesn't mean they have to operate on ClimaCore spaces, but they need to be able to read the Field objects passed from the coupler (and so they'll depend on ClimaCore). We may want to generalize this in ClimaEarth to some universal objects, but for the AMIP setup everything depends on ClimaCore anyway.

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -142,3 +142,13 @@ function make_lsm_domain(
         surface_domain,
     )
 end
+
+# required by Interfacer
```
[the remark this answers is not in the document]
LenkaNovak: This will be done in the next PR, which will focus on revamping the `push_pull.jl` file (it wasn't necessary for the `combine_surfaces!` and `update_surface_fractions!` functions here)

experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
```diff
@@ -72,3 +86,15 @@ end
...
 """
 clean_sst(SST, _info) = (swap_space!(zeros(axes(_info.land_fraction)), SST) .+ float_type_bcf(_info)(273.15))
+
+# required by Interfacer
+get_field(sim::SlabOceanSimulation, ::Val{:air_temperature}) = sim.integrator.u.T_sfc
+get_field(sim::SlabOceanSimulation, ::Val{:z0m}) = sim.integrator.p.params.z0m
+get_field(sim::SlabOceanSimulation, ::Val{:z0b}) = sim.integrator.p.params.z0b
+get_field(sim::SlabOceanSimulation, ::Val{:beta}) = convert(eltype(sim.integrator.u), 1.0)
+get_field(sim::SlabOceanSimulation, ::Val{:albedo}) = sim.integrator.p.params.α
+get_field(sim::SlabOceanSimulation, ::Val{:area_fraction}) = sim.integrator.p.area_fraction
```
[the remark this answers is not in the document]
LenkaNovak: The `SlabOceanSimulation` and `PrescribedIceSimulation` models should be perceived as individual models, which have their own independent data structures. I think this way it will be less confusing when swapping them for other models / modifying them. But if we decide to support these models as a long-term stable solution, then we should probably create a new module for them. I'd be in favor of deciding this as part of ClimaEarth though...

experiments/AMIP/modular/components/ocean/slab_seaice_init.jl
```diff
@@ -1,4 +1,18 @@
-# energy equation for sea ice
+"""
+    PrescribedIceSimulation{P, Y, D, I}
+
+Ice concentration is prescribed, and solving the following energy equation:
```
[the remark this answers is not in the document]
LenkaNovak: Sure! (and it's only `T_sfc` that is prognostic)

experiments/AMIP/modular/components/ocean/slab_seaice_init.jl
LenkaNovak: Not in this PR for atmos (this will be done in the next PR), and `air_temperature` here is actually the `prognostic T_sfc` in the surface models, so it's updated internally. I will change this name to `surface_temperature`, which is more accurate.

docs/src/interfacer.md
LenkaNovak: These are roughness lengths for momentum and buoyancy. I made these more verbose. We should formalize/unify the fields (and their units) being passed by the coupler though. I'll open an issue on this.

experiments/AMIP/modular/components/ocean/slab_seaice_init.jl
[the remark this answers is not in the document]
juliasloan25: Thank you!

Issue #323, opened 2023-06-15
Performance tips


valeriabarra:
A few recommendations that we should always keep in mind to avoid any performance degradation that we may have introduced in some recent developments:

Try to avoid: 
- Closures over global variables (FT)
- Use of `Val` when not necessary-- these symbols are not needed by the compiler, so there's no need to add them just so that we can specialize on them. (this was introduced in the `update_field!` in #320 )
- Use of parent, when not necessary (though, we saw that when exchanging quantities from fields that are not from the same space, or are from the same space but not instance (see #249  for reference) this is not possible)

PR #337, opened 2023-06-27
Enable `q_sfc` calculation from each surface model

### LINKED ISSUE #338: Enable `q_sfc` to be calculated by each surface model
Necessary to be able to use Bucket's `q_sfc`, which takes into account its internal values of snow and bare soil are fractions. Here enable this, and apply the same procedure for the other surfaces. 

# Major changes
- coupler can calculate `rho_sfc`
- all surface models will be able to hold their own `rho_sfc` and calculate their own `q_sfc` (including the Stubs)
- correct specification of `Ice()` when calculating `q_vap_saturation_generic` (before only `Liquid()` was enabled)
- initial conditions are re-specified to be more consistent with atmos, enabling more stable runs

# To do
- [x] enable coupler to calculate rho_sfc based on surface and atmos states
- [x] add calculation of `q_sfc` to the remaining component models
- [x] set the new q_sfc in the `compute_atmos_turbulent_fluxes!` function
- [x] resolve grid imprinting problem when calling q_sfc from land - due to atmos init conditions being too abrupt
  - modified surface initial conditions to be more consistent with the new atmos conditions
  - ensure and document the correct order of field exchange initialization 
- [x] resolve why nondeterministic ci runs - atmos now has a random initial perturbation
- [x] add tests for new src functions

# QA
- [x] ensure tests are passing (physical, conservation, performance)

LenkaNovak:
## Purpose 

Closes #338 (see for details)

----


experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -170,6 +167,7 @@ This uses the `ClimaAtmos.jl` driver, with parameterization options specified in
 ## init atmos model component
 include("components/atmosphere/climaatmos_init.jl")
 atmos_sim = atmos_init(FT, parsed_args);
+thermo_params = get_thermo_params(atmos_sim) # TODO: this should be shared by all models
```
juliasloan25, line 170: Is there a plan for this TODO to be addressed soon(ish)?

experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -274,16 +282,20 @@ elseif mode_name == "slabplanet"
...
     ## sea ice (here set to zero area coverage)
     ice_sim = SurfaceStub((;
         T_sfc = ClimaCore.Fields.ones(boundary_space),
+        ρ_sfc = ClimaCore.Fields.zeros(boundary_space),
         z0m = FT(0),
         z0b = FT(0),
         beta = FT(1),
         α = FT(1),
         area_fraction = ClimaCore.Fields.zeros(boundary_space),
+        phase = phase = TD.Ice(),
```
juliasloan25, line 297: Extra `phase =`

experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -238,11 +236,14 @@ if mode_name == "amip"
     SST_init = interpolate_midmonth_to_daily(date0, SST_info)
     ocean_sim = SurfaceStub((;
         T_sfc = SST_init,
+        ρ_sfc = ClimaCore.Fields.zeros(boundary_space),
         z0m = FT(1e-3),
         z0b = FT(1e-3),
         beta = FT(1),
         α = FT(0.06),
         area_fraction = (FT(1) .- land_fraction),
+        phase = TD.Liquid(),
```
juliasloan25, line 245: Why don't we need to specify the water phase in the land model? I know beta provides information about how saturated the surface is for evaporation, but do we need to describe whether its saturated with water vs ice vs both?

experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -378,29 +390,41 @@ cs = CoupledSimulation{FT}(
...
 
-## share states and fluxes between models
-# 1) import states into the coupler
+#Coupler needs; (sfc.T_sfc, atmos.ts) -> atmos.rho_sfc
+# sfc need: coupler.rho_sfc
+#Coupler needs; sfc.q_sfc
+# coupler needs q_sfc, sfc.surface_conditions, atmos_surface conditions for updates
+
+#Atmos needs; coupler.T_S (, coupler.albedo)
+# dd=dddd
```
juliasloan25, line 402: This line seems like a mistake

experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -378,29 +390,41 @@ cs = CoupledSimulation{FT}(
...
 =#
 
-## share states and fluxes between models
-# 1) import states into the coupler
+#Coupler needs; (sfc.T_sfc, atmos.ts) -> atmos.rho_sfc
+# sfc need: coupler.rho_sfc
+#Coupler needs; sfc.q_sfc
+# coupler needs q_sfc, sfc.surface_conditions, atmos_surface conditions for updates
+
+#Atmos needs; coupler.T_S (, coupler.albedo)
```
juliasloan25, line 401: should these both be combined as a tuple? (not a big deal since it's just a comment)

experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -378,29 +390,41 @@ cs = CoupledSimulation{FT}(
...
 compute_combined_turbulent_fluxes!(cs.model_sims, cs.fields, turbulent_fluxes)
-# 5) reset atmos model's states and step with the correct `sfc_conditions` to caclucate radiative fluxes
-reinit!(atmos_sim)
-parsed_args["ode_algo"] == "ARS343" ? step!(atmos_sim, Δt_cpl) : nothing # calcualte F_rad from surface conditions
-# 6) export fluxes from atmos model
+
+# 4) given the new sfc_conditions, atmos calls the radiative flux callback
+reinit!(atmos_sim) # sets a nonzero radiation flux
+
+# 5) coupler re-imports updated atmos fluxes, and sends them to surface models
```
juliasloan25, line 425: Great comments in this section!

perf/flame.jl
```diff
@@ -17,7 +17,7 @@ run_name_list =
     ["default_modular_unthreaded", "coarse_single_modular", "target_amip_n32_shortrun", "target_amip_n1_shortrun"]
 run_name = run_name_list[parse(Int, ARGS[2])]
 allocs_limit = Dict()
-allocs_limit["perf_default_modular_unthreaded"] = 6113200
+allocs_limit["perf_default_modular_unthreaded"] = 8638304
```
juliasloan25, line 20: Why do we need to increase the allocation limit? Does the new way of calculating fluxes allocate more, or is this from upstream ClimaAtmos changes?

juliasloan25 (commented):
Looks good! Just some minor comments/questions

experiments/AMIP/modular/coupler_driver_modular.jl
LenkaNovak: Not immediately, but we could think about this soon - this will depend on how we want to deal with general parameter passing, so we'll need to make sure all of the dependent teams (especially, Atmos, Land, ML, and eventually Ocean) are onboard and modify their interface accordingly. I've opened an [issue](https://github.com/CliMA/ClimaCoupler.jl/issues/342) to track this.

experiments/AMIP/modular/coupler_driver_modular.jl
LenkaNovak: It's already contained internally in Land. It allows for both `Liquid()` and `Ice()`, depending on the dynamic snow fraction.

experiments/AMIP/modular/coupler_driver_modular.jl
LenkaNovak: This was actually some additional info I wrote before adding the comments below - I meant to rm this, since the comments should be sufficient. Thank you for spotting it!

perf/flame.jl
LenkaNovak: In this case, it's expected because we're increasing the cache of the surface models, so they can store their own `q_sfc` and `rho_sfc`. I'm sure there are some potential alloc gains nevertheless though - Valeria and Charlie have already started looking into this [here](https://github.com/CliMA/ClimaCoupler.jl/issues/323).

juliasloan25 (approved):
LGTM!

LenkaNovak:
> LGTM!

Awesome, thank you, @juliasloan25 🚀

PR #363, opened 2023-07-27
Add model checkpoint and restart

### LINKED ISSUE #364: Add model state checkpoint and restart
This introduces a new module that coordinates checkpointing of model states and restarting from those states. Needed for current AMIP testing. 

## Components
- [x] `Checkpointer` module with POC unit test
- [x] implementation in the AMIP driver
- [x] docs

## QA
- [x] no behavioral change
- [x] tests show that saved state == retrieved state

LenkaNovak:
## Purpose 

Closes #364

----

### DIFF
--- a/experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
+++ b/experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
@@ -1,6 +1,6 @@
 using ClimaCore
 
-import ClimaCoupler.Interfacer: OceanModelSimulation, get_field, update_field!
+import ClimaCoupler.Interfacer: OceanModelSimulation, get_field, update_field!, name
 import ClimaCoupler.FieldExchanger: step!, reinit!
 
 """
@@ -28,6 +28,7 @@ struct OceanSlabParameters{FT <: AbstractFloat}
     z0b::FT
     α::FT
 end
+name(::SlabOceanSimulation) = "SlabOceanSimulation"
 
 # init simulation
 function slab_ocean_space_init(::Type{FT}, space, p) where {FT}
@@ -128,3 +129,12 @@ end
 step!(sim::SlabOceanSimulation, t) = step!(sim.integrator, t - sim.integrator.t, true)
 
 reinit!(sim::SlabOceanSimulation) = reinit!(sim.integrator)
+
+"""
+    get_model_state_vector(sim::SlabOceanSimulation)
+
+Extension of Checkpointer.get_model_state_vector to get the model state.
+"""
+function get_model_state_vector(sim::SlabOceanSimulation)
+    return sim.integrator.u
+end
--- a/experiments/AMIP/modular/coupler_driver_modular.jl
+++ b/experiments/AMIP/modular/coupler_driver_modular.jl
@@ -87,6 +87,7 @@ if isinteractive()
     # parsed_args["dt_save_restart"] = "5days" #hide
     parsed_args["precip_model"] = "0M" #hide
     parsed_args["job_id"] = "interactive_debug_run"
+    parsed_args["monthly_checkpoint"] = true
 end
 
 ## read in some parsed command line arguments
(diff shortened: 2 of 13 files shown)


experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -388,6 +393,18 @@ cs = CoupledSimulation{FT}(
     diagnostics,
 );
 
+#=
+## Restart component model states if specified
+=#
+restart_dir = nothing
+restart_t = nothing
```
akshaysridhar, line 400: Are these temporary assignments / determined through CLI args? `restart_t` should not be valid without a `restart_dir` specified right? In `CheckPointer.jl`, `Y .= Y_new` where Y_new is evaluated at `restart_t`; without an appropriate `restart_dir` this should not modify `Y`.

experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
```diff
@@ -128,3 +129,12 @@ end
...
 
 reinit!(sim::SlabOceanSimulation) = reinit!(sim.integrator)
+
+"""
+    get_model_state_vector(sim::SlabOceanSimulation)
+
+Extension of Checkpointer.get_model_state_vector to get the model state.
+"""
+function get_model_state_vector(sim::SlabOceanSimulation)
+    return sim.integrator.u
```
akshaysridhar, line 139: I'm not exactly familiar with these containers - above, we `return sim.integrator.u.bucket` with the `BucketSimulation` - is there a similar `sim.integrator.u.slabocean` or is this indeed correct? In which case is it useful to have a fallback method dispatching on the abstract type (?)

test/checkpointer_tests.jl
```diff
@@ -0,0 +1,39 @@
...
+    sim = DummySimulation(Fields.FieldVector(T = ones(boundary_space)))
+    checkpoint_model_state(sim, comms_ctx, t, output_dir = "test_checkpoint")
+
+    # new sim run
+    sim_new = DummySimulation(Fields.FieldVector(T = zeros(boundary_space)))
+    set_checkpoint_model_state!(sim_new, t, output_dir = "test_checkpoint")
+    @test sim_new.state.T == sim.state.T
+
+    # remove checkpoint directory
+    rm("./test_checkpoint/", force = true, recursive = true)
```
akshaysridhar, line 38: Tests seem reasonable; there are entries for all 3 added functions. Minor comment -  in staying with the `checkpoint-restart` terminology would it be useful to perhaps `set_checkpoint_model_state -> restart_model_state` ?

src/Checkpointer.jl
```diff
@@ -0,0 +1,70 @@
...
+    return nothing
+
+end
+
+"""
+    set_checkpoint_model_state!(sim::Interfacer.ComponentModelSimulation, t::AbstractFloat; output_dir = "output")
+
+Sets the model state of a simulation from a HDF5 file from a given time, t (in seconds).
+"""
+function set_checkpoint_model_state!(sim::Interfacer.ComponentModelSimulation, t::AbstractFloat; output_dir = "output")
```
akshaysridhar, line 53: Minor suggestion re. naming conventions -> `output` suggests that the `set_checkpoint_model_state!` is writing to file - this is true in the `checkpoint_model_state` function above, whereas the `restart/set_checkpoint` functionality is internally dealing with an `input` directory. Otherwise maybe expanding the docstring to clarify the role of this `dir` in the keyword arg is sufficient.

experiments/AMIP/modular/coupler_driver_modular.jl
LenkaNovak: No, but I think adding these as `CLI args` makes sense actually - it will make our life easier when testing on the cluster. :-)

experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
LenkaNovak: It's correct. Each component model has its own `_init` file where it specifies model-specific paths to states and cache. I've tested it locally and it ran, but can add it to a buildkite driver in the long runs as well though.

test/checkpointer_tests.jl
LenkaNovak: Sure, good idea!

src/Checkpointer.jl
LenkaNovak: Totally! Renamed!

akshaysridhar (approved):
LGTM - I see the buildkite pipeline exercises the `monthly_checkpoint` -> In addition to the unit test do we want to exercise the `restart` capability in the buildkite runs as well ?

LenkaNovak:
Thanks! I think having it in the unit tests should be enough, since there are no other child functions that are not already tested for these sim types (there's only the `get_model_state_vector`, which is already tested using the `monthly_checkpoint`)

PR #369, opened 2023-08-01
Add MPI fix for restarts

### LINKED ISSUE #368: Add MPI fix for restarts
We need to enable MPI runs to read files. Small fix required in argument passing. 

## Components
- [x] implement fix
- [x] add slurm script for easier local checks

## QA
- [x] Include MPI test

LenkaNovak:
Closes #368

----


src/Checkpointer.jl
```diff
@@ -46,11 +46,16 @@ function checkpoint_model_state(
...
 """
-    restart_model_state!(sim::Interfacer.ComponentModelSimulation, t::Int; input_dir = "input")
+    restart_model_state!(sim::Interfacer.ComponentModelSimulation, comms_ctx::ClimaComms.AbstractCommsContext, t::Int; input_dir = "input")
 
 Sets the model state of a simulation from a HDF5 file from a given time, t (in seconds).
 """
-function restart_model_state!(sim::Interfacer.ComponentModelSimulation, t::Int; input_dir = "input")
+function restart_model_state!(
+    sim::Interfacer.ComponentModelSimulation,
+    comms_ctx::ClimaComms.AbstractCommsContext,
```
akshaysridhar, line 55: This would be a breaking change right? Perhaps its effect is limited to the coupler so this doesn't affect any release schedules. LGTM.

src/Checkpointer.jl
```diff
@@ -59,7 +64,7 @@ function restart_model_state!(sim::Interfacer.ComponentModelSimulation, t::Int;
     @info "Setting " Interfacer.name(sim) " state to checkpoint: $input_file, corresponding to day $day second $sec"
 
     # open file and read
-    hdfreader = InputOutput.HDF5Reader(input_file)
+    hdfreader = InputOutput.HDF5Reader(input_file, comms_ctx)
```
akshaysridhar, line 67: In general do we want to refer to these as `hdf5reader` ? Here and elsewhere - users might expect general compatibility with hdf formats otherwise.

test/mpi_tests/local_checks.sh
akshaysridhar, line 1: Do we want to allow these SLURM batch scripts to be checked in ? In general I thought we wanted to rely on the buildkite system right?

akshaysridhar:
Thanks @LenkaNovak ; Overall structure looks reasonable - I've left some minor comments above.

src/Checkpointer.jl
LenkaNovak: Yeah, there are no dependencies on this. Admittedly, this is what should have been in the original Checkpointer PR a few days ago.

src/Checkpointer.jl
LenkaNovak: Sure, we were following the syntax of ClimaCore and ClimaAtmos, but we can change the naming to something more legible in the local scope of this function. Would be good to have this discussion more widely at some point 😅 how about `checkpoint_writer` and `restart_reader`?

test/mpi_tests/local_checks.sh
LenkaNovak: Yeah, this is more for offline tests. We can either leave it here, or put it in the coupler Wiki. Preferences?

LenkaNovak: Offline, we decided to hook this up to CI.

PR #373, opened 2023-08-02
fix BCReader.interpol

### LINKED ISSUE #372: `interpol` bug
In ClimaLSM's implementation of `BCReader.interpol,` we found an error in the logic of the math (see https://github.com/CliMA/ClimaLSM.jl/pull/282). 

We need to check the logic of the implementation in ClimaCoupler, and fix it if necessary. Either way we should add a test for this function to prevent this from happening in the future.

juliasloan25:
## Purpose 
Closes #372

## To-do
- [x] fix logic in `interpol`
- [x] add test

----

### DIFF
--- a/src/BCReader.jl
+++ b/src/BCReader.jl
@@ -273,9 +273,8 @@ or returns the first Field if interpolation is switched off.
 - Fields.field
 """
 function interpolate_midmonth_to_daily(date, bcf_info::BCFileInfo{FT}) where {FT}
-    if bcf_info.interpolate_daily && bcf_info.segment_length[1] > FT(0)
-        (; segment_length, segment_idx, all_dates, monthly_fields) = bcf_info
-
+    (; segment_length, segment_idx, all_dates, monthly_fields, interpolate_daily) = bcf_info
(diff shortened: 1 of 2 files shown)


src/BCReader.jl
```diff
@@ -273,9 +273,8 @@ or returns the first Field if interpolation is switched off.
 - Fields.field
 """
 function interpolate_midmonth_to_daily(date, bcf_info::BCFileInfo{FT}) where {FT}
-    if bcf_info.interpolate_daily && bcf_info.segment_length[1] > FT(0)
-        (; segment_length, segment_idx, all_dates, monthly_fields) = bcf_info
-
+    (; segment_length, segment_idx, all_dates, monthly_fields) = bcf_info
+    if bcf_info.interpolate_daily && bcf_info.segment_length[1] > FT(0) && date != all_dates[Int(segment_idx[1])]
```
LenkaNovak, line 277: We could extract `interpolate_daily` and use `segment_length` from above, so we don't have to use `bcf_info.<...>`

LenkaNovak (approved):
LGTM, thanks, Julia!

src/BCReader.jl
juliasloan25: Good point!

PR #387, opened 2023-08-10
Enable DSS for Slab Models

### LINKED ISSUE #381: Apply dss to slab model Y+dY
Address issues in #380 in all slab models.

## Components
- [x] implement dss! hooks in model-specific init files for ocean and sea sea ice slab models
- [x] note difference in surface fluxes - expecting a smoother field
- [x] ensure Buildkite drivers produce physically sensible results, note performance changes

### Below we use an example of the idealized helem=4 aquaplanet 
- 1. surface flux differences - fairly subtle (compared to the land ones in #380 )

energy fluxes without / with DSS (W/m2)
  
Screen Shot 2023-08-11 at 1 05 49 PM

difference

Screen Shot 2023-08-11 at 1 07 24 PM

- 2. performance - small increase in allocs and walltime (as expected from additional communication)
perf_dss

LenkaNovak:
## Purpose 

Closes #381 (see for details and QA assessment)

----


experiments/AMIP/modular/components/slab_utils.jl
```diff
@@ -1,19 +1,23 @@
...
 get_slab_energy(slab_sim, T_sfc) =
     slab_sim.integrator.p.params.ρ .* slab_sim.integrator.p.params.c .* T_sfc .* slab_sim.integrator.p.params.h
+
+"""
+      dss!(Y::ClimaCore.Fields.FieldVector, p::NamedTuple, t)
+
+Computes the weighted direct stiffness summation and updates `Y` in place.
+In the case of a column domain, no dss operations are performed.
+"""
+function dss!(Y::ClimaCore.Fields.FieldVector, p::NamedTuple, t)
```
akshaysridhar, line 15: `ClimaLSM.jl` and `ClimaCore.jl` also define `dss!` @sriharshakandala is this is a problem ? 

See : https://github.com/CliMA/ClimaLSM.jl/blob/81af155b52cb1ed81af91076f78ab1e2722e1915/src/shared_utilities/utils.jl#L88C15-L88C15

experiments/AMIP/modular/components/slab_utils.jl
```diff
@@ -1,19 +1,23 @@
...
+    for key in propertynames(Y)
+        field = getproperty(Y, key)
+        buffer = get_dss_buffer(axes(field), p)
+        ClimaCore.Spaces.weighted_dss!(field, buffer)
```
akshaysridhar, line 19: LGTM

experiments/AMIP/modular/components/slab_init.jl
```diff
@@ -1,146 +0,0 @@
-# slab_rhs! (TODO: rm during clean up)
-using ClimaCore
```
akshaysridhar, line 2: Not entirely familiar with all the files in the previous iterations: What was the functionality within this, and where exactly in the new modifications is it retained?

LenkaNovak: `slab_init.jl` used to contain a general slab model initialization interface. Some time ago, we decided to use separate slab ocean and slab sea ice models, because their formulation diverged. So this slab model template became redundant. In this PR, the `slab_utils.jl` was revamped to contain only the currently used common functions for the two models, including the `dss!`.

experiments/AMIP/modular/components/slab_utils.jl
LenkaNovak: adding a type was throwing an type error - not sure how `ClimaTimeSteppers` is handling the `integrator.t` (something we should look into separately), but `t` isn't used by this function, so that's why I removed it. We can also rename this function, since it's used directly only locally.

akshaysridhar: Can we remove `t` from the args entirely ?

experiments/AMIP/modular/components/slab_utils.jl
sriharshakandala: If we are calling `weighted_dss` internally, I would recommend prefixing `weighted` to this specialized function name!

sriharshakandala: Is there a reason we are passing argument `t` if we are not using it inside the function?

experiments/AMIP/modular/components/slab_utils.jl
LenkaNovak: Sure, I can do `dss_slab!(Y::ClimaCore.Fields.FieldVector, p::NamedTuple, _)`. We need to retain the third argument due to [this](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/c104553f695dbdbc229173943eb82752c39a5356/src/functions.jl#L9).

experiments/AMIP/modular/components/slab_utils.jl
LenkaNovak: Sounds good, thanks (see comment above)

experiments/AMIP/modular/components/slab_utils.jl
akshaysridhar: This might be a reasonable compromise - @sriharshakandala perhaps you can look of the CTS interface here in relation to this PR; LGTM otherwise.

experiments/AMIP/modular/components/slab_utils.jl
sriharshakandala: If we are actually calling `weighted_dss` function inside, can we also clarify this in the calling function name! (e.g. `weighted_dss_slab1` or so. We currently support both pure `dss` and `weighted_dss` (which combines the effect of `M^-1 x dss`)in `ClimaCore.jl` and it would be good to disambiguate!

sriharshakandala: Also, is there a reason we need to add a third argument/`_` here?

sriharshakandala:
As discussed, since the primary objective of this PR is to add the buffer, I am ok with renaming the function and addressing the unused third argument issue a bit later since it is entangled with `CliMATimeSteppers.jl`

sriharshakandala (approved):

sriharshakandala:
Generally speaking, the application of DSS is a part of the RHS /tendency function! However, it seems to be isolated and treated differently than usual, which is probably the reason this ambiguity arises in the Time Stepper here!

experiments/AMIP/modular/components/slab_utils.jl
LenkaNovak: Sounds good, changed to `weighted_dss_slab!` and merging. Thank you both!

experiments/AMIP/modular/components/slab_utils.jl
akshaysridhar: Thanks @sriharshakandala @LenkaNovak : Following offline discussion, since we agreed to modify the function name, we can start a separate issue with the `t` arg in the `dss!` function and address it in a separate PR. With rebase this PR looks reasonable to me!

akshaysridhar (approved):

LenkaNovak:
> As discussed, since the primary objective of this PR is to add the dss buffer, I am ok with renaming the function and addressing the unused third argument issue a bit later since it is entangled with `CliMATimeSteppers.jl`.
> 
> Generally speaking, the application of DSS is a part of the RHS /tendency function computation! However, it seems to be isolated and treated differently than usual, which is probably the reason this ambiguity arises in the Time Stepper here!

That's how I understand it too. Here it is called as a callback - I guess some models (and their tendencies) could be pointwise and be independent of ClimaCore. The coupler could then use the ClimaCore Field objects to apply the models spatially, so in this case we'd definitely need a callback. In the slab ocean / Bucket cases, maybe either method works (as long as it's ok for the dss to be applied only once and after the stepping, rather during each substep)?

Issue #202, opened 2022-12-22
Update coupler to handle land albedo from file


juliasloan25:
ClimaLSM has been updated to read in surface albedo from a data file, rather than passing in a function for it as before (see ClimaLSM https://github.com/CliMA/ClimaLSM.jl/pull/125). 

The coupler should be updated to have a command line option `albedo_from_file` to specify whether we want to read in the albedo from a data file, with the default being true. The code relating to the bucket model in the coupler will also need to be updated to reflect the changes in ClimaLSM.

Addressed by #197

LenkaNovak:
addressed by #197

PR #394, opened 2023-08-17
Use time-varying albedo in bucket

### LINKED ISSUE #318: Time-varying albedo
We currently have the capability to use albedo read in from a file thanks to functionality provided by ClimaLSM. However, the albedo is read only once from the supplied data file, and does not change over time. Here, we want to extend our implementation to read in albedo as it changes over time.

https://github.com/CliMA/ClimaLSM.jl/issues/243 implements the necessary functionality in ClimaLSM, and we just have to update to use this new option.

Part of #276 
Closed by #394

### Relevant Links
- `BulkAlbedoMap` definition [here](https://github.com/CliMA/ClimaLSM.jl/blob/ccb194aa6ba6d6750f60d11f664e45b6af716a1e/src/Bucket/Bucket.jl#L99)
- `MapInfo` definition [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/Regridder.jl#L357)
- Albedo artifact files [here](https://github.com/CliMA/ClimaLSM.jl/blob/v0.2.5/src/Bucket/artifacts/artifacts.jl)
  - Bare ground albedo dataset (no time dimension)
  - CESM2 albedo dataset (midmonth dates from 1/15/1850 to 12/15/2014)
  - Both datasets contain variables "lat" (dim. 192), "lon" (dim. 288), and "sw_alb" (dim. lat x lon)
    - Note that our SST and SIC datasets have lat and lon dimensions of 180 and 360, respectively

### Other
- Note that the date range in the CESM2 albedo dataset (1850-2014) is different from the range of our SST and SIC datasets (1870-2022). This is fine for now since we're running shorter-length simulations, but we may need to revisit this when we want to have longer runs.

juliasloan25:
## Purpose 
Closes #318 

Edit: The interface for `ClimaLSM.Bucket.BulkAlbedoTemporal` needed to change for compatibility with the coupler. This change has been made in https://github.com/CliMA/ClimaLSM.jl/pull/300. ClimaLSM.jl has made an additional patch release for this change, bringing it up to v0.3.1.

## To Do
The changes made here will be similar to those made in #202 for reading in albedo from a constant map
- [x] Use ClimaLSM ~v0.3.0~ v0.3.1
- [x] Add option in cli_options.jl
  - [x] Change type from Bool to String
- [x] Add case for temporal map albedo in bucket_init.jl
- [x] Update buildkite runs to use String `albedo_type` option
- [x] Add test for this case in buildkite

----


experiments/AMIP/modular/cli_options.jl
```diff
@@ -31,10 +31,10 @@ function argparse_settings()
         help = "Method to partition turbulent fluxes. [`PartitionedStateFluxes`, `CombinedStateFluxes`]"
         arg_type = String
         default = "CombinedStateFluxes"
-        "--albedo_from_file"
-        help = "Access land surface albedo information from data file"
-        arg_type = Bool
-        default = true
+        "--albedo_type"
+        help = "Access land surface albedo information from data file. [`function`, `map_static`, `map_temporal`]"
```
valeriabarra, line 35: Perhaps `function` is not the best name here. I can think of some alternatives. What about `spatially_varying` ? I guess `function` does not sound too clear because we're still reading these data from a file (as it used to be the case before), whereas "function" to me sounds more like we would define/prescribe this rather than reading it from a file

perf/Manifest.toml
```diff
@@ -101,9 +101,9 @@ version = "0.4.2"
 
 [[deps.BandedMatrices]]
 deps = ["ArrayLayouts", "FillArrays", "LinearAlgebra", "PrecompileTools", "SparseArrays"]
-git-tree-sha1 = "6e9b384d5e21452bc81cae3552c74f430e320985"
+git-tree-sha1 = "106122bad983acd8e21aef3f7a7f91a3ac739077"
 uuid = "aae01518-5342-5314-be14-df237901396f"
-version = "0.17.35"
+version = "0.17.36"
```
valeriabarra, line 106: Are all these other package updates automatically triggered when you `] up ClimaLSM`? Or you updated the entire environment?

experiments/AMIP/modular/cli_options.jl
LenkaNovak: I think function is the idealized function of latitude here, but we could use `idealized` like Atmos does with Insolation.

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -181,14 +184,18 @@ function bucket_init(
     earth_param_set = create_lsm_parameters(FT)
 
     α_snow = FT(0.8) # snow albedo
-    if albedo_from_file # Read in albedo from data file (default)
-        albedo = BulkAlbedoMap{FT}(regrid_dirpath, α_snow = α_snow, comms = comms_ctx)
-    else # Use spatially-varying function for surface albedo
+    if albedo_type == "map_static" # Read in albedo from static data file (default)
+        albedo = BulkAlbedoStatic{FT}(regrid_dirpath, comms = comms_ctx)
+    elseif albedo_type == "map_temporal"
+        albedo = BulkAlbedoTemporal{FT}(regrid_dirpath, date_ref, t_start, space)
```
LenkaNovak, line 190: I would add a bit more info on the files and data (is this linearly interpolated monthly data? What is the source of  the file?).

LenkaNovak (commented):
LGTM, thanks Julia! Buildkite output looks good, but I noticed that the logging doesn't display the run command log (apparently the limit was exceeded). From your local runs, is it obvious why that is? (e.g., lots of warning messages?)

perf/Manifest.toml
juliasloan25: Ahh good point, I'll revert them and just up ClimaLSM

juliasloan25:
> LGTM, thanks Julia! Buildkite output looks good, but I noticed that the logging doesn't display the run command log (apparently the limit was exceeded). From your local runs, is it obvious why that is? (e.g., lots of warning messages?)

@LenkaNovak I've actually been getting a `failed process` error when I run our driver locally. It happens in `rll_mesh` when `land_fraction` is being created. This used to happen sporadically, but recently (for the last few months), it happens every time. I can try running it on the cluster, but it would also be nice to figure out why this happens locally.

valeriabarra:
> > LGTM, thanks Julia! Buildkite output looks good, but I noticed that the logging doesn't display the run command log (apparently the limit was exceeded). From your local runs, is it obvious why that is? (e.g., lots of warning messages?)
> 
> @LenkaNovak I've actually been getting a `failed process` error when I run our driver locally. It happens in `rll_mesh` when `land_fraction` is being created. This used to happen sporadically, but recently (for the last few months), it happens every time. I can try running it on the cluster, but it would also be nice to figure out why this happens locally.

Is it the same as these [failures](https://buildkite.com/clima/climacoupler-ci/builds/1655#018a24f5-a07c-42bb-b7a2-3fc64af0dcf2) that I've seen in my other `up_deps.jl` [PR](https://github.com/CliMA/ClimaCoupler.jl/pull/410)?

.buildkite/pipeline.yml
```diff
@@ -154,6 +154,13 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: "Slabplanet: albedo from temporal map"
+        key: "slabplanet_albedo_maptemporal"
+        command: "julia --color=yes --project=experiments/AMIP/modular/ experiments/AMIP/modular/coupler_driver_modular.jl --run_name slabplanet_albedo_maptemporal --FLOAT_TYPE Float64 --coupled true --surface_setup PrescribedSurface  --moist equil --vert_diff true --rad gray --energy_check true --mode_name slabplanet --t_end 10days --dt_save_to_sol 9days --dt_cpl 200 --dt 200secs --mono_surface true --h_elem 4 --precip_model 0M --anim true --albedo_type map_temporal --job_id slabplanet_albedo_maptemporal"
```
juliasloan25, line 159: @LenkaNovak (starting a thread so we can add comments) - I realized I left 2 `@show` statements in the ClimaLSM code that updates the albedo at each timestep ([here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/standalone/Bucket/Bucket.jl#L640C1-L641)). I think this is what's causing this run to exceed the log limit. I can fix it in ClimaLSM and make another patch release.

juliasloan25: Should I remove the `@show`s before merging this PR, or do you think it's okay to merge this as is for now and fix ClimaLSM later?

valeriabarra: I would personally remove the first, bump a patch release and update this PR again.

experiments/AMIP/modular/cli_options.jl
LenkaNovak: On balance, it's probably better to be consistent with ClimaLSM.

LenkaNovak:
> > LGTM, thanks Julia! Buildkite output looks good, but I noticed that the logging doesn't display the run command log (apparently the limit was exceeded). From your local runs, is it obvious why that is? (e.g., lots of warning messages?)
> 
> @LenkaNovak I've actually been getting a `failed process` error when I run our driver locally. It happens in `rll_mesh` when `land_fraction` is being created. This used to happen sporadically, but recently (for the last few months), it happens every time. I can try running it on the cluster, but it would also be nice to figure out why this happens locally.

Hmm, that's odd. Let's open a new issue on this then and address it separately.

experiments/AMIP/modular/cli_options.jl
juliasloan25: Okay, I'll leave it like this for now

.buildkite/pipeline.yml
juliasloan25: Okay, I removed them in ClimaLSM, released ClimaLSM v0.3.2, and updated this PR to use that version. It should be fixed now

LenkaNovak (approved):
LGTM, thank you, @juliasloan25 !

PR #418, opened 2023-09-12
Split coupler and atmos configs

### LINKED ISSUE #417: split coupler atmos configs
After #388 we will want to split atmos and coupler configuration, so different toml files / set of dictionaries can be used to initialize Atmos, as well as the defaults. The coupler defaults are still in the `cli_options.jl` and called as `parsed_args`, but now the `parsed_args` get overwritten by the setup args in the config toml files. Whatever args are specified in the coupler also overwrite the args in the component models.

- [x] split the files in the driver config
- [x] test that non-default atmos params can be called
- [x] ensure changes applied in `pipeline.yml`, `longruns/pipeline.yml` (delete the perf configs, if possible)

LenkaNovak:
## Purpose 

Closes #417 

----


LenkaNovak:
moved to the rebased #431

PR #420, opened 2023-09-14
Fix water conservation check

### LINKED ISSUE #404: check water conservation (incl. snow)
We need to check the water conservation logger (incl. snow), and apply appropriate allowable bounds. 

## Components
- [x] fix conservation checker / plotting
  - note that now Atmos uses the kg/m^2/s for evaporation, and m/s for precipitation!
- [x] add snow - small errors remain - see #421 
- [x] generalize ConservationChecker, to allow for more general tests and extensible development (drop the dependency on underlying models' data structures)  #436 
- [x] add unit test for water
- [x] apply allowable bounds to unit tests
- [x] remove the usage of limiters, which cause worse water conservation, see below (atmos default also doesn't use them)

# QA
- [x] unit tests pass to FP precision
- [x] no change in performance 
- [x] slabplanet conserves water and energy (though not to FP precision - see above)
Screen Shot 2023-09-26 at 8 15 10 AM

Screen Shot 2023-09-26 at 8 15 40 AM

Note, if we apply limiters, water conservation breaks:

Screen Shot 2023-09-26 at 9 47 51 AM

LenkaNovak:
## Purpose 

Addressed in #436

----


LenkaNovak:
Addressed in #436

PR #431, opened 2023-09-21
split atmos and coupler configs - rebased

LenkaNovak:
## Purpose 

Closes https://github.com/CliMA/ClimaCoupler.jl/issues/417

(rebased from #418 )

----

### DIFF
--- a/experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
+++ b/experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
@@ -21,10 +21,32 @@ struct ClimaAtmosSimulation{P, Y, D, I} <: AtmosModelSimulation
 end
 name(::ClimaAtmosSimulation) = "ClimaAtmosSimulation"
 
-function atmos_init(::Type{FT}, parsed_args::Dict) where {FT}
+"""
+    get_atmos_config(coupler_dict::Dict)
+
+Returns the specified atmospheric configuration (`atmos_config_dict`) overwitten by arguments
+in the coupler dictionary (`config_dict`).
+"""
+function get_atmos_config(coupler_dict)
+    atmos_file = coupler_dict["atmos_config_file"]
+    # override default or specified configs with coupler arguments, and set the correct atmos config_file
+    if isnothing(atmos_file)
+        @info "Using Atmos default configuration"
+        merge(CA.default_config_dict(), coupler_dict, Dict("config_file" => atmos_file))
+    else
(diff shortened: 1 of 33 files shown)


experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -21,10 +21,32 @@ struct ClimaAtmosSimulation{P, Y, D, I} <: AtmosModelSimulation
...
+function get_atmos_config(coupler_dict)
+    atmos_file = coupler_dict["atmos_config_file"]
+    # override default or specified configs with coupler arguments, and set the correct atmos config_file
+    if isnothing(atmos_file)
+        @info "Using Atmos default configuration"
+        merge(CA.override_default_config(atmos_file), coupler_dict, Dict("config_file" => atmos_file))
+    else
+        @info "Using Atmos configuration from $atmos_file"
+        merge(
+            CA.override_default_config(joinpath(pkgdir(CA), atmos_file)),
```
juliasloan25, line 39: Why do we need to call `joinpath` here but not in the `if` branch?

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -3,6 +3,7 @@ using ClimaCore
 using ClimaLSM
 import ClimaLSM
 import ClimaTimeSteppers as CTS
+import Thermodynamics as TD
```
juliasloan25, line 6: Do we need this? I don't see `TD` used in this file

juliasloan25 (approved):
LGTM!

experiments/AMIP/modular/components/land/bucket_init.jl
LenkaNovak: Ahh, it's used in the `bucket_utils.jl`

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
LenkaNovak: In the if part, `atmos_file = nothing`, so this [override function](https://github.com/CliMA/ClimaAtmos.jl/blob/e2268a7b3c9920df5afa10484672f13f5a6e6d81/src/solver/yaml_helper.jl#L63) defaults to `CA.default_config_dict()`. But you're right that this is not obvious. I'll chage this to `CA.default_config_dict()`.

LenkaNovak:
Thank you for the review, @juliasloan25 !

PR #436, opened 2023-09-25
Revamp ConservationChecker

### LINKED ISSUE #404: check water conservation (incl. snow)

LenkaNovak:
## Purpose 

Closes #404 

----

### DIFF
--- a/experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
+++ b/experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
@@ -71,9 +71,14 @@ end
 # extensions required by the Interfacer
 get_field(sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux}) =
     Fields.level(sim.integrator.p.ᶠradiation_flux, half)
-get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation}) =
-    sim.integrator.p.col_integrated_rain .+ sim.integrator.p.col_integrated_snow # all fallen snow melts for now
-get_field(sim::ClimaAtmosSimulation, ::Val{:snow_precipitation}) = sim.integrator.p.col_integrated_snow .* FT(0)
+function get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation})
+    ρ_liq = CAP.ρ_cloud_liq(sim.integrator.p.params)
+    sim.integrator.p.col_integrated_rain .* ρ_liq # kg/m^2/s
+end
+function get_field(sim::ClimaAtmosSimulation, ::Val{:snow_precipitation})
+    ρ_liq = CAP.ρ_cloud_liq(sim.integrator.p.params)
+    sim.integrator.p.col_integrated_snow .* ρ_liq  # kg/m^2/s
+end
 
 get_field(sim::ClimaAtmosSimulation, ::Val{:turbulent_energy_flux}) =
     Geometry.WVector.(sim.integrator.p.sfc_conditions.ρ_flux_h_tot)
@@ -252,3 +257,46 @@ Extension of Checkpointer.get_model_state_vector to get the model state.
 function get_model_state_vector(sim::ClimaAtmosSimulation)
     return sim.integrator.u
 end
+
+"""
+    get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:F_radiative_TOA})
+
+Extension of Interfacer.get_field to get the net TOA radiation, which is a sum of the
+upward and downward longwave and shortwave radiation.
+"""
+function get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:F_radiative_TOA})
+    radiation = atmos_sim.integrator.p.radiation_model
+    FT = eltype(atmos_sim.integrator.u)
+    # save radiation source
+    if radiation != nothing
+        face_space = axes(atmos_sim.integrator.u.f)
+        z = parent(Fields.coordinate_field(face_space).z)
+        Δz_top = round(FT(0.5) * (z[end, 1, 1, 1, 1] - z[end - 1, 1, 1, 1, 1]))
+        n_faces = length(z[:, 1, 1, 1, 1])
+
+        LWd_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_lw_flux_dn), face_space),
+            n_faces - half,
+        )
+        LWu_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_lw_flux_up), face_space),
+            n_faces - half,
+        )
+        SWd_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_sw_flux_dn), face_space),
+            n_faces - half,
+        )
+        SWu_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_sw_flux_up), face_space),
+            n_faces - half,
+        )
+
+        return @. -(LWd_TOA + SWd_TOA - LWu_TOA - SWu_TOA) # [W/m^2]
+    else
+        return FT(0)
+    end
+end
+
+get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:energy}) = atmos_sim.integrator.u.c.ρe_tot
+
+get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:water}) = atmos_sim.integrator.u.c.ρq_tot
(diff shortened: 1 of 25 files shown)


experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -71,9 +71,14 @@ end
 # extensions required by the Interfacer
 get_field(sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux}) =
     Fields.level(sim.integrator.p.ᶠradiation_flux, half)
-get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation}) =
-    sim.integrator.p.col_integrated_rain .+ sim.integrator.p.col_integrated_snow # all fallen snow melts for now
-get_field(sim::ClimaAtmosSimulation, ::Val{:snow_precipitation}) = sim.integrator.p.col_integrated_snow .* FT(0)
+function get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation})
+    ρ_liq = CAP.ρ_cloud_liq(sim.integrator.p.params)
+    sim.integrator.p.col_integrated_rain .* ρ_liq # kg/m^2/s
```
juliasloan25, line 76: Will we be able to remove this access to `sim.integrator.p.col_integrated_rain` and generalize it as well at some point in the future?

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -252,3 +257,40 @@ Extension of Checkpointer.get_model_state_vector to get the model state.
...
+        SWd_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_sw_flux_dn), face_space),
+            n_faces - half,
+        )
+        SWu_TOA = Fields.level(
+            RRTMGPI.array2field(FT.(atmos_sim.integrator.p.radiation_model.face_sw_flux_up), face_space),
+            n_faces - half,
+        )
+
+        return @. -(LWd_TOA + SWd_TOA - LWu_TOA - SWu_TOA) # [W/m^2]
```
juliasloan25, line 288: Could you add a comment like "Net radiative flux is difference between the combined longwave and shortwave downward fluxes and the upward fluxes"?

experiments/AMIP/modular/components/land/bucket_utils.jl
```diff
@@ -20,6 +14,13 @@ function get_land_energy(bucket_sim::BucketSimulation, e_per_area)
     return e_per_area
 end
 
+function get_field(bucket_sim::BucketSimulation, ::Val{:water})
+    ρ_cloud_liq = ClimaLSM.LSMP.ρ_cloud_liq(bucket_sim.model.parameters.earth_param_set)
+    return
+    @. (bucket_sim.integrator.u.bucket.σS + bucket_sim.integrator.u.bucket.W + bucket_sim.integrator.u.bucket.Ws) *
```
juliasloan25, line 20: Could you add a comment about what this equation is doing? Maybe explaining what sigmaS, W, and Ws are

experiments/AMIP/modular/components/ocean/slab_ocean_init.jl
```diff
@@ -157,3 +157,8 @@ Extension of Checkpointer.get_model_state_vector to get the model state.
 function get_model_state_vector(sim::SlabOceanSimulation)
     return sim.integrator.u
 end
+
+get_field(sim::SlabOceanSimulation, ::Val{:energy}) =
+    sim.integrator.p.params.ρ .* sim.integrator.p.params.c .* sim.integrator.u.T_sfc .* sim.integrator.p.params.h
```
juliasloan25, line 162: Could you also explain briefly where this equation comes from (and same for sea ice energy below)?

src/ConservationChecker.jl
```diff
@@ -16,193 +16,184 @@ using UnPack
 using Plots
 using ClimaAtmos: RRTMGPI
 using ClimaLSM
-using ..Utilities: CoupledSimulation, swap_space!
-using ..Interfacer: SurfaceStub
+using ClimaCoupler.Utilities: CoupledSimulation, swap_space!
+import ClimaCoupler: Interfacer
```
juliasloan25, line 20: I think Charlie has said it's better to use `import ..Interfacer` since that makes sure we use the dev version of the package we're working on, rather than potentially using the latest release of main. But there are a lot of places we aren't doing that, so maybe we could have another PR that goes through and changes that.

src/ConservationChecker.jl
```diff
@@ -16,193 +16,184 @@ using UnPack
...
 """
-struct EnergyConservationCheck{A} <: AbstractCheck
-    ρe_tot_atmos::A
-    ρe_tot_land::A
-    ρe_tot_ocean::A
-    ρe_tot_seaice::A
-    toa_net_source::A
-    ice_base_source::A
+mutable struct EnergyConservationCheck <: AbstractConservationCheck
+    sums::NamedTuple
```
juliasloan25, line 33: Should this be `sims`? or is `sums` correct?

src/ConservationChecker.jl
```diff
@@ -16,193 +16,184 @@ using UnPack
...
+            push!(ccs.toa_net_source, radiation_sources_accum)
+
+            # save atmos
+            previous = getproperty(ccs, sim_name)
+            current = sum(Interfacer.get_field(sim, Val(:energy))) # # ∫ J / m^3 dV
+
+            push!(previous, current)
+            total += current + radiation_sources_accum
+
+        elseif sim isa Interfacer.SurfaceModelSimulation || sim isa Interfacer.SurfaceStub
```
juliasloan25, line 115: I believe we have `SurfaceStub <: SurfaceModelSimulation`, so this condition could be `elseif sim <: Interfacer.SurfaceModelSimulation`

test/conservation_checker_tests.jl
```diff
@@ -14,300 +14,181 @@ using NCDatasets
...
+    Δt = cs.Δt_cpl
+
+    # analytical solution
+    tot_energy_an = sum(F_r .* 3Δt .+ 1e6 .* 1.25)
+    tot_water_an = sum(.-P .* 3Δt .* 0.5 .+ Fields.ones(space))
+
+    # run check_conservation!
+    check_conservation!(cs, runtime_check = true)
+    check_conservation!(cs, runtime_check = true)
+    check_conservation!(cs, runtime_check = true)
```
juliasloan25, line 103: why do we have 3 calls to this in a row?

juliasloan25 (approved):
Looks good! I had a few questions, mostly about clarifying comments :)

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
LenkaNovak: the `..._init.jl` files is what we'd like the component models to adopt, so eventually this will go to ClimaAtmos. I think this won't be until the Atmos interface is modularized and tested though. We definitely don't want our `src/` code to contain any model specific data structures, and this is the last module that contained them! 😅

src/ConservationChecker.jl
LenkaNovak: Sounds good. When developing locally, I usually use `Pkg` to call `dev` on the entire local Coupler package, but happy to change to this (as a separate issue), if that makes development easier for you. :)

src/ConservationChecker.jl
LenkaNovak: Ahh well spotted! You're right, `Interfacer.SurfaceModelSimulation` should be enough.

test/conservation_checker_tests.jl
LenkaNovak: We call them several times to let any potential errors accrue.

src/ConservationChecker.jl
LenkaNovak: it's sums of the sims in fact - originally I had called it fields, but then didn't want it to be confused with `ClimaCore.Fields.Field`s.

PR #438, opened 2023-09-27
update dependencies

### LINKED ISSUE #433: Update for new dep versions
ClimaCore, ClimaAtmos, and ClimaLSM both have new releases, so we need to update for them

ClimaCore v0.10.51, ClimaAtmos v0.16.1, ClimaLSM v0.3.3

juliasloan25:
## Purpose 
update dependencies for new releases of ClimaCore, ClimaAtmos, ClimaLSM

closes #433 

## To-do
- [x] update dependencies for ClimaCore, ClimaAtmos, ClimaLSM
- [x] update compats for Thermodynamics, SurfaceFluxes #414 #415
- [x] remove some duplicated compat entries

note: I wanted to update the compat for SciMLBase to "2" here too, but can't because the compat in ClimaAtmos and ClimaLSM is restricted to "1" #434

----


juliasloan25:
bors try

experiments/AMIP/modular/Manifest.toml
LenkaNovak, line 1: 👍 includes all new deps we want for this experiment

test/Project.toml
LenkaNovak, line 1: Since we're not checking in `test/Manifest.toml` would it make sense to specify the specific versions in `test/Project.toml`?

LenkaNovak (approved):
Thanks, @juliasloan25 !Buildkite runs show a slight change in physical runs in slabplanet and AMIP, but this is to be expected given the release changes in ClimaAtmos and ClimaLSM. Performance is improved, so LGTM.

test/Project.toml
juliasloan25: My understanding is that we don't check in the test Manifest so that the tests are run with the most updated versions of the packages (at least if someone runs `]up`). The test environment will still use the compat restrictions specified in the top-level Project, so it would test the maximum allowed versions based on those constraints

PR #439, opened 2023-09-28
upgrade pipelines to CUDA 12.2

### LINKED ISSUE #440: Fix CUDA error in CI
We've been occasionally seeing an error `InitError: CUDA error (code 999, CUDA_ERROR_UNKNOWN)` in CI. We think this may be due to CPU jobs running on (unrequested) GPUs. This PR updates to a newer version of CUDA, which will hopefully prevent this error.

Closed by #439

juliasloan25:
## Purpose 
Closes #440

See analogous PR in ClimaAtmos: https://github.com/CliMA/ClimaAtmos.jl/pull/2169

## To-do
- [x] update CUDA specifications to 12.2

----


.buildkite/longruns/pipeline.yml
```diff
@@ -2,7 +2,7 @@ env:
   JULIA_VERSION: "1.9.3"
   MPI_IMPL: "openmpi"
   OPENMPI_VERSION: "4.1.1"
-  CUDA_VERSION: "11.2"
+  CUDA_VERSION: "12.2"
```
simonbyrne, line 5: I've moved away from specifying everything in env variables. This will probably require a bigger change

simonbyrne (approved):

.buildkite/longruns/pipeline.yml
juliasloan25: I can open an issue for this in the coupler. Is it just openmpi and cuda that should be removed from the env vars? It looks like Atmos still has a fair amount of env vars: https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/pipeline.yml#L6-L17

simonbyrne: i made some changes to how the configuration works, to make it a little more explicit (basically you have to explicitly list the modules you want to load). I can help you with it tomorrow

PR #442, opened 2023-09-29
add dss_state! function

### LINKED ISSUE #435: Check that initial conditions are continuous fields
All tendencies must be continuous fields. We ensure this in the horizontal by calling DSS in each model at each timestep, and we ensure this in the vertical by providing continuous inputs. Since we don't have a function that can check if an input field is continuous, we should dss each of the initial states to make sure our initial conditions produce continuous fields.

See Simon's comment [here](https://github.com/CliMA/ClimaCoupler.jl/issues/428#issuecomment-1728235304) for more details.

Addressed by #442

juliasloan25:
## Purpose 
Use new `dss_state!` function to check that initial conditions are continuous

Closes #435

----


experiments/AMIP/modular/coupler_driver_modular.jl
```diff
@@ -436,13 +437,19 @@ import_combined_surface_fields!(cs.fields, cs.model_sims, cs.boundary_space, tur
 import_atmos_fields!(cs.fields, cs.model_sims, cs.boundary_space, turbulent_fluxes)
 update_model_sims!(cs.model_sims, cs.fields, turbulent_fluxes)
 
-# 2) each surface component model calculates its own vapor specific humidity (q_sfc)
+# 2) call DSS on each of the simulation states to ensure our initial conditions are continuous fields
```
LenkaNovak, line 440: Would it be possible to dss right after initializing the models? The functions above this already create some cache variables that depend on the initialized states. We could also include these as part of the init functions.

src/Interfacer.jl
```diff
@@ -127,4 +128,48 @@ Returns simulation name, if defined, or `Unnamed` if not.
 name(::ComponentModelSimulation) = "Unnamed"
 name(::SurfaceStub) = "SurfaceStub"
 
+"""
```
LenkaNovak, line 131: Hmm. These `dss_state` calls contain data structures, like the `integrator`. We don't really want this in the source code, because it's not general - e.g., other models will have their own ways doing dss (or will not need it). It would be better to move all these to the respective init files.

test/interfacer_tests.jl
```diff
@@ -85,3 +92,68 @@ end
     @test parent(get_field(stub, Val(:area_fraction)))[1] == FT(1)
     @test parent(get_field(stub, Val(:surface_temperature)))[1] == FT(2)
 end
+
```
LenkaNovak, line 95: Instead of having tests, we could always do asserts in the init functions. What do you think?

LenkaNovak (changes requested):
Hi @juliasloan25 , thank you for this PR! Here are some comments on generality and placing. Happy to chat if needs be.

test/interfacer_tests.jl
juliasloan25: I'm not sure what we would assert. I was originally thinking of writing a function to check if a field is continuous, then assert that that function is true for each initialized field. But since it's nontrivial to check if a field is continuous, I don't know what we should assert if we take that approach

src/Interfacer.jl
juliasloan25: Good point! I'll move them to the init files

experiments/AMIP/modular/coupler_driver_modular.jl
juliasloan25: Yeah, I'll move the dss calls to the init functions

test/interfacer_tests.jl
juliasloan25: I think it makes sense to move the functions to the init files, but it would be nice to keep the tests for them too. I wonder if we should have a place to add tests for component model functions too. That might make it easier for the component model developers when they have to move these functions to their own repos

LenkaNovak: I see what you mean. We already have [component_model_tests.jl](https://github.com/CliMA/ClimaCoupler.jl/blob/main/test/component_model_tests.jl), which will be converted to a folder with the individual components in the sea ice PR. We can add it here.

experiments/AMIP/modular/components/land/bucket_init.jl
```diff
@@ -4,6 +4,7 @@ using ClimaLSM
 import ClimaLSM
 import ClimaTimeSteppers as CTS
 import Thermodynamics as TD
+using Dates: DateTime
```
LenkaNovak, line 7: Just checking that this is required here?

experiments/AMIP/modular/components/ocean/prescr_seaice_init.jl
```diff
@@ -2,6 +2,8 @@ import ClimaTimeSteppers as CTS
 import ClimaCoupler.Interfacer: SeaIceModelSimulation, get_field, update_field!, name
 import ClimaCoupler.FieldExchanger: step!, reinit!
 import ClimaCoupler.FluxCalculator: update_turbulent_fluxes_point!
+using ClimaCoupler: Regridder
```
LenkaNovak, line 5: Again, checking that this is needed?

test/component_model_tests/bucket_tests.jl
```diff
@@ -0,0 +1,59 @@
...
+using ClimaCore: Fields, Spaces
+
+include(pkgdir(ClimaCoupler, "experiments/AMIP/modular/components/land/bucket_init.jl"))
+include(pkgdir(ClimaCoupler, "experiments/AMIP/modular/components/land/bucket_utils.jl"))
+
+# struct DummySimulationBucket{I} <: BucketSimulation
+#     integrator::I
+# end
+
+# TODO bucket doesn't currently work with Float32, but we want to eventually test with both FTs
```
LenkaNovak, line 13: 👍

experiments/AMIP/modular/components/land/bucket_init.jl
juliasloan25: Yes, it's used [here](https://github.com/CliMA/ClimaCoupler.jl/blob/115f3c37798ac368df1ae2e2eb9691badd630432/experiments/AMIP/modular/components/land/bucket_init.jl#L176) but wasn't being imported before. Maybe we were only calling this from the driver so we didn't need to explicitly import it? For the component model tests (and maybe for code quality?) we need to have it here now

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -300,3 +303,20 @@ end
...
+
+"""
+    dss_state!(sim::AtmosModelSimulation)
+
+Perform DSS on the state of a component simulation, intended to be used
+before the initial step of a run. This method acts on atmosphere simulations.
+These sims don't store a dss buffer in their cache, so we must allocate
+one here.
+"""
+function dss_state!(sim::AtmosModelSimulation)
```
LenkaNovak, line 315: This should be `ClimaAtmosSimulation`, since another atmos model may not require this.

experiments/AMIP/modular/components/ocean/prescr_seaice_init.jl
juliasloan25: Used [here](https://github.com/CliMA/ClimaCoupler.jl/blob/115f3c37798ac368df1ae2e2eb9691badd630432/experiments/AMIP/modular/components/ocean/prescr_seaice_init.jl#L81)

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
juliasloan25: oops, good point

test/component_model_tests/climaatmos_tests.jl
```diff
@@ -0,0 +1,40 @@
+using Test
+import ClimaCoupler
+using ClimaCoupler.Interfacer: AtmosModelSimulation
+using ClimaCoupler.TestHelper: create_space
+using ClimaCore: Fields, Spaces
+
+include(pkgdir(ClimaCoupler, "experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl"))
+
+struct DummySimulationAtmos{I} <: AtmosModelSimulation
```
LenkaNovak, line 9: This should be ClimaAtmosSimulation, otherwise we're not testing whether this works on ClimaAtmos states.

LenkaNovak (commented):
Hi Julia, this looks great! Just a couple of clarifications on package imports, and the Atmos dss + tests should really be dispatched on `ClimaAtmosSimulation`, since the component tests are supposed to test the specific component models. Otherwise the assumption would be that all future atmos models will need this dss function.

experiments/AMIP/modular/components/land/bucket_init.jl
LenkaNovak: Sounds good! 👍

LenkaNovak (approved):
LGTM

PR #454, opened 2023-10-02
replace SIC_init with SIC_current

### LINKED ISSUE #334: change SIC_init to SIC in the driver
This is a bug found by @juliasloan25. We're using `SIC_init` at each timestep rather than using the updated value read in from the data file https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/modular/coupler_driver_modular.jl#L506C7-L506C7

juliasloan25:
## Purpose 
closes #334 

## To-do
Buildkite checks
- [x] conservation checks (slabplanet)
  - slabplanet plots look identical to current main
- [x] no singificant change to performance (build_history)
- [x] reasonalble AMIP paperplot (note: there will be a small behavioral change in the AMIP run, but should be tiny.)
  - plots look identical to me (see below)

current [main](https://buildkite.com/clima/climacoupler-ci/builds/1825#018af399-ab2d-4f8c-abb1-bac4ac5eea8d) AMIP paperplot:
image

this PR AMIP paperplot:
image

----

### DIFF
--- a/experiments/AMIP/modular/coupler_driver_modular.jl
+++ b/experiments/AMIP/modular/coupler_driver_modular.jl
@@ -520,17 +520,15 @@ function solve_coupler!(cs)
             if cs.dates.date[1] >= next_date_in_file(cs.mode.SST_info)
                 update_midmonth_data!(cs.dates.date[1], cs.mode.SST_info)
             end
-            update_field!(
-                ocean_sim,
-                Val(:surface_temperature),
-                interpolate_midmonth_to_daily(cs.dates.date[1], cs.mode.SST_info),
-            )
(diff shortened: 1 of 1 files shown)


LenkaNovak (approved):
LGTM. Since the plots look identical, it would be good to double check that the SIC is being now being updated (even if it's just an interactive check where you compare the `SIC_init` and `SIC_current` fields). Otherwise I think it's good to go!

juliasloan25:
I've checked that `SIC_current` is getting updated and is different from `SIC_init`. I'll go ahead and rebase then merge if CI still looks good

PR #455, opened 2023-10-02
move CoupledSim to Interfacer

### LINKED ISSUE #333: Move `CoupledSimulation` to `Interfacer`
Moving the object `CoupledSimulation` from `ClimaCoupler.Utilities` to `ClimaCoupler.Interface` is more consistent with the current design and definition of other Simulation types.

juliasloan25:
## Purpose 
closes #333

### Components:
- [x] move the code
- [x] revamp unit tests

### QA
- [x] ensure there is no behavioral change in Buildkites (AMIP paperplots (namely the 32 day run), conservation check for slabplanet, performance in build_history)

### Questions:
- `src/CoupledSimulations/coupled_simulation.jl` [CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/CoupledSimulations/coupled_simulation.jl#L27) vs Interfacer.[CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/pull/455/files#diff-2d14afcf445ddbcfbab2975b494c7561f70fb6ec9ef569732d48aae65f6c61ffR35) - why is this type defined in two places?
- `timestepping.md` refers to ClimaCoupler.[CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/blob/main/docs/src/timestepping.md) - which of the two types is this?
- should we keep [AbstractSimulation](https://github.com/CliMA/ClimaCoupler.jl/pull/455/files#diff-2d14afcf445ddbcfbab2975b494c7561f70fb6ec9ef569732d48aae65f6c61ffR29) defined in Interfacer? This is also defined in `src/CoupledSimulations/coupled_simulation.jl` as well

----

### DIFF
--- a/docs/src/interfacer.md
+++ b/docs/src/interfacer.md
@@ -3,7 +3,7 @@
 This module contains functions for defining the interface for coupling component models, as well as stub objects that contain prescribed fields.
 
 ## Coupled Simulation
-- `CoupledSimulation` (`cs`, currently in Utilities - TODO) stores info for ESM run. We require that each `cs` contains four (`atmos_sim`, `land_sim`, `ocean_sim` and `ice_sim`) components. While this requirement will not be eventually needed, for the time being, if a simulation surface type is not needed for a given run, it should be initialized with `SurfaceStub` with a zero `area_fracion`. The `atmos_sim` should always be specified.
+- `CoupledSimulation` (`cs`) stores info for ESM run. We require that each `cs` contains four (`atmos_sim`, `land_sim`, `ocean_sim` and `ice_sim`) components. While this requirement will not be eventually needed, for the time being, if a simulation surface type is not needed for a given run, it should be initialized with `SurfaceStub` with a zero `area_fracion`. The `atmos_sim` should always be specified.
 
 ## Component model simulations
 - all Simulations that are not the `CoupledSimulation` fall under `ComponentModelSimulation`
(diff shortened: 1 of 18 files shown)


juliasloan25:
@LenkaNovak could you help me understand the questions I have in the PR description?

> - `src/CoupledSimulations/coupled_simulation.jl` [CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/CoupledSimulations/coupled_simulation.jl#L27) vs Interfacer.[CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/pull/455/files#diff-2d14afcf445ddbcfbab2975b494c7561f70fb6ec9ef569732d48aae65f6c61ffR35) - why is this type defined in two places?
> - `timestepping.md` refers to ClimaCoupler.[CoupledSimulation](https://github.com/CliMA/ClimaCoupler.jl/blob/main/docs/src/timestepping.md) - which of the two types is this?
> - should we keep [AbstractSimulation](https://github.com/CliMA/ClimaCoupler.jl/pull/455/files#diff-2d14afcf445ddbcfbab2975b494c7561f70fb6ec9ef569732d48aae65f6c61ffR29) defined in Interfacer? This is also defined in `src/CoupledSimulations/coupled_simulation.jl` as well

LenkaNovak:
Sure:
-  1.`src/CoupledSimulations/` files were placeholders for the `Simulations`-like interface. Since we're not going to implement this in the near future, we can probably remove it (as long as it doesn't clash with the sea_breeze example. If it does, then we can revamp this as part of a different PR that updates the `ClimaCore` examples). Same goes for `src/CouplerState/`
- 2. `timestepping.md` refers to this outdated code and should be removed. There will be more similar files to clean up for the release, so we should go over these on Monday. 
- 3. similar to point 1. The `AbstactSimulaiton` in `src/CoupledSimulations/` is outdated, but it may break the sea breeze case if we remove it without modifying the example's run script.

juliasloan25:
I think getting rid of the `src/CoupledSimulations` and `src/CoupledState` folders will change enough code that I want to do it in a separate PR: #458

LenkaNovak (approved):

PR #459, opened 2023-10-05
CO2 read

### LINKED ISSUE #408: Enable prescribed CO2
In AMIP II, CO2 is prescribed at 348 ppmv (see [requirements](https://pcmdi.llnl.gov/mips/amip/requirements.html#3.2%20Subproject)), but we want to demonstrate time varying CO2. Several options exist, but for now ClimaAtmos interface allows only one global value. Since CO2 is generally well mixed, we implement temporally varying global values (represented by the Mauna Loa record). This follows some CMIP experiments. At the next stage will allow CO2 to be a function latitude or use gridded files, as is done in the latest CMIP models. 

# components
- [x] ensure we can change radiation CO2 externally from ClimaAtmos
- [x] read global mean CO2 from file: https://gml.noaa.gov/webdata/ccgg/trends/co2/co2_mm_mlo.txt 
  - since general file reads will be dealt with in ClimaUtilities soon, here the CO2 prescription was implemented with minimal disruption to our src code. This meant:
    - [x] converting the .txt file into a netcdf file of the same format as the SST and SIC files
    - [x] saving the file as a new artifact on Box 
    - [x] enabled the specification in the driver
 
In ClimaUtilities, we want to deal with this more generally @juliasloan25 . We will need to
- agree on accepted formats for forcing / boundary condition files
  - formats of files with 3d and 2d fields (presumably netcdf with time, z, lat, lon makes most sense)
  - formats of files with 1d fields (CSV with standard headings, TBD)
  - time variable (hours since epoch? e.g. 1800-01-01 00:00:00) 
- come up with a better interface that reduces code in the driver

# QA
- [x] demonstrate change in Buildkite
  - with co2 https://buildkite.com/clima/climacoupler-ci/builds/1878
  - w/o co2 https://buildkite.com/clima/climacoupler-ci/builds/1873
- [x] come but not huge increase in walltime (1h42 vs 1h30 for 30 days) 

#Potential next steps
- add other gases and latitudinal dependence: https://gmd.copernicus.org/articles/10/2057/2017/
  - this requires enabling `atmos_sim.integrator.p.radiation_model.volume_mixing_ratio_co2` to be a 2d field
- add gridded CO2 data: https://www.nature.com/articles/s41597-022-01196-7
  - this requires enabling `atmos_sim.integrator.p.radiation_model.volume_mixing_ratio_co2` to be a 2d field

LenkaNovak:
closes #408 

----


experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -107,6 +107,14 @@ get_field(sim::ClimaAtmosSimulation, ::Val{:gas_constant_air}) =
 
 get_surface_params(sim::ClimaAtmosSimulation) = CAP.surface_fluxes_params(sim.integrator.p.params)
 
+function update_field!(atmos_sim::ClimaAtmosSimulation, ::Val{:co2_gm}, field)
+    if atmos_sim.integrator.p.radiation_model.radiation_mode isa CA.RRTMGPI.GrayRadiation
+        @warn "Gray radiation_modeladiation model initialized, skipping CO2 update"
```
juliasloan25, line 112: typo in warning message?

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
```diff
@@ -107,6 +107,14 @@ get_field(sim::ClimaAtmosSimulation, ::Val{:gas_constant_air}) =
 
 get_surface_params(sim::ClimaAtmosSimulation) = CAP.surface_fluxes_params(sim.integrator.p.params)
 
+function update_field!(atmos_sim::ClimaAtmosSimulation, ::Val{:co2_gm}, field)
```
juliasloan25, line 110: does "gm" in `co2_gm` here stand for "global mean"?

juliasloan25 (commented):
The code changes look good to me! 

I noticed that the CI time for this run is 1hr 42 mins... It would be nice to keep runs to less than 1 hour (or even shorter if we can). Would it be possible to change the length of this simulation, or move it to the longruns if we need it to run for this long? I think we should talk about this for some of the other runs too, but not as part of this PR

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
LenkaNovak: Good catch!

experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl
LenkaNovak: yes, adopting the syntax of ClimaAtmos

LenkaNovak:
> The code changes look good to me!
> 
> I noticed that the CI time for this run is 1hr 42 mins... It would be nice to keep runs to less than 1 hour (or even shorter if we can). Would it be possible to change the length of this simulation, or move it to the longruns if we need it to run for this long? I think we should talk about this for some of the other runs too, but not as part of this PR

Thank you, @juliasloan25 for the review! 🚀 

(as discussed offline, we will be splitting the CI runs as in #406 to speed up our workflow.)

PR #467, opened 2023-10-12
Fix CO2 date call 

LenkaNovak:
Fix the parsed date to the correct current date.

----


juliasloan25 (approved):
LGTM! Good catch

PR #468, opened 2023-10-13
move folders to exp/ClimaCore

### LINKED ISSUE #458: remove outdated src/ folders
We have two folders `src/CoupledSimulations/` and `src/CouplerState`, which contain placeholder code that isn't needed anymore. We should remove these folders, the files in them, and remove/modify files that refer to them as necessary.

More details here: https://github.com/CliMA/ClimaCoupler.jl/pull/455#issuecomment-1745629334

Update: Replacing this usage in the seabreeze experiment will be more involved than we expected, and isn't a huge priority right now. Instead of completely removing these two folders, we'll move the code in them to `experiments/ClimaCore/` to emphasize that they aren't part of the main codebase, but are still used for these experiments. Eventually we want these experiments to test our maintained functionality, but this is a temporary solution for now.

#468 moves these files into `experiments/ClimaCore/`

juliasloan25:
## Purpose 
closes #458 by using experiment-related folders from `src/` to `experiments/ClimaCore/`. Related tests are moved to a folder called `test/experiment_tests`.

### To Do
- [x] move `src/CoupledSimulations/`, `src/CouplerState/` to `experiments/ClimaCore/`
- [x] move related tests to to `test/experiment_tests/`
- [x] rename `CoupledSimulation` to avoid confusion with `Interfacer.CoupledSimulation`
- [x] update docs
- [x] update ClimaCoupler.jl
- [x] update experiments code loading to use relocated files

----


experiments/ClimaCore/sea-ice/slab_ice.jl
```diff
@@ -16,6 +16,9 @@ using OrdinaryDiffEq
 
 using Statistics
```
LenkaNovak, line 17: We can delete this file in fact. I don't this it contains much more than we already have in the eisenman CI runs rn. 🧊

test/Project.toml
```diff
@@ -19,6 +19,7 @@ MPIPreferences = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
```
LenkaNovak, line 22: Just curious, why did we not need PrettyTables before?

test/experiment_tests/clock.jl
```diff
@@ -1,5 +1,7 @@
 using Test
```
LenkaNovak, line 1: would it be possible to move the `clock.jl` and `coupler_state.jl` tests to one file that corresponds to `experiments/ClimaCore/CoupledSimultations/coupled_simulation.jl`, so it's more contained? We could also retain the `Sim` syntax for the filenemes too.

LenkaNovak (approved):
LGTM Julia, just a few minor comments! Lmk if any clarification is needed :)

test/Project.toml
juliasloan25: Before, it was in our main top-level Project. That worked because `coupled_simulation.jl` and `clock.jl` were in `src/`, and the tests just loaded their modules. Now that they're moved to `experiments/`, I'm loading the files from the tests (e.g. [here](https://github.com/CliMA/ClimaCoupler.jl/pull/468/files#diff-b6c291f3d19e31f4ed167224c95d3c7f39bee2eff5a18bd2813d5ff370b6a935R4)), so we need to add PrettyTables to both the experiments and test environments. It's only used once [here](https://github.com/CliMA/ClimaCoupler.jl/pull/468/files#diff-23d845a9405e9ab3c4883643314f9b5e5b1213b21a4d7a05430857b47bbce7c2R182), so I would be open to removing that line and then removing these dependencies, but I'm not sure if we want to remove that functionality.

test/experiment_tests/clock.jl
juliasloan25: As it is right now, `test/experiment_tests/coupler_state.jl` tests `experiments/ClimaCore/CouplerState/coupler_state.jl`, and `test/experiment_tests/clock.jl` tests `experiments/ClimaCore/CoupledSimulations/coupled_simulation.jl` (which may be confusing naming). Do you think I should still combine the test files, or should I rename the clock one?

test/Project.toml
LenkaNovak: Ok makes sense!

test/experiment_tests/clock.jl
LenkaNovak: Renaming is definitely a good idea. I would prefer containing this in as few files as possible, since it's not code we want to retain long term, but either works in practice. :)

juliasloan25: Sounds good, I agree! I condensed all of the code into `experiments/ClimaCore/CoupledSims/coupled_sim.jl`, and the tests into `test/experiment_tests/coupled_sims.jl`

PR #476, opened 2023-10-24
use ClimaLSM branch

### LINKED ISSUE #470: race condition in land albedo read
This only happens for some specifications of the MPI setup as [here](https://buildkite.com/clima/climacoupler-longruns/builds/251). Currently, we get around this using [this fix](https://github.com/CliMA/ClimaCoupler.jl/commit/ca49a4100e85f09250c21755446069733330a3ed)

[This](https://github.com/CliMA/ClimaCoupler.jl/blob/ln/longruns-nwmi3/experiments/AMIP/modular/components/land/bucket_init.jl#L191C1-L203C82) needs to be removed once land is fixed.

juliasloan25:
## Purpose 
fixes #470

## To-do
- [x] use ClimaLSM fix branch
- [ ] test on cluster

----


juliasloan25:
#502 uses the updated version of ClimaLSM containing this fix, so we don't need to use the branch of ClimaLSM anymore

PR #493, opened 2023-11-06
Run CI on each PR

Sbozzolo:
Bors is gone.


Sbozzolo:
@LenkaNovak, when this has merged, can you try merging a new PR that removes `bors.toml` to check that everything works as you would expect?

To merge, you can just click the "Enable auto-merge" button on your PR.

LenkaNovak:
I'll do that, thanks @Sbozzolo

Sbozzolo:
> I'll do that, thanks @Sbozzolo

The CI is failing and it seems that there's something wrong in `ClimaCoupler`. Have you seen this before?

LenkaNovak:
Not until this morning (I was just testing if it's just my own branch but it looks like even the `main` is affected). It was fine on Friday. I can't reproduce this locally, and checking in my local Manifest (as a test) didn't help. Any idea what's changed? @charleskawczynski @simonbyrne

LenkaNovak:
> Not until this morning (I was just testing if it's just my own branch but it looks like even the `main` is affected). It was fine on Friday. I can't reproduce this locally, and checking in my local Manifest (as a test) didn't help. Any idea what's changed? @charleskawczynski @simonbyrne

Ahh, the ClimaAtmos release is probably the culprit. I'm pinning to Atmos 0.16.2 now.

LenkaNovak:
> > Not until this morning (I was just testing if it's just my own branch but it looks like even the `main` is affected). It was fine on Friday. I can't reproduce this locally, and checking in my local Manifest (as a test) didn't help. Any idea what's changed? @charleskawczynski @simonbyrne
> 
> Ahh, the ClimaAtmos release is probably the culprit. I'm pinning to Atmos 0.16.2 now.

Hmm, still breaks!

juliasloan25:
Now that I think about it, for Clima packages we should pin to the specific patch version we've most recently updated to, since patch releases can introduce breaking changes.

Could this failure be due to not having compats for standard library packages?

LenkaNovak:
> Now that I think about it, for Clima packages we should pin to the specific patch version we've most recently updated to, since patch releases can introduce breaking changes.
> 
> Could this failure be due to not having compats for standard library packages?

Thanks for chiming in, @juliasloan25 ! This is a good point and I agree we should do this. I'm not sure which package is causing this yet though. I've tried to pin all of the major ones and it hasn't worked for me yet... But having cleared my local `.julia`, I can now reproduce this locally!

Sbozzolo:
> Now that I think about it, for Clima packages we should pin to the specific patch version we've most recently updated to, since patch releases can introduce breaking changes.

Patch releases should not introduce breaking changes. If they do, please report them (we should learn to be more mindful about this). 

There was at least one mistake on ClimaAtmos side. `0.16.3` should have been breaking because all the cache stuff has changed. 

> Could this failure be due to not having compats for standard library packages?

No, I think that only affects the registrar.

charleskawczynski:
https://github.com/CliMA/ClimaCoupler.jl/pull/493#issuecomment-1797069509 cc @sriharshakandala

sriharshakandala:
> [#493 (comment)](https://github.com/CliMA/ClimaCoupler.jl/pull/493#issuecomment-1797069509) cc @sriharshakandala

Looks like this was missed. It should have been tagged `0.17` if this was a breaking release!. We can tag `0.17` in the next day or so!

.github/workflows/Documentation.yml
```diff
@@ -2,10 +2,6 @@ name: Documentation
 
 on:
   push:
-    branches:
-      - main
```
simonbyrne, line 6: I think we still want to run this on `main`, so that the `dev` docs get built.

PR #519, opened 2023-11-22
update ClimaCore experiments

juliasloan25:
## Purpose 
- [x] get tc1 running
- [ ] clean up tc1
- [ ] rm bc-wave-slab, tc2, tc4


LenkaNovak (approved):
Thank you @juliasloan25 , looks great! Just a few minor comments. Eventually this should be our entry-level tutorial, so each line should be clear/explained, but we can always do this together during our doc week. Also, it would be good to include the conservation plot in the artifacts on Buildkite. Otherwise, it's good to go. 🚀

PR #526, opened 2023-11-28
remove OrdinaryDiffEq, DiffEqBase

juliasloan25:
## Purpose 
remove ODE and DiffEqBase dependencies, replace with SciMLBase

currently developing on top of #519 - cherry pick commit making change here after that PR merged (will reduce diff)

## Note
We have one test for eisenman sea ice that requires us to add SciMLBase and ClimaTimeSteppers to the `test/` environment. I would rather remove this test from `test/` and just rely on this sea ice stepping being tested in `experiments/AMIP`. If it isn't currently being tested in our experiments CI, we should add a test case for it.

----


experiments/AMIP/modular/components/ocean/eisenman_seaice_init.jl
```diff
@@ -226,8 +226,8 @@ function eisenman_seaice_init(
         thermo_params = thermo_params,
         dss_buffer = ClimaCore.Spaces.create_dss_buffer(ClimaCore.Fields.zeros(space)),
     )
-    problem = ODEProblem(ode_function, Y, FT.(tspan), cache)
-    integrator = init(problem, ode_algo, dt = FT(dt), saveat = FT(saveat), adaptive = false)
+    problem = ODEProblem(ode_function, Y, Float64.(tspan), cache)
+    integrator = init(problem, ode_algo, dt = Float64(dt), saveat = Float64(saveat), adaptive = false)
```
Sbozzolo, line 230: Why is this change needed to remove OrdinaryDiffEq and DiffEqBase?

Sbozzolo: (Also below)

experiments/ClimaCore/atm-ocn-lnd/atmos_simulation.jl
```diff
@@ -123,8 +122,12 @@ function atmos_simulation(
 
     # Put all prognostic variable arrays into a vector and ensure that solve can partition them
     Y_atm = ArrayPartition((T_atm_0.Yc, T_atm_0.Yf, zeros(3)))
-    prob_atm = ODEProblem(∑tendencies_atm!, Y_atm, (start_time, stop_time), (parameters, [land_surface_temperature]))
-    simulation = init(prob_atm, SSPRK33(), dt = Δt_min, saveat = 1 * Δt_min)
+
+    ode_algo = CTS.ExplicitAlgorithm(CTS.RK4())
```
Sbozzolo, line 126: Here you are changing the algorithm. Is RK4 the best one to use?

Sbozzolo: (Same below)

test/component_model_tests/eisenman_seaice_tests.jl
```diff
@@ -11,6 +11,7 @@ import ClimaCoupler: Regridder
 import CLIMAParameters as CP
 import Thermodynamics as TD
 import Thermodynamics.Parameters as TP
+using SciMLBase: step!
```
Sbozzolo, line 14: I don't think this line is needed

experiments/AMIP/modular/components/ocean/eisenman_seaice_init.jl
juliasloan25: it isn't, but I just noticed we were converting the times to `FT` and figured I would remove that to try to reduce confusion about the type of the time

experiments/ClimaCore/atm-ocn-lnd/atmos_simulation.jl
juliasloan25: I changed to RK4 because it's an `ExplicitAlgorithm` in ClimaTimeSteppers, whereas SSP333 is an `IMEXAlgorithm` ([docs](https://clima.github.io/ClimaTimeSteppers.jl/dev/api/ode_solvers/#ClimaTimeSteppers.IMEXAlgorithm)). The interface for `ExplicitAlgorithm` is simpler (doesn't require specifying Newton's method args), and we're stepping everything explicitly anyway so I thought this was a simpler change. The simulations in `experiments/ClimaCore` are pretty simple/short so the timestepper shouldn't make much of a difference

Sbozzolo: Maybe you can have the simulations output checked before merging this?

experiments/AMIP/modular/components/ocean/eisenman_seaice_init.jl
juliasloan25: the time float type just gets converted to Float64 internally in ClimaTimeSteppers, so it doesn't matter too much. I just thought it was misleading to explicitly cast it to FT when we don't expect (or want) it to have that type

experiments/ClimaCore/atm-ocn-lnd/atmos_simulation.jl
juliasloan25: This simulation doesn't even run at the moment, which I need to fix next week. The other simulations that I've changed are tested on buildkite and they run without error and the outputs (energy conservation) look good (no noticeable difference from the previous algorithm)

experiments/AMIP/modular/components/ocean/eisenman_seaice_init.jl
juliasloan25: I'm going to make the time FT changes in the PR testing Float32 compatibility more comprehensively, since I think it relates to that more. I'll keep it as a separate commit when I add it there

Sbozzolo: Sounds good! Thank you

PR #529, opened 2023-11-30
run slabplanet on gpu

### LINKED ISSUE #530: run slabplanet sim on GPU
initial step of #390 

## General approach
Run driver on GPU, fix non-compatible pieces as they come up. Document them and keep track of patterns and anti-patterns

Hopefully we'll inherit GPU compatibility from the packages we're using (ClimaCore, ClimaAtmos, ClimaLSM, etc 😄 )

closed by #529

juliasloan25:
## Purpose 
go through `coupler_driver_modular.jl`, update functions that won't run on GPU so that the slabplanet setup will run on GPU.

### Status
This currently runs on `clima`, but not on `central` due to a too-large parameter getting passed to the gpu
See https://github.com/CliMA/ClimaCore.jl/issues/1597

closes #530

## To-do
- [x] find better binary_mask solution
- [x] clean up
- [x] add GPU slabplanet case to buildkite
- [x] check AMIP paperplots change - why?
- [x] remove Ref() in atmos_init

## Content
- [x] update `mpi_init`
  - simplified, moved to driver
- [x] update atmos_init 0-setting
  - shouldn't use parent, but we can fix that later
- [x] Regridder use CPU context for file reading/writing
  - fixes `unique_nodes` error
- [x] don't pass `FT` as a argument (try to get from e.g. space instead)
  - ex. in slab_ocean_space_init, but probably more places too
- [x] `pi` is of type `Irrational(:pi)`, which isn't isbits (should be wrapped in FT)
  - in slab_ocean_space_init
- [x] `binary_mask` is not `isbits`
  - when `threshold = eps(FT)` or `0`, can replace with `clamp`, but need other solution for other thresholds
- [x] bug: anom_ampl undefined in [this line](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/components/ocean/slab_ocean_init.jl#L57) of slab_ocean_space_init, code never reaches this point because `anomaly = false` always
  - function changed to remove unused code branches

## Specific todos (later work - documented in SDI)
- [ ] look for places where `FT` passed as argument
- [ ] look for places where `pi` not wrapped in FT
- [ ] remove use of parent
- [ ] fix plotting (make sure on CPU not GPU)

## Notes
- Bucket model is not entirely GPU-compatible (esp. file reading - BulkAlbedoTemporal, BulkAlbedoStatic)
  - need to fix in ClimaLSM before coupler can be GPU-compatible
  - for now, can test with `BulkAlbedoFunction`, which is GPU-compatible

----


Sbozzolo:
` Regridder unique_nodes error - ?` reminds me of something I fixed in ClimaLSM. I don't remember the details, but it could have been in the first PR.

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -226,25 +226,36 @@ function bucket_init(
 
     # Initial conditions with no moisture
     Y, p, coords = initialize(model)
-    Y.bucket.T = map(coords.subsurface) do coord
```
juliasloan25, line 229: moved `map` inside each branch of if/else to avoid branching on GPU

experiments/AMIP/components/ocean/slab_ocean_init.jl
```diff
@@ -36,30 +36,33 @@ struct OceanSlabParameters{FT <: AbstractFloat}
 end
 name(::SlabOceanSimulation) = "SlabOceanSimulation"
 
-# init simulation
-function slab_ocean_space_init(::Type{FT}, space, p) where {FT}
+function get_T_sfc(params, coord)
```
juliasloan25, line 39: moved code into separate function to use `broadcast` instead of `map` on `coords`

experiments/AMIP/Project.toml
```diff
@@ -33,6 +34,7 @@ PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 RRTMGP = "a01a1ee8-cea4-48fc-987c-fc7878d79da1"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
Sbozzolo, line 37: Please remove `Revise` from the dependencies.

Do you need `CUDA`?

experiments/AMIP/components/ocean/eisenman_seaice_init.jl
```diff
@@ -189,7 +189,8 @@ function ∑tendencies(dY, Y, cache, _)
     @. dY.q_sfc = -Y.q_sfc / Δt
 
     # update ice area fraction (binary mask for now)
-    cache.ice_area_fraction .= Regridder.binary_mask.(Y.h_ice, threshold = eps())
+    mask_threshold = eps()
```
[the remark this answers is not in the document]
Sbozzolo: You could also define a second `binary_mask` method 
```
function binary_mask(var) 
FT = eltype(var)
 var > eps(FT) ? FT(1) : FT(0)
```
This is almost equivalent as having `eps(FT)` as default. When you call binary_mask with two arguments you call a different method that sets a different threshold.

experiments/AMIP/components/ocean/prescr_seaice_init.jl
```diff
@@ -78,8 +78,10 @@ function ice_rhs!(du, u, p, _)
     rhs = @. (-F_turb_energy - F_radiative + F_conductive) / (params.h * params.ρ * params.c)
 
     # do not count tendencies that lead to temperatures above freezing, and mask out no-ice areas
-    area_mask = Regridder.binary_mask.(area_fraction, threshold = eps(FT))
-    unphysical = @. Regridder.binary_mask.(T_freeze - (Y.T_sfc + FT(rhs) * FT(p.dt)), threshold = FT(0)) .* area_mask
+    area_threshold = eps(FT)
+    area_mask = Regridder.binary_mask.(area_fraction, area_threshold)
+    calc_threshold = FT(0)
```
Sbozzolo, line 83: Why 0 instead of eps here?

experiments/AMIP/components/ocean/slab_ocean_init.jl
```diff
@@ -36,30 +36,33 @@ struct OceanSlabParameters{FT <: AbstractFloat}
...
+        radlon = FT(coord.long / 180 * pi)
+        stdev = FT(5 / 180 * pi)
+        anom_ampl = FT(0)
+        anom = FT(anom_ampl * exp(-((radlat - lat_0)^2 / 2stdev^2 + (radlon - lon_0)^2 / 2stdev^2)))
+    elseif anomaly_tropics == true
+        anom = FT(29 * exp(-coord.lat^2 / (2 * 26^2)))
+    end
+    T_sfc = FT(T_sfc_0 + anom)
+    return T_sfc
+end
```
Sbozzolo, line 57: Lot's of values are hardcoded here, including some booleans. There seem to be some overlap in what is happening here with a previous comment of mine.

```julia
if anomaly == true
```
is equivalent to
```julia
if anomaly
```
when anomaly is a `Bool`.

But `anomaly` is hardcoded to be false. This function should be fixed to do what is supposed and only what it should without hardcoded booleans.

experiments/AMIP/components/ocean/slab_ocean_init.jl
```diff
@@ -36,30 +36,33 @@ struct OceanSlabParameters{FT <: AbstractFloat}
...
-            lon_0 = FT(-90) / FT(180) * pi
-            radlon = coord.long / FT(180) * pi
-            stdev = FT(5) / FT(180) * pi
-            anom = anom_ampl * exp(-((radlat - lat_0)^2 / 2stdev^2 + (radlon - lon_0)^2 / 2stdev^2))
-        elseif anomaly_tropics == true
-            anom = FT(29) * exp(-coord.lat^2 / (2 * 26^2))
-        end
-        T_sfc = T_sfc_0 + anom
-    end
+    T_sfc = get_T_sfc.(Ref(params), coords)
```
Sbozzolo, line 65: Instead of passing the chunky `params` struct, you can pass the only value that you need in `get_T_sfc` (T_sfc_0). Even better, you can start by setting `T_sfc .= T_sfc_0` and then `.+=` a function that computes the anomaly.

Instead of `get_T_sfc`, you'd have `anomaly(coords)`

experiments/AMIP/coupler_driver.jl
```diff
@@ -65,6 +68,7 @@ using Plots
 using Statistics: mean
 import ClimaAtmos as CA
 import YAML
+import CUDA
```
Sbozzolo, line 71: Is CUDA needed?

experiments/AMIP/coupler_driver.jl
```diff
@@ -713,8 +717,8 @@ if ClimaComms.iamroot(comms_ctx)
         )
     end
 
-    ## sample animations
-    if !is_distributed && config_dict["anim"]
+    ## sample animations (not compatible with MPI)
+    if !CA.is_distributed(ClimaComms.context(boundary_space)) && config_dict["anim"]
```
Sbozzolo, line 721: Don't use have a `comms_ctx` variable already (instead of fetching it from the boundary space)?

perf/Project.toml
```diff
@@ -3,6 +3,7 @@ ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
 ArtifactWrappers = "a14bc488-3040-4b00-9dc1-f6467924858a"
 AtmosphericProfilesLibrary = "86bc3604-9858-485a-bdbe-831ec50de11d"
 CLIMAParameters = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
Sbozzolo, line 6: Is CUDA needed?

src/FieldExchanger.jl
```diff
@@ -112,41 +112,33 @@ function update_sim!(atmos_sim::Interfacer.AtmosModelSimulation, csf, turbulent_
...
-function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, turbulent_fluxes, area_fraction = nothing)
+function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, turbulent_fluxes, area_fraction)
     FT = eltype(area_fraction)
 
     # atmospheric surface density
     Interfacer.update_field!(sim, Val(:air_density), csf.ρ_sfc)
 
     # turbulent fluxes
+    mask_threshold = eps(FT)
+    mask = FT.(Regridder.binary_mask.(area_fraction, mask_threshold))
```
Sbozzolo, line 131: I don't think `FT.` is needed here

experiments/AMIP/components/ocean/prescr_seaice_init.jl
LenkaNovak: Because for the `area_mask`, the `area_fraction` we're evaluating could be very slightly positive and still regarded to be ice free, hence `eps(FT)`. For `unphysical`, this is a crude guess as to whether the `Y.T_sfc` will reach below the freezing point or not, so FT(0) is more accurate.

LenkaNovak (commented):
After addressing Gabriele's comments, I only have a few minor ones, thank you, @juliasloan25. I'm also noting a small difference in the AMIP paperplots from the last merge. It doesn't look wrong, it's just slightly different. Would we not expect no behavioral change? 

Do we want to open a new issue on adding the albedo read from file to the GPU AMIP runs, once that's sorted in ClimaLSM?

experiments/AMIP/components/ocean/slab_ocean_init.jl
```diff
@@ -72,7 +75,8 @@ function slab_ocean_rhs!(dY, Y, cache, t)
     p, F_turb_energy, F_radiative, area_fraction = cache
     FT = eltype(Y.T_sfc)
     rhs = @. -(F_turb_energy + F_radiative) / (p.h * p.ρ * p.c)
-    parent(dY.T_sfc) .= parent(rhs .* Regridder.binary_mask.(area_fraction, threshold = eps(FT))) * p.evolving_switch
+    mask_threshold = eps(FT)
+    parent(dY.T_sfc) .= parent(rhs .* Regridder.binary_mask.(area_fraction, mask_threshold)) * p.evolving_switch
```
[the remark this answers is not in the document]
Sbozzolo: I think you should be able to use `Field`s directly.

experiments/AMIP/components/ocean/slab_ocean_init.jl
juliasloan25: @LenkaNovak is it okay to remove the `anomaly == true` branch?

experiments/AMIP/components/ocean/slab_ocean_init.jl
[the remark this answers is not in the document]
juliasloan25: I removed `parent` here, but we should also have a separate PR that goes through all the code and removes use of parent. I think we should first prioritize getting the GPU runs working, then we can look at code quality improvements like this

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -226,25 +226,36 @@ function bucket_init(
...
             ΔT = FT(40 * cos(radlat)^4)
-        else
+            T_sfc_0 + ΔT
+        end
+    else
+        Y.bucket.T = map(coords.subsurface) do coord
             ΔT = FT(0)
+            T_sfc_0 + ΔT
         end
-        T_sfc_0 + ΔT
```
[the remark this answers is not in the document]
juliasloan25: @LenkaNovak I'm not sure where the hardcoded values come from, do you have any input?

juliasloan25:
I've addressed all the review feedback. The MacOS github actions are failing due to some issue with Plots/GR that I'm trying to figure out (see [Plots.jl issue](https://github.com/JuliaPlots/Plots.jl/issues/4879)). But other than that, everything looks good (even the GPU slabplanet run is doing fine on the cluster).

@LenkaNovak I wanted to check that these most recent changes I made are okay with you, especially with the changes to the temperature anomaly function for the slab ocean. I removed all the code branches that we weren't using, with the thinking that we could always go back into our repo history to find them if we want to add them back. I was also wondering if you could provide any comments for the hardcoded values used for both the bucket and slab ocean temperature anomalies. Thanks in advance :)

Project.toml
```diff
@@ -35,7 +35,7 @@ DocStringExtensions = "0.8, 0.9"
 Insolation = "0.8"
 JLD2 = "0.4"
 NCDatasets = "0.11, 0.12, 0.13"
-Plots = "1.39.0"
+Plots = "< 1.40.0"
```
Sbozzolo, line 38: Ideally let's not merge this in. 

I'd rather temporarily disable CI on MacOS than pinning a package.

experiments/AMIP/components/atmosphere/climaatmos_init.jl
```diff
@@ -309,9 +308,7 @@ function get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:F_radiative_TOA})
     # save radiation source
     if radiation != nothing
         face_space = axes(atmos_sim.integrator.u.f)
-        z = parent(Fields.coordinate_field(face_space).z)
-        Δz_top = round(FT(0.5) * (z[end, 1, 1, 1, 1] - z[end - 1, 1, 1, 1, 1]))
-        n_faces = length(z[:, 1, 1, 1, 1])
+        n_faces = length(face_space.grid.vertical_grid.topology.mesh.faces)
```
Sbozzolo, line 311: I think that there are better accessors. You might want to use `Spaces.vertical_topology(face_space)`

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -159,6 +159,46 @@ function ClimaLSM.initialize_drivers(a::CoupledRadiativeFluxes{FT}, coords) wher
...
+    radlat = coord.lat / FT(180) * pi
+    anom_ampl = FT(0) # this is zero, no anomaly
+    lat_0 = FT(60) / FT(180) * pi
+    lon_0 = FT(-90) / FT(180) * pi
+    radlon = coord.long / FT(180) * pi
+    stdev = FT(5) / FT(180) * pi
+
+    anom = anom_ampl * exp(-((radlat - lat_0)^2 / 2stdev^2 + (radlon - lon_0)^2 / 2stdev^2))
+    return anom
+end
```
Sbozzolo, line 179: Please, document these functions: where they come from, what they mean, and so on.

You also don't need `FT`s when your expression involves integers and `coord`. 

Also, it seems that you are converting from degrees to radians. There's a much more readable built-in function for that: `deg2rad`.

Finally, this expression is identically 0 all the time because `anom_ampl` is 0. I'd recommend understanding what is this function trying to accomplish and checking that its behavior is the intended one.

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -159,6 +159,46 @@ function ClimaLSM.initialize_drivers(a::CoupledRadiativeFluxes{FT}, coords) wher
...
+
+"""
+    temp_anomaly_aquaplanet(coord)
+
+Introduce a temperature IC anomaly for the aquaplanet case.
+"""
+function temp_anomaly_aquaplanet(coord)
+    FT = eltype(coord)
+    return FT(29) * exp(-coord.lat^2 / (2 * 26^2))
+end
```
Sbozzolo, line 189: Document this function, especially because it has some numbers that are not intuitive like 26 and 29. You don't need `FT` here.

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -159,6 +159,46 @@ function ClimaLSM.initialize_drivers(a::CoupledRadiativeFluxes{FT}, coords) wher
...
+"""
+    temp_anomaly_amip(coord)
+
+Introduce a temperature IC anomaly for the AMIP case.
+"""
+function temp_anomaly_amip(coord)
+    FT = eltype(coord)
+    radlat = coord.lat / FT(180) * pi
+    return FT(40 * cos(radlat)^4)
+end
```
Sbozzolo, line 200: Same as before with regards to documentation, radians, and FTs.

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -226,26 +266,19 @@ function bucket_init(
...
-        T_sfc_0 + ΔT
-    end
+
+    # Get temperature anomaly function
+    T_functions = Dict(
+        "aquaplanet" => temp_anomaly_aquaplanet,
+        "amip" => temp_anomaly_amip,
+        "zonally_asymmetric" => temp_anomaly_zonally_asymmetric,
+    )
+    haskey(T_functions, land_temperature_anomaly) || error("land temp anomaly function not supported")
```
Sbozzolo, line 276: It would be useful for your error message to contain the value that was passed, `land_temperature_anomaly` here.
```suggestion
haskey(T_functions, land_temperature_anomaly) || error("land temp anomaly function $land_temperature_anomaly not supported")
```

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -226,26 +266,19 @@ function bucket_init(
...
+    temp_anomaly = T_functions[land_temperature_anomaly]
+
+    # Set temperature IC including anomaly
+    T_sfc_0 = FT(271.0)
```
Sbozzolo, line 280: Where does this (and the other) numbers come from?

Sbozzolo: Also, I think you don't need `FT` here.

experiments/AMIP/components/ocean/prescr_seaice_init.jl
```diff
@@ -78,8 +78,9 @@ function ice_rhs!(du, u, p, _)
     rhs = @. (-F_turb_energy - F_radiative + F_conductive) / (params.h * params.ρ * params.c)
 
     # do not count tendencies that lead to temperatures above freezing, and mask out no-ice areas
-    area_mask = Regridder.binary_mask.(area_fraction, threshold = eps(FT))
-    unphysical = @. Regridder.binary_mask.(T_freeze - (Y.T_sfc + FT(rhs) * FT(p.dt)), threshold = FT(0)) .* area_mask
+    area_mask = Regridder.binary_mask.(area_fraction)
+    threshold = FT(0)
```
Sbozzolo, line 82: It doesn't really matter, but for your information
```julia
julia> @btime zero(FT)
  63.319 ns (0 allocations: 0 bytes)
```
versus
```julia
julia> @btime FT(0)
  73.974 ns (1 allocation: 16 bytes)
```

experiments/AMIP/components/ocean/slab_ocean_init.jl
```diff
@@ -72,8 +73,7 @@ function slab_ocean_rhs!(dY, Y, cache, t)
     p, F_turb_energy, F_radiative, area_fraction = cache
     FT = eltype(Y.T_sfc)
     rhs = @. -(F_turb_energy + F_radiative) / (p.h * p.ρ * p.c)
-    parent(dY.T_sfc) .= parent(rhs .* Regridder.binary_mask.(area_fraction, threshold = eps(FT))) * p.evolving_switch
-
+    @. dY.T_sfc .= rhs .* Regridder.binary_mask.(area_fraction) * p.evolving_switch
```
Sbozzolo, line 76:
```suggestion
    @. dY.T_sfc = rhs * Regridder.binary_mask(area_fraction) * p.evolving_switch
```

src/ConservationChecker.jl
```diff
@@ -217,6 +217,7 @@ end
 
 # setup the GKS socket application environment as nul for better performance and to avoid GKS connection errors while plotting
 ENV["GKSwstype"] = "nul"
+ENV["JULIA_GR_PROVIDER"] = "GR"
```
Sbozzolo, line 220: If this didn't help, you can remove it.

experiments/AMIP/components/land/bucket_init.jl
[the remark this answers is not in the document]
szy21: Zonal symmetric aquaplanet SST is from the moist held-suarez paper eq. 6: https://gmd.copernicus.org/articles/9/1263/2016/gmd-9-1263-2016.pdf

experiments/AMIP/components/atmosphere/climaatmos_init.jl
LenkaNovak: Could we also rename this to `nz_faces`?

experiments/AMIP/components/land/bucket_init.jl
LenkaNovak: We can remove this case - it's too specific to idealized experiments, which we aren't planning to do using this release.

experiments/AMIP/components/land/bucket_init.jl
LenkaNovak: This follows the atmos setup [here](https://github.com/CliMA/ClimaAtmos.jl/blob/85467c4d3013c3495e925f8952971012f67fb6ad/src/initial_conditions/atmos_state.jl#L73).

experiments/AMIP/components/land/bucket_init.jl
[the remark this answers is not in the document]
LenkaNovak: We should document this in Atmos as well. :)

src/FieldExchanger.jl
```diff
@@ -112,41 +112,32 @@ function update_sim!(atmos_sim::Interfacer.AtmosModelSimulation, csf, turbulent_
...
 """
-function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, turbulent_fluxes, area_fraction = nothing)
+function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, turbulent_fluxes, area_fraction)
     FT = eltype(area_fraction)
 
     # atmospheric surface density
     Interfacer.update_field!(sim, Val(:air_density), csf.ρ_sfc)
 
     # turbulent fluxes
+    mask = Regridder.binary_mask.(area_fraction)
```
LenkaNovak, line 130: This probably allocates as it is called every dt. I wonder if we should have a field in cache which holds temporary calculations on the boundary space, like Atmos does. @Sbozzolo , what do you think? There will be many instances of this though, so I'd suggest leaving this for separate PR.

LenkaNovak (approved):
LGTM, @juliasloan25. Following @Sbozzolo 's comments, I only have minor comments. We should also open an issue on reducing the allocations with temporaty fields, if we all agree this would help. Thank you!

src/FieldExchanger.jl
Sbozzolo: Yes, good point (for a future PR)!

experiments/AMIP/components/land/bucket_init.jl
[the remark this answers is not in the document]
juliasloan25: I've added some comments explaining the aquaplanet and AMIP cases, and removed the zonally asymmetric case. Let me know if it's not satisfactory

juliasloan25:
> Do we want to open a new issue on adding the albedo read from file to the GPU AMIP runs, once that's sorted in ClimaLSM?

Just opened an issue for it https://github.com/CliMA/ClimaCoupler.jl/issues/583

src/FieldExchanger.jl
juliasloan25: I opened an issue here https://github.com/CliMA/ClimaCoupler.jl/issues/584. I can change this back to calling `binary_mask` inside the `update_field!` calls. I thought this way was more readable, but I agree we should avoid extra allocations.

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -159,6 +159,29 @@ function ClimaLSM.initialize_drivers(a::CoupledRadiativeFluxes{FT}, coords) wher
     return (;)
 end
 
+"""
+    temp_anomaly_aquaplanet(coord)
+
+Introduce a temperature IC anomaly for the aquaplanet case.
+The values for this case follow the initial conditions of the ClimaAtmos
```
LenkaNovak, line 166: After Zhaoyi's confirmation, I'd say "follow the moist Help-Suarez setup of Thatcher & Jabplonowski (2016, eq. 6), consistent with ClimaAtmos aquaplanet." Otherwise LGTM.

LenkaNovak (approved):

experiments/AMIP/components/land/bucket_init.jl
```diff
@@ -159,6 +159,29 @@ function ClimaLSM.initialize_drivers(a::CoupledRadiativeFluxes{FT}, coords) wher
...
+"""
+    temp_anomaly_amip(coord)
+
+Introduce a temperature IC anomaly for the AMIP case.
+The values used in this case have been tuned to align with observed temperature
+and result in stable simulations.
+"""
+function temp_anomaly_amip(coord)
+    radlat = deg2rad(coord.lat)
+    return 40 * cos(radlat)^4
```
szy21, line 182:
```suggestion
    return 40 * cosd(coord.lat)^4
```

experiments/AMIP/components/atmosphere/climaatmos_init.jl
```diff
@@ -77,9 +77,8 @@ function atmos_init(::Type{FT}, atmos_config_dict::Dict) where {FT}
     # set initial fluxes to zero
     @. integrator.p.precomputed.sfc_conditions.ρ_flux_h_tot = Geometry.Covariant3Vector(FT(0.0))
     @. integrator.p.precomputed.sfc_conditions.ρ_flux_q_tot = Geometry.Covariant3Vector(FT(0.0))
-    @. integrator.p.precomputed.sfc_conditions.ρ_flux_uₕ.components =
-        zeros(axes(integrator.p.precomputed.sfc_conditions.ρ_flux_uₕ.components))
-    parent(integrator.p.radiation.ᶠradiation_flux) .= parent(zeros(axes(integrator.p.radiation.ᶠradiation_flux)))
+    @. integrator.p.radiation.ᶠradiation_flux = Geometry.WVector(FT(0))
+    integrator.p.precomputed.sfc_conditions.ρ_flux_uₕ.components .= Ref(SMatrix{1, 2}([FT(0), FT(0)]))
```
juliasloan25, line 81: without `Ref`, this fails with `LoadError: MethodError: no method matching static_check_broadcast_shape(::ClimaCore.Spaces.SpectralElementSpace2D{…}, ::Tuple{…})`

Sbozzolo: Okay, maybe revert this change and we can think about it in the future.

juliasloan25: I'm going to keep it as `Ref(SMatrix...)` because what we had before wasn't GPU compatible. But I agree we should change it in the future

Sbozzolo: I am not sure that `Ref(SMatrix)` is correct. What we need is multiple separate instances of `SMatrix(zero)`, not multiple views over the same one.

Could you please check that using `Ref(SMatrix...)` does that?

juliasloan25: After using `Ref(SMatrix...)` to set the field, I get:
```
(parent(field))[:, :, :, 1] === (parent(field))[:, :, :, 2] = false
(parent(field))[:, :, :, 1] == (parent(field))[:, :, :, 2] = true
```
so it looks like they're separate instances

PR #534, opened 2023-12-02
test Float32 compatibility

### LINKED ISSUE #449: enable Float32 compatibility
Float32 compatibility is currently blocked by land model time type constraint [here](https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/utils.jl#L88). We need to remove this type constraint from ClimaLSM, which will hopefully allow us to run simulations with Float32

see: https://github.com/CliMA/ClimaCoupler.jl/issues/396, https://github.com/CliMA/ClimaLSM.jl/issues/327

juliasloan25:
## Purpose 
closes #449

## To Do
- [x] make ClimaLSM Float32 compatible
- [x] extend tests to check Float32 and Float64
  - all component models, and unit tests for which it makes sense (FT is used or computation is done)
- [x] add Float32 runs (AMIP and slabplanet) to buildkite


src/FieldExchanger.jl
```diff
@@ -122,32 +122,31 @@ Updates the surface component model cache with the current coupler fields of F_t
 """
 function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, turbulent_fluxes, area_fraction = nothing)
 
-    FT = eltype(area_fraction)
```
juliasloan25, line 125: `area_fraction` gave Float64 even when run with Float32 in tests

test/conservation_checker_tests.jl
```diff
@@ -26,173 +25,177 @@ struct TestAtmos{I} <: Interfacer.AtmosModelSimulation
...
-    @test isfile("test_cons_plots/energy.png")
-    @test isfile("test_cons_plots/energy_log.png")
-    @test isfile("test_cons_plots/water.png")
-    @test isfile("test_cons_plots/water_log.png")
-
-    # remove output
-    rm(output_plots; recursive = true)
 
+for FT in (Float32, Float64)
+    global get_field(s::TestAtmos, ::Val{:F_radiative_TOA}) = ones(s.i.space) .* FT.(200)
```
juliasloan25, line 42: I don't love that these are `global` but it was the only way to make them visible within the `testset`s, and at least this is just in the tests

.buildkite/pipeline.yml
```diff
@@ -239,6 +246,12 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: "Moist earth with slab surface - Float32: monin gray no_sponge idealinsol freq_dt_cpl"
```
LenkaNovak, line 249: do we also want to add an AMIP case for FT32?

test/conservation_checker_tests.jl
```diff
@@ -26,173 +25,177 @@ struct TestAtmos{I} <: Interfacer.AtmosModelSimulation
     i::I
 end
 name(s::TestAtmos) = "TestAtmos"
-get_field(s::TestAtmos, ::Val{:F_radiative_TOA}) = ones(s.i.space) .* 200
```
juliasloan25, line 29: make these multiline functions that get FT from `s.i.space`, instead of using FT from global scope

test/diagnostics_tests.jl
```diff
@@ -20,124 +20,125 @@ import ClimaCoupler.Diagnostics:
...
+            pre_save(cs.diagnostics[1].operations.accumulate, cs, cs.diagnostics[1])
+            @test cs.diagnostics[1].field_vector[1] == expected_results[c_i][2]
 
-        post_save(cs.diagnostics[1].operations.accumulate, cs, cs.diagnostics[1])
-        @test cs.diagnostics[1].field_vector[1] == expected_results[c_i][3]
+            post_save(cs.diagnostics[1].operations.accumulate, cs, cs.diagnostics[1])
+            @test cs.diagnostics[1].field_vector[1] == expected_results[c_i][3]
+        end
     end
 end
```
LenkaNovak, line 144: It would be good to use FT locally (discussed offline)

test/flux_calculator_tests copy 2.jl
```diff
@@ -0,0 +1,415 @@
+
```
LenkaNovak, line 1: rm file?

test/flux_calculator_tests copy.jl
```diff
@@ -0,0 +1,415 @@
+
```
LenkaNovak, line 1: rm file?

test/interfacer_tests.jl
```diff
@@ -15,19 +15,18 @@ import ClimaCoupler.Interfacer:
...
-
 for FT in (Float32, Float64)
+    boundary_space = TestHelper.create_space(FT)
+    global get_field(::SurfaceModelSimulation, ::Val{:var}) = ones(boundary_space)
+    global get_field(::SurfaceModelSimulation, ::Val{:var_float}) = FT(2)
+    global get_field(::SurfaceModelSimulation, ::Val{:surface_temperature}) = ones(boundary_space) .* FT(300)
+
     @testset "test CoupledSim construction, float_type for FT=$FT" begin
         cs = CoupledSimulation{FT}(
             nothing, # comms_ctx
```
LenkaNovak, line 32: local FT again?

LenkaNovak (approved):
LGTM, a few minor comments. Thank you!

test/flux_calculator_tests copy 2.jl
juliasloan25: oops!

src/FieldExchanger.jl
juliasloan25: I was able to fix this so we can get FT from `area_fraction` again

LenkaNovak (approved):
Revs LGTM. There was one more non-local FT call, I think, otherwise feel free to merge. Thank you!

PR #543, opened 2023-12-20
debug MPI AMIP

juliasloan25:
## Purpose 
debug MPI AMIP hanging

see #538 for more details

Issue #538, opened 2023-12-07
MPI AMIP occasional hang


juliasloan25:
Our MPI AMIP run is hanging almost every time it's run.

It looks like the error originated in the PR where we updated dependencies https://github.com/CliMA/ClimaCoupler.jl/pull/535, which updated from ClimaCore v0.10 to v0.11 and also removed a pin for HDF5 at 1.12. That branch had the first hangs, though some commits hung and others passed. This branch was merged on 12/6.

Hangs on [this line](https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/Regridder.jl#L335), where `ClimaCore.InputOutput.read_field` is called.

Edit: We've discovered that the issue is coming from memoizing Grids using a WeakValueDict, which was introduced in ClimaCore v0.11.0 See ClimaCore issue [#1589](https://github.com/CliMA/ClimaCore.jl/issues/1589) for more details. A fix will need to be implemented in ClimaCore.

### Examples of failure on buildkite
- [debug branch 2298 (has useful print statements)](https://buildkite.com/clima/climacoupler-ci/builds/2298#018cd097-0dbe-4d22-bc9d-dc1333139e2c)
- [dep updates PR 12/5 2208](https://buildkite.com/clima/climacoupler-ci/builds/2208)
- [dep updates PR 12/6 2223](https://buildkite.com/clima/climacoupler-ci/builds/2223#018c40b2-d596-4bbf-ae4e-c57216a67182)
- [dep updates PR 12/6 2220](https://buildkite.com/clima/climacoupler-ci/builds/2220)
- [on main 12/7 2229](https://buildkite.com/clima/climacoupler-ci/builds/2229#018c41cb-fba6-4b5b-b7ef-1660dd6e7aaa)
- [depot branch 12/8 2237](https://buildkite.com/clima/climacoupler-ci/builds/2237#_)
- this run [passed](https://buildkite.com/clima/climacoupler-ci/builds/2281), then [failed](https://buildkite.com/clima/climacoupler-ci/builds/2299) when re-run without any changes to the code.
- same error in MPI BCReader and MPI Regridder tests, [debug branch 1/4 2311](https://buildkite.com/clima/climacoupler-ci/builds/2311#_)

### Debugging notes
- increased `slurm_mem` for MPI AMIP job from 16GB to 32GB
[...]

PR #555, opened 2024-01-08
MPI debugging rebased on main

juliasloan25:
## Purpose 
continuation from #543 but rebased on current main (notably including new depot)

issue #538


juliasloan25:
won't merge - fix in #556

PR #561, opened 2024-01-18
use albedo function in config

### LINKED ISSUE #560: use function option for function albedo run
We have a slabplanet run `slabplanet_albedo_function` that is meant to test the bucket albedo prescribed function option. However we're using the wrong option in the config file ("static_map" should be "function" in [this line](https://github.com/CliMA/ClimaCoupler.jl/blob/main/config/model_configs/slabplanet_albedo_function.yml#L13))

juliasloan25:
## Purpose 
use albedo function for run meant to test this option

closes #560


LenkaNovak (approved):
LGTM, Julia. Feel free to merge.

PR #564, opened 2024-01-19
rename dt_save_to_disk to dt_save_state_to_disk

szy21:
## Purpose 

Following recent changes in ClimaAtmos.

## Content

----


juliasloan25 (commented):
I think there are two other files that need to be updated: `config/longrun_configs/amip_target_topo.yml` and `amip_target.yml`

szy21:
> I think there are two other files that need to be updated: `config/longrun_configs/amip_target_topo.yml` and `amip_target.yml`

Thanks! I have fixed them.

juliasloan25 (approved):

LenkaNovak (approved):
LGTM. In case you're using it, the parsed args in `local_checks.sh` do not take precedence over the config files, so the arguments here may need to be refactored into the relevant yml config files.

szy21:
> LGTM. In case you're using it, the parsed args in `local_checks.sh` do not take precedence over the config files, so the arguments here may need to be refactored into the relevant yml config files.

Good to know, thanks!

PR #565, opened 2024-01-20
truncate stacktraces for non-interactive runs

juliasloan25:
## Purpose 
When Julia 1.10+ is used interactively, stacktraces contain reduced type information to make them shorter. This PR does the same for non-interactive runs
As was done in ClimaAtmos [#2549](https://github.com/CliMA/ClimaAtmos.jl/pull/2549)


Sbozzolo (approved):
I just realized that this might not work on julia 1.9 (but I don't think that's a problem)

Issue #568, opened 2024-01-20
Small fixes for conservation


LenkaNovak:
## Purpose 

Following the ClimaAtmos and ClimaLSM updates, we need to update the following to retain conservation. 

## To-do

- [x] retain output diagnostic files at the end of the run by default
- [x] use rho_ice where snow (fixes non conservation in main)
- [x] add energy from precipitation approximation - requires (`ᶜS_ρq_tot` to be initialized with 0)
- [x] add asserts to conservation (1e-3 for now) - this will catch any future departures from conservation immediately

# QA
- [x] ensure stable CI
- [x] ensure stable longruns
  - diffusion only still needs stabilizing - following params of [this config](https://github.com/CliMA/ClimaCoupler.jl/blob/7b860c23161b34810cdb41d86459ae44ab6fc423/config/longrun_configs/amip_target_topo.yml)  and [this build](https://buildkite.com/clima/climacoupler-longruns/builds/402)
    - Frierson diffusion breaks for for AMIP and AMIP topo [build](https://buildkite.com/clima/climacoupler-longruns/builds/435#018d6c23-6fef-467a-8122-5d3329805bf2) vs using `vert_diff: true` ([build](https://buildkite.com/clima/climacoupler-longruns/builds/436))
      - we will bring this back once issues with surface properties and atmospheric optical are resolved (O1.5.4)

PR #556, opened 2024-01-12
dependency updates

juliasloan25:
## Purpose 
update to ClimaAtmos v0.19.0, ClimaLSM v0.5.0. also updates to julia v1.10

wait to merge until #538 is fixed (see https://github.com/CliMA/ClimaCore.jl/issues/1589)

closes #554 (dependency updates)
closes #547 (julia v1.10 update)
closes #538 (MPI hang)
supersedes #549 
supersedes #550 
supersedes #558 

## To-do
- [x] ClimaLSM breaking changes
- [x] ClimaAtmos breaking changes (none)
- [x] change branch protection rule for main: ci 1.9 -> ci 1.10

PR #570, opened 2024-01-22
dep updates (Atmos v0.19.2, LSM v0.6.0)

juliasloan25:
## Purpose 
update to ClimaAtmos v0.19.2, ClimaLSM v0.6.0

replaces #566 

## To-do
- [x] update dependencies
- [x] change bucket precip sign for new convention in https://github.com/CliMA/ClimaLSM.jl/pull/445:
  - > Conform to agreed upon convention that fluxes are positive if towards atmos and negative if towards ground.
Therefore, precipitation should be negative and snow melt should be negative (corresponds to an increase in the energy of snow, or a negative heat flux).


LenkaNovak (commented):
LGTM, but it looks like we're getting slabplanet non-conservation (e.g. see default slabplanet in Buildkite). This is persisting since #556. Just trying to figure out what's changed.

LenkaNovak:
Running the `interactive_debug` seems to be conservative, which points to the current Buildkite tests being unstable (but not blowing up yet). I can fix this as part of https://github.com/CliMA/ClimaCoupler.jl/issues/568

LenkaNovak (approved):

PR #575, opened 2024-01-25
add amip with diagnostic edmf job

### LINKED ISSUE #574: Stable prescribed CO2, topo, sea ice and SST + diagnostic EDMF simulation
A demonstration of this run in the Buildkite longruns pipeline.

szy21:
## Purpose 

Adds an AMIP with edmf job to the longrun. Instead of only adding diagnostic edmf to the current `amip_target_topo` job, I chose to add the most realistic configuration (time-varying insolation and all-sky radiation). Because with edmf it is more than a factor of 2 slower I set t_end to 40days. This one should run, I can check once buildkite works again.

Also changes dt and dt_cpl in `amip_targe_topo` to be the same.

Closes #574 

## Content

----


szy21:
I will revert the pipeline change after testing.

szy21:
Ah, we need a new ClimaAtmos release for this to work, as it is using a new atmos longrun config file. I can rebase the PR after we release ClimaAtmos and update ClimaCoupler this week.

LenkaNovak:
@szy21, I know you're waiting for the update for CA, but FYI the macOS problem should be sorted once you rebase. Also, would it be possible to set up this run (but with a prescribed surface) in the Atmos Buildkite CPU longruns, so we can keep track of performance differences?

.buildkite/longruns/pipeline.yml
```diff
@@ -51,152 +51,152 @@ steps:
...
-        command: "mpiexec julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/slabplanet_target_evolve_ocn.yml"
-        artifact_paths: "experiments/AMIP/output/slabplanet/slabplanet_target_evolve_ocn_artifacts/*"
-        env:
-          CLIMACORE_DISTRIBUTED: "MPI"
-          BUILD_HISTORY_HANDLE: ""
-        agents:
-          slurm_ntasks_per_node: 16
-          slurm_nodes: 4
-          slurm_mem_per_cpu: 16G
+  # - group: "Coupler integration and conservation tests"
```
LenkaNovak, line 54: These need to be uncommented otherwise good to go.

LenkaNovak (approved):

szy21:
> @szy21, I know you're waiting for the update for CA, but FYI the macOS problem should be sorted once you rebase. Also, would it be possible to set up this run (but with a prescribed surface) in the Atmos Buildkite CPU longruns, so we can keep track of performance differences?

I would prefer to keep the most important jobs in the GPU longrun, as that is what we are focusing on right now. The corresponding one in atmos is different nevertheless, as I am using idealized insolation to be more consistent with the time-invariant boundary condition. I can add a job in the CPU longrun with the exactly same setup as the coupler one for a few weeks or months for comparison.

LenkaNovak:
> > @szy21, I know you're waiting for the update for CA, but FYI the macOS problem should be sorted once you rebase. Also, would it be possible to set up this run (but with a prescribed surface) in the Atmos Buildkite CPU longruns, so we can keep track of performance differences?
> 
> I would prefer to keep the most important jobs in the GPU longrun, as that is what we are focusing on right now. The corresponding one in atmos is different nevertheless, as I am using idealized insolation to be more consistent with the time-invariant boundary condition. I can add a job in the CPU longrun with the exactly same setup as the coupler one for a few weeks or months for comparison.

That makes sense. And yeah, that would be awesome, thank you! Or I can also do it, if you have a lot on this week. :)

szy21:
@LenkaNovak https://github.com/CliMA/ClimaAtmos.jl/pull/2624 added the job if you want to take a look

.buildkite/longruns/pipeline.yml
```diff
@@ -226,6 +226,18 @@ steps:
...
+        key: "amip_target_topo_diagedmf"
+        command: "mpiexec julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/amip_target_topo_diagedmf.yml"
+        artifact_paths: "experiments/AMIP/output/amip/amip_target_topo_diagedmf_artifacts/*"
+        env:
+          CLIMACORE_DISTRIBUTED: "MPI"
+          BUILD_HISTORY_HANDLE: ""
+        agents:
+          slurm_ntasks_per_node: 16
+          slurm_nodes: 4
+          slurm_mem_per_cpu: 20G
```
szy21, line 240: I use 20G as it seems from my experience with edmf it needs some more memory. We can reduce it in the future.

PR #579, opened 2024-01-29
Fix initial file read

### LINKED ISSUE #580: Fix BCReader
In the case that `date0 == file` date, the BCReader throws an error (e.g., [here](https://buildkite.com/clima/climacoupler-longruns/builds/402#018d47cc-6bce-48d8-99d5-0efc0c84822a/978-1197)). 

# Components
- [x] include this case in BCReader.jl]
- [x] Split initialization from the rest of the cases
- [x] clarify and extend the associated unit tests accordingly.

LenkaNovak:
closes #580

----

### DIFF
--- a/src/BCReader.jl
+++ b/src/BCReader.jl
@@ -95,7 +95,7 @@ and returns the info packaged in a single struct.
 - `boundary_space`: [Spaces.AbstractSpace] the space to which we are mapping.
 - `comms_ctx`: [ClimaComms.AbstractCommsContext] context used for this operation.
 - `interpolate_daily`: [Bool] switch to trigger daily interpolation.
-- `segment_idx0`: [Vector{Int}] index of the file data that is closest to date0.
+- `segment_idx0`: [Vector{Int}] reference date which, after initialization, refers to the the first file date index used minus 1 (segment_idx[1] - 1)
 - `scaling function`: [Function] scales, offsets or transforms `varname`.
 - `land_fraction`: [Fields.field] fraction with 1 = land, 0 = ocean / sea-ice.
 - `date0`: [Dates.DateTime] start date of the file data.
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
 function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
     # monthly count
     (; bcfile_dir, comms_ctx, hd_outfile_root, varname, all_dates, scaling_function) = bcf_info
-    midmonth_idx = bcf_info.segment_idx[1]
-    midmonth_idx0 = bcf_info.segment_idx0[1]
+    segment_idx = bcf_info.segment_idx[1] # index of the current date in the file. [segment_idx, segment_idx+1] indexes the current segment between which we interpolate
+    segment_idx0 = bcf_info.segment_idx0[1] # reference index (first segment_idx - 1)
 
-    if (midmonth_idx == midmonth_idx0) && (Dates.days(date - all_dates[midmonth_idx]) < 0) # for init
-        midmonth_idx = bcf_info.segment_idx[1] -= Int(1)
-        midmonth_idx = midmonth_idx < Int(1) ? midmonth_idx + Int(1) : midmonth_idx
-        @warn "this time period is before BC data - using file from $(all_dates[midmonth_idx0])"
-        bcf_info.monthly_fields[1] .= scaling_function(
-            Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[Int(midmonth_idx0)], varname, comms_ctx),
-            bcf_info,
-        )
-        bcf_info.monthly_fields[2] .= deepcopy(bcf_info.monthly_fields[1])
-        bcf_info.segment_length .= Int(0)
+    # upon initialization (segment_idx == segment_idx0) with model date before the final file date
+    if (segment_idx == segment_idx0) && !((date - all_dates[end]).value >= 0)
+        #  case 1: model date is before the first segment read from file
+        if (date - all_dates[segment_idx0]).value < 0
+            @warn "This time period is before BC data - using file from $(all_dates[segment_idx0])"
+            bcf_info.monthly_fields[1] .= scaling_function(
+                Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[Int(segment_idx0)], varname, comms_ctx),
+                bcf_info,
+            )
+            bcf_info.monthly_fields[2] .= deepcopy(bcf_info.monthly_fields[1])
+            bcf_info.segment_length .= Int(0)
+            bcf_info.segment_idx[1] -= Int(1)
+            bcf_info.segment_idx0[1] -= Int(2)
 
-    elseif Dates.days(date - all_dates[end - 1]) > 0 # for fini
-        @warn "this time period is after BC data - using file from $(all_dates[end - 1])"
+            # case 2: model date is after the first segment read from file
+        elseif (date - all_dates[Int(segment_idx0) + 1]).value >= 0
+            nearest_idx = argmin(
+                abs.(
+                    parse(FT, TimeManager.datetime_to_strdate(date)) .-
+                    parse.(FT, TimeManager.datetime_to_strdate.(all_dates[:]))
+                ),
+            )
+            @warn "Initializing with `segment_idx = $nearest_idx"
+            bcf_info.segment_idx[1] = nearest_idx
+            bcf_info.segment_idx0[1] = nearest_idx
+            update_midmonth_data!(date, bcf_info)
+
+            # case 3: model date is within the first segment read from file
+        elseif (date - all_dates[segment_idx0]).value >= 0
+            @warn "On $date updating file data reads: file dates = [ $(all_dates[segment_idx]) , $(all_dates[segment_idx+1]) ]"
+            bcf_info.segment_length .= (all_dates[segment_idx + 1] - all_dates[segment_idx]).value
+            bcf_info.monthly_fields[1] .= scaling_function(
+                Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[segment_idx], varname, comms_ctx),
+                bcf_info,
+            )
+            bcf_info.monthly_fields[2] .= scaling_function(
+                Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[segment_idx + 1], varname, comms_ctx),
+                bcf_info,
+            )
+            bcf_info.segment_idx0[1] -= Int(1)
+        end
+
+        # case 4: date is at or after the last date in file
+    elseif (date - all_dates[end]).value >= 0
+        @warn "This time period is after BC data - using file from $(all_dates[end])"
         bcf_info.monthly_fields[1] .= scaling_function(
             Regridder.read_from_hdf5(
                 bcfile_dir,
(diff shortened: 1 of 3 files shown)


src/BCReader.jl
```diff
@@ -196,8 +195,9 @@ function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
         bcf_info.monthly_fields[2] .= deepcopy(bcf_info.monthly_fields[1])
         bcf_info.segment_length .= Int(0)
 
-    elseif Dates.days(date - all_dates[end - 1]) > 0 # for fini
-        @warn "this time period is after BC data - using file from $(all_dates[end - 1])"
+        # case 2: date is after the last date in file
+    elseif (date - all_dates[end]).value > 0 # for fini
```
juliasloan25, line 199: I think we can remove the comment "for fini" - not sure what that's meant to say

src/BCReader.jl
```diff
@@ -236,7 +235,9 @@ function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
             Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[Int(midmonth_idx + 1)], varname, comms_ctx),
             bcf_info,
         )
+        midmonth_idx = bcf_info.segment_idx[1] += Int(1)
```
juliasloan25, line 238: I think `midmonth_idx` doesn't need to be updated here since it isn't used in the function after this point, and will get reset to `bcf_info.segment_idx` the next time this function is called

src/BCReader.jl
```diff
@@ -185,9 +185,8 @@ function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
     midmonth_idx = bcf_info.segment_idx[1]
     midmonth_idx0 = bcf_info.segment_idx0[1]
 
-    if (midmonth_idx == midmonth_idx0) && (Dates.days(date - all_dates[midmonth_idx]) < 0) # for init
-        midmonth_idx = bcf_info.segment_idx[1] -= Int(1)
-        midmonth_idx = midmonth_idx < Int(1) ? midmonth_idx + Int(1) : midmonth_idx
+    # case 1: date is before the first date in the file
+    if (midmonth_idx == midmonth_idx0) && ((date - all_dates[midmonth_idx]).value < 0) # for init
```
juliasloan25, line 189: Why do we check `midmonth_idx == midmonth_idx0` in this case? I would expect it to be to handle the case where the first date in the file is the first date of the simulation, but then we should have `<=` instead of `<` in the second boolean expression?

Now that `midmonth_idx == midmonth_idx0` is in case 3 maybe we can remove it from here?

src/BCReader.jl
```diff
@@ -211,21 +211,20 @@ function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
...
             abs.(
                 parse(FT, TimeManager.datetime_to_strdate(date)) .-
                 parse.(FT, TimeManager.datetime_to_strdate.(all_dates[:]))
             ),
         )
-        # TODO test this
-        bcf_info.segment_idx[1] = midmonth_idx = midmonth_idx0 = nearest_idx
+        bcf_info.segment_idx .= nearest_idx
+        bcf_info.segment_idx0 .= nearest_idx
         @warn "init data does not correspond to start date. Initializing with `SIC_info.segment_idx[1] = midmonth_idx = midmonth_idx0 = $nearest_idx` for this start date"
```
juliasloan25, line 224: Can we modify this warning to remove `midmonth_idx` and midmonth_idx0 since those are just local to this function? Also the message is hardcoded for SIC - maybe we could instead say "Initializing with `segment_idx = segment_idx0 = $nearest_idx` for this start date"

src/BCReader.jl
```diff
@@ -211,21 +211,20 @@ function update_midmonth_data!(date, bcf_info::BCFileInfo{FT}) where {FT}
         bcf_info.monthly_fields[2] .= deepcopy(bcf_info.monthly_fields[1])
         bcf_info.segment_length .= Int(0)
 
-        # throw error when there are closer initial indices for the bc file data that matches this date0
-    elseif Dates.days(date - all_dates[Int(midmonth_idx + 1)]) > 2
+        # case 3: during initialization, warn and reset if there are closer file dates to the current date than the file date implied by midmonth_idx0+1
+    elseif (midmonth_idx == midmonth_idx0) && (date - all_dates[Int(midmonth_idx0) + 1]).value >= 0
```
juliasloan25, line 215: I'm a little confused about the logic here - It seems like `(date - all_dates[Int(midmonth_idx0) + 1]).value >= 0` implies that the current date is past the date of the initial index in the file, but does that mean that there's a closer file date to the current date than `all_dates[midmonth_idx0 + 1]`?

And the `midmonth_idx == midmonth_idx0` expression checks that we're on the initial date of the file, right? Maybe we could add some more comments clarifying these points

test/bcreader_tests.jl
```diff
@@ -267,20 +267,29 @@ for FT in (Float32, Float64)
             @test bcf_info.monthly_fields[2] == bcf_info.monthly_fields[1]
             @test bcf_info.segment_length[1] == Int(0)
 
-            #  case 3: date - all_dates[segment_idx + 1] > 2
+            #  case 3: date - all_dates[segment_idx + 1] >= 0
             reset_bcf_info(bcf_info)
-            date = DateTime(bcf_info.all_dates[bcf_info.segment_idx[1] + 1] + Dates.Day(3))
+            bcf_info.segment_idx[1] = bcf_info.segment_idx0[1]
+            date = DateTime(bcf_info.all_dates[bcf_info.segment_idx0[1] + 1] + Dates.Day(3))
```
juliasloan25, line 273: To test the edge case where `date - all_dates[segment_idx + 1] == 0`, maybe we should add a test for case 3 where the `+ Dates.Day(3)` here is removed

test/bcreader_tests.jl
```diff
@@ -267,20 +267,29 @@ for FT in (Float32, Float64)
...
                 ),
             )
+
             @test bcf_info.segment_idx[1] == bcf_info.segment_idx0[1] == nearest_idx
 
-            #  case 4: everything else
+            #  case 4: Dates.days(date - all_dates[segment_idx]) >= 0
+            reset_bcf_info(bcf_info)
+            bcf_info.segment_idx[1] = bcf_info.segment_idx0[1]
+            date = DateTime(bcf_info.all_dates[bcf_info.segment_idx0[1]] + Dates.Day(3))
```
juliasloan25, line 287: Similar to above - I think we should also test the case where `date - all_dates[segment_idx] == 0`

test/mpi_tests/bcreader_mpi_tests.jl
```diff
@@ -219,11 +218,11 @@ end
         @test bcf_info.monthly_fields[2] == bcf_info.monthly_fields[1]
         @test bcf_info.segment_length[1] == Int(0)
 
-        #  case 3: date - all_dates[segment_idx + 1] > 2
+        #  case 3: date - all_dates[segment_idx + 1] >= 0
         reset_bcf_info(bcf_info)
-        date = DateTime(bcf_info.all_dates[bcf_info.segment_idx[1] + 1] + Dates.Day(3))
+        bcf_info.segment_idx[1] = bcf_info.segment_idx0[1]
+        date = DateTime(bcf_info.all_dates[bcf_info.segment_idx0[1] + 1] + Dates.Day(3))
```
juliasloan25, line 224: same comments as the other test file about adding `== 0` test cases

juliasloan25 (commented):
I left a few comments! This piece of the code has always been a bit confusing to me so I want to make sure I totally understand the changes and that the tests are comprehensive before we merge

juliasloan25:
Btw, our macOS github actions are failing right now because of a bug introduced by Plots.jl v1.40.0. There's an issue open in their repo about this [here](https://github.com/JuliaPlots/Plots.jl/issues/4879). Maybe for now we can pin to v1.39 and update it later

LenkaNovak:
> Btw, our macOS github actions are failing right now because of a bug introduced by Plots.jl v1.40.0. There's an issue open in their repo about this [here](https://github.com/JuliaPlots/Plots.jl/issues/4879). Maybe for now we can pin to v1.39 and update it later

Sounds good, nice catch! 🚀 Thank you for the review!

src/BCReader.jl
LenkaNovak: It was more a clarification statement (since it shouldn't allocate), but if that's clear I'll remove it.

src/BCReader.jl
LenkaNovak: `midmonth_idx == midmonth_idx0` is always true upon initialization (see [here](https://github.com/CliMA/ClimaCoupler.jl/blob/8aa2ba0b23bd50142e2d7c6d622244c413860c4c/src/BCReader.jl#L163C1-L164C22)). Case 1 is a case where the simulation date is before the first file date (i.e. `all_dates[midmonth_idx0]`). Case 3 is for when upon iniltialization, the simulation date is after the last file date. I can clarify this with comments (and in ClimaUtilities, we can come up with a nicer solution).

src/BCReader.jl
LenkaNovak: Ooh, nice catch!

src/BCReader.jl
LenkaNovak: Yes, case 3 is for when upon iniltialization, the simulation date is after the segment that was selected by the user (i.e. my date is 1 March, but the first segment I asked for was 1 Jan - 1 Feb)

szy21:
I don't think I understand the logic of file reader enough, so maybe I can just ask you later. No need to wait for that to merge this PR though. Maybe it would be helpful to add some examples in the code comment, e.g.
```
# case 1: date is before the first date in the file, e.g. date: 19790210, file:19790215-19790315
```
(I'm not sure if this is exactly correct, but something like this)

src/BCReader.jl
```diff
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
...
-        )
-        bcf_info.monthly_fields[2] .= deepcopy(bcf_info.monthly_fields[1])
-        bcf_info.segment_length .= Int(0)
-
-    elseif Dates.days(date - all_dates[end - 1]) > 0 # for fini
-        @warn "this time period is after BC data - using file from $(all_dates[end - 1])"
+    segment_idx = bcf_info.segment_idx[1] # index of the current date in the file. [segment_idx, segment_idx+1] indexes the current segment between which we interpolate
+    segment_idx0 = bcf_info.segment_idx0[1] # reference index (first segment_idx - 1)
+
+    # upon initialization (segment_idx == segment_idx0) and not initialized after final file date
```
juliasloan25, line 188: comment should say before final file date in this case

src/BCReader.jl
```diff
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
...
+                    parse(FT, TimeManager.datetime_to_strdate(date)) .-
+                    parse.(FT, TimeManager.datetime_to_strdate.(all_dates[:]))
+                ),
+            )
+            @warn "Initializing with `segment_idx = $nearest_idx"
+            bcf_info.segment_idx[1] = nearest_idx
+            bcf_info.segment_idx0[1] = nearest_idx
+            update_midmonth_data!(date, bcf_info)
+
+            # case 3: model date is within the first segment read from file
```
juliasloan25, line 215: I think this should say current segment instead of first segment, since we're using `segment_idx`

src/BCReader.jl
```diff
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
...
+        elseif (date - all_dates[segment_idx]).value >= 0
+            @warn "On $date updating file data reads: file dates = [ $(all_dates[segment_idx]) , $(all_dates[segment_idx+1]) ]"
```
juliasloan25, line 217: can we remove the warning here? I find it a little annoying to print out so much stuff during simulations and tests

src/BCReader.jl
juliasloan25: actually I'm not sure what you mean by "not initialized after final file data"

I would think this if statement is checking if we're at initialization (`segment_idx == segment_idx0`), and before the final file date (`!((date - all_dates[end]).value >= 0)`)

src/BCReader.jl
```diff
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
...
-
-    elseif Dates.days(date - all_dates[end - 1]) > 0 # for fini
-        @warn "this time period is after BC data - using file from $(all_dates[end - 1])"
+    segment_idx = bcf_info.segment_idx[1] # index of the current date in the file. [segment_idx, segment_idx+1] indexes the current segment between which we interpolate
+    segment_idx0 = bcf_info.segment_idx0[1] # reference index (first segment_idx - 1)
+
+    # upon initialization (segment_idx == segment_idx0) and not initialized after final file date
+    if (segment_idx == segment_idx0) && !((date - all_dates[end]).value >= 0)
+        #  case 1: model date is before the first segment read from file
+        if (date - all_dates[segment_idx]).value < 0
```
juliasloan25, line 191: should this be `segment_idx0` if we want it to be the first segment read from the file?

src/BCReader.jl
```diff
@@ -182,22 +182,54 @@ The times for which data is extracted depends on the specifications in the
...
+            bcf_info.segment_length .= (all_dates[segment_idx + 1] - all_dates[segment_idx]).value
+            bcf_info.monthly_fields[1] .= scaling_function(
+                Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[segment_idx], varname, comms_ctx),
+                bcf_info,
+            )
+            bcf_info.monthly_fields[2] .= scaling_function(
+                Regridder.read_from_hdf5(bcfile_dir, hd_outfile_root, all_dates[segment_idx + 1], varname, comms_ctx),
+                bcf_info,
+            )
+            bcf_info.segment_idx0[1] -= Int(1)
```
juliasloan25, line 227: why do we decrement `segment_idx0` here? I thought it would be constant throughout the simulation

src/BCReader.jl
juliasloan25: also could you clarify in this comment that cases 1-3 are for initialization, and 4-6 are for after initialization please :)

juliasloan25:
Thank you for addressing the review comments, it's a lot more clear now! I like how you broke up the cases some more :) I had a few more questions/comments still

src/BCReader.jl
LenkaNovak: Sure, segment_idx and segment_idx0 equal in this case anyway. :)

Project.toml
```diff
@@ -35,7 +35,7 @@ DocStringExtensions = "0.8, 0.9"
 Insolation = "0.8"
 JLD2 = "0.4"
 NCDatasets = "0.11, 0.12, 0.13"
-Plots = "1.39.0"
+Plots = "=1.39.0"
```
juliasloan25, line 38: Gabriele wants to avoid pinning Plots, so I have a separate PR that changes the backend we use for plotting in our tests. I'm trying to merge it early tomorrow

src/BCReader.jl
juliasloan25: oh, right!

src/BCReader.jl
LenkaNovak: I think it's a good check that the files are being read, but I know what you mean about there being too many warnings. How about we leave them in the init and remove from what will be the monthly update?

juliasloan25: that sounds good!

Project.toml
juliasloan25: https://github.com/CliMA/ClimaCoupler.jl/pull/582

src/BCReader.jl
LenkaNovak: `segment_idx0` is a reference date which refers to the the first `segment_idx - 1` (comment on L185). I changed its definition, because this way it ensures that it is always lower than the `segment_idx` after initialization.

Project.toml
LenkaNovak: Ok sounds great, thanks!

src/BCReader.jl
juliasloan25: That makes sense! Could you update its description in the docstring for `bcfile_info_init` as well?

src/BCReader.jl
juliasloan25: This if statement will only be entered if we're at initialization because of the first check, so I think the comment should say `upon initialization (segment_idx == segment_idx0) and before the final file date`. If it's already initialized and before the final file date, it would enter case 5

juliasloan25 (approved):

src/BCReader.jl
LenkaNovak: I already did. The above is outdated ;-)

src/BCReader.jl
LenkaNovak: Sure! Good point!

src/BCReader.jl
juliasloan25: I just meant because right now it looks like it's `upon initialization (segment_idx == segment_idx0) and initialized before the final file date`, and shouldn't have the "initialized". Sorry if this is nit-picky!

LenkaNovak: Ahh I see, it was a bit ambiguous. I've changed it to `upon initialization (segment_idx == segment_idx0) with model date before the final file date` to be extra clear.

PR #582, opened 2024-01-31
use PythonPlot for test plots

juliasloan25:
## Purpose 
The conservation checker plots are currently failing on MacOS due to an incompatibility with the GR backend introduced in Plots.jl v1.40.0. This PR works around that bug by replacing the GR backend with PythonPlot for this test.


Sbozzolo:
Can GR be removed?

juliasloan25:
PythonPlot is only being used in `test` so we still need GR in the other envs (it's in `experiments/AMIP` and `perf`). I don't see GR in `test/Project.toml` so I think it's removed there. Though I'm not sure why I had to add PythonPlot explicitly if we didn't do that with GR before. If I remove it and run the tests I get:

> LoadError: ArgumentError: Package PythonPlot not found in current path.
 Run `import Pkg; Pkg.add("PythonPlot")` to install the PythonPlot package.

Sbozzolo:
> > Can GR be removed?
> 
> PythonPlot is only being used in `test` so we still need GR in the other envs (it's in `experiments/AMIP` and `perf`). I don't see GR in `test/Project.toml` so I think it's removed there. Though I'm not sure why I had to add PythonPlot explicitly if we didn't do that with GR before. If I remove it and run the tests I get:
> 
> > LoadError: ArgumentError: Package PythonPlot not found in current path.
> > Run `import Pkg; Pkg.add("PythonPlot")` to install the PythonPlot package.

If GR is incompatible with the latest MacOS, we should use PythonPlot everywhere. I am guessing you have to add PythonPlot because GR is the default backend (you can see that it is a required dependency in the [Project.toml](https://github.com/JuliaPlots/Plots.jl/blob/master/Project.toml)

PR #585, opened 2024-01-31
use Plots v1.40, Gr v0.73.2

juliasloan25:
## Purpose 
The MacOS/GR incompatibility was fixed in GR v0.73.2. This PR updates our environments to use that version


juliasloan25:
compat changes aren't needed

PR #588, opened 2024-02-01
remove output, plotting directories

### LINKED ISSUE #587: remove output, plotting directories
We currently have a folder `output/` in our main repo directory. This folder is empty except for a one-line README, so I think we can remove it

There's also a folder `plotting/`, which contains a README and one script for plotting fields. This is 3 years old and not used anymore, so I think we should remove this folder too.

juliasloan25:
## Purpose 
closes #587 

removes directories, updates reference to `output/` dir


LenkaNovak (approved):
LGTM 💥

Issue #390, opened 2023-08-16
O1.2.5 Atmos-land simulations on GPU at 1 SYPD on 4 A100s


LenkaNovak:
# The Climate Modeling Alliance
### Purpose

We need to promote the GPU capability into the coupler code, still sharing the same the distribution context between Atmos and the surfaces (which are on the same grid), as we do with MPI. 

This will be done in two stages, first making the minimal changes needed to get simulations and tests to run, then secondly propagating these changes throughout the repo and ensuring high code quality throughout.

### Cost/Benefits/Risks

b: faster AMIP + DYAMOND

### People and Personnel

- Lead: @sriharshakandala 
- Collaborators: @juliasloan25 @LenkaNovak 
- Reviewers: 

### Inputs

- ClimaAtmos release (with a fixed insolation)
- ClimaLSM GPUfication (for AMIP and DYAMOND)

### Results and Deliverables

- [x] GPU  Buildkite run with the target AMIP setup, which is within ~5% performance of the standalone Atmos run (as in the MPI case - though double check this).
- [x] key cases in Buildkite on GPU: slabplanet, target AMIP, DYAMOND (excluding and including topography)

### Task Breakdown And Schedule

```[tasklist]
### initial runs \[1 April 2024]
- [x] setup a DYAMOND sim on CPU for future GPU profiling https://github.com/CliMA/ClimaCoupler.jl/issues/457
- [x] update for the ClimaAtmos v0.16.3 for GPU compatibility #502
- [x] https://github.com/CliMA/ClimaCoupler.jl/issues/530
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/590
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/572
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/630
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/662
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/583
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/658
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/607
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/740
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/571
```
```[tasklist]
### GPU performance \[1 June 2024]
- [ ] https://github.com/CliMA/ClimaCoupler.jl/issues/705
[...]

PR #589, opened 2024-02-01
add GPU slabplanet file read albedo runs

### LINKED ISSUE #590: run GPU slabplanet with albedo from file
#530 gives us GPU compatibility for slabplanet with functional albedo. Next, we want to test the same simulation with albedo from prescribed files, both static and temporal.

Before we can get this to work, we need more support from the next release of ClimaLand (v0.7 or v0.8)

closed by #589 
part of #390

juliasloan25:
## Purpose 
Add GPU slabplanet runs reading albedo from a static input file and from a temporally-varying input file.

closes #590
part of #390 

note: this PR changes GPU runs to soft fail conservation checks. I've opened an issue here so we remember to revisit this: https://github.com/CliMA/ClimaCoupler.jl/issues/607
note also: we can't currently show animations for these runs because they aren't GPU-compatible, but we have an issue for this too https://github.com/CliMA/ClimaCoupler.jl/issues/613

### Current status
- conservation plots look different between CPU and GPU runs

### To do
- [x] double check that all runs are outputting plots


LenkaNovak (approved):
LGTM, Julia. I hadn't appreciated just how much the GPU runs depart from the CPU ones. This is beyond the floating point errors / small mismatches in energy checks. Let's try to target #607 next.

PR #599, opened 2024-02-05
Alphabetical config arguments

### LINKED ISSUE #595: Sort config alphabetically
## Components
- [x] alphabetical rearrangement of all configs
- [x] change name of the gpu run to preclude race conditions
- [x] fix path to the atmos standalone run in longruns

## Future
- could we apply this to climaformat.jl?

LenkaNovak:
closes #595 

----


.buildkite/pipeline.yml
```diff
@@ -348,8 +348,8 @@ steps:
       # GPU RUNS: slabplanet
       - label: "GPU Slabplanet: albedo from function"
         key: "gpu_slabplanet_albedo_function"
-        command: "julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/slabplanet_albedo_function.yml"
-        artifact_paths: "experiments/AMIP/output/gpu/slabplanet/slabplanet_albedo_function_artifacts/*"
+        command: "julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_slabplanet_albedo_function.yml"
```
juliasloan25, line 351: do you prefer to have a separate YAML file for the GPU run? We could just use the same config file for both CPU and GPU runs

config/longrun_configs/amip_n1_shortrun.yml
```diff
@@ -1,24 +1,24 @@
...
-z_stretch: false
-rayleigh_sponge: true
 alpha_rayleigh_uh: 0
 alpha_rayleigh_w: 10
+apply_limiter: false
 dt_cpl: 150
-dt: "150secs"
 dt_rad: "1hours"
+dt_save_to_sol: "100days"
+dt: "150secs"
```
juliasloan25, line 7: `dt` should be above `dt_cpl` for this to be alphabetical. I don't think it really matters as long as we have a convention, but if we want to unify this across repos, we might want to use the same convention that a function like `sort` would output. For example, we could sort all the files in `config/` by executing `find config/ -type f -exec sort -o {} {} \; ` from `ClimaCoupler.jl/`. We might find it useful to use that here and in atmos, land, etc.

juliasloan25 (commented):
I added a suggestion about using Linux's `sort` command to sort everything, which would move `dt` and `FLOAT_TYPE` for the files that have it. I think this would make it easier to standardize across repos, but if you'd rather keep this as-is that's also fine with me

config/longrun_configs/amip_n1_shortrun.yml
LenkaNovak: That's a good point! I guess not everyone is using VSCode, this is more universal. Just changed it.

.buildkite/pipeline.yml
LenkaNovak: Yes, otherwise this sometimes leads to race conditions when the CPU and GPU runs are sharing the same output folders in the same build (e.g. https://buildkite.com/clima/climacoupler-ci/builds/2550).

juliasloan25: got it! good to know

LenkaNovak: I agree though, it would be nicer to have a way of calling the same run yml with different architectures in our CI. Maybe we can bring this up in the next software meeting, since the other repos use the same system as us.

juliasloan25 (approved):

Issue #607, opened 2024-02-09
investigate GPU conservation


juliasloan25:
Our GPU runs have higher water and energy conservation than CPU runs with identical setups. The CPU error tends to be around `1e-5`, while the GPU error is around `1e-3` or `1e-4`. We should look into why the conservation is worse on GPU, and try to improve it.

For now (as of #589), we've set the GPU runs to soft fail if the conservation error is larger than `1e-3`, but we want them to be able to pass this threshold (and even a smaller one, ideally).

part of #390 
PR #614

Partially done in https://github.com/CliMA/ClimaCoupler.jl/issues/735

### Approaches to try
- add debug plots, as was done in [this branch](https://github.com/CliMA/ClimaCoupler.jl/compare/main...ln/interactive-debug)
  - requires ability to convert fields from GPU to CPU - see https://github.com/CliMA/ClimaCore.jl/pull/1619
- check float type stability
  - Atmos cache always initialized with NaN
  - Land state always initialized with Float64 in coupler
  - might not be relevant if our runs are already with Float64, but could affect Float32 runs

### Bucket info
ClimaLand has standalone global bucket runs on CPU and GPU that we compare. For all 3 albedo options, comparing the mean values of the states from CPU and GPU runs gives a difference on the order of 1e-15. For the temporal map albedo case, which runs for 50 days, we get:
```
mean(cpu_state .- gpu_state) = 1.433428363408908e-15
eps(Float64) = 2.220446049250313e-16
```
The functional and static map albedo cases show similar discrepancies, and run for 7 days each. These differences are much smaller than what we see in coupled runs, so the difference is probably not coming from the bucket model.

#### functional albedo
CPU
rse[end] = 1.6532404532423364e-5
rse[end] = 0.0005462468417393023
GPU
rse[end] = 7.841693994284924e-5
rse[end] = 0.0005103673696554503

 <img src="https://github.com/CliMA/ClimaCoupler.jl/assets/51397186/cfcfd42a-8dc4-459f-82ef-507a8d453936" alt="CPU analytic function" width="800"/>

#### static map
CPU
rse[end] = 1.5598104032642762e-5
rse[end] = 7.682827509990757e-5
GPU
rse[end] = 0.00012549802174549964
rse[end] = 0.0015490000882394841

[...]

PR #614, opened 2024-02-12
debug GPU conservation

juliasloan25:
## Purpose 
Debug conservation difference between CPU and GPU #607

## To-do
- save sim states to HDF5 daily
- only run CPU/GPU slabplanet sims on buildkite

## Content

----


LenkaNovak:
Note, small degradation of conservation total at radiation timestep on CPU (FT64, so small but not negligible):

Screen Shot 2024-03-21 at 7 05 51 AM
Screen Shot 2024-03-21 at 7 06 38 AM

LenkaNovak:
CPU (left) vs GPU (right) based on @juliasloan25's investigation: 
- significant deviations in states of all models between GPU and CPU after 1 day, e.g. for atmos:
Screen Shot 2024-03-21 at 7 25 35 AM
- given there was no deviation in states after 1 dt, let's try to keep radiation constant (`dt_rad` > 1day) as the next step.

juliasloan25:
> Note, small degradation of conservation total at radiation timestep on CPU (FT64, so small but not negligible):
> 
> Screen Shot 2024-03-21 at 7 05 51 AM Screen Shot 2024-03-21 at 7 06 38 AM

Which runs are these plots from? It looks like we have `dt_rad: 1hours` for most of the runs here, so I just to make sure we're attributing the conservation change to the right thing :)

LenkaNovak:
> > Note, small degradation of conservation total at radiation timestep on CPU (FT64, so small but not negligible):
> > ...
> 
> Which runs are these plots from? It looks like we have `dt_rad: 1hours` for most of the runs here (or at least the AMIP runs - slabplanet is using whatever our default is), so I just to make sure we're attributing the conservation change to the right thing :)

The plots above are was for slabplanet with static albedo. Thanks for setting off the new runs... 🤞

LenkaNovak:
Hmm, keeping the rad constant doesn't seem to help. Would it be possible to run for 400s and add more debug plots in the coupling loop (with `t` appended to the plots' titles). Hopefully that'll help us narrow this down.

juliasloan25:
> Hmm, keeping the rad constant doesn't seem to help. Would it be possible to run for 400s and add more debug plots in the coupling loop (with `t` appended to the plots' titles). Hopefully that'll help us narrow this down.

Sure! Do you think 400s will be enough? I could also run for something like 6 hours and plot every hour

LenkaNovak:
> > Hmm, keeping the rad constant doesn't seem to help. Would it be possible to run for 400s and add more debug plots in the coupling loop (with `t` appended to the plots' titles). Hopefully that'll help us narrow this down.
> 
> Sure! Do you think 400s will be enough? I could also run for something like 6 hours and plot every hour

I would try 400s it for now (in case the first time step is overridden by the initial `step!` and `reinit`, but hourly plots like you suggest would be the next thing to try ;-)

LenkaNovak:
I'm seeing differences in conservation but no visible difference in the states. 
Screen Shot 2024-03-21 at 10 38 42 AM


It is likely a round off error. The good thing is that we can see the divergence it after 400s, but to see where exactly it is coming from, we can
- 1. print the extrema of the plotted fields with more decimal places
- 2. save and exactly compare some key variable arrays (I would start with coupler fields and land fields)

juliasloan25:
I've added tests to compare the states of each component model simulation at the end of the simulation between CPU and GPU. Looking at slabplanet with static albedo run for 400s, we see that the land and ocean states are approximately equal between CPU and GPU, but the land state is not. The max difference in specific atmos variables is shown below. Given that the only state with differences after 400s is the atmosphere, it will be helpful to set up CPU/GPU conservation tests in atmos.

**slabplanet: albedo from static map (CPU vs GPU) - [build](https://buildkite.com/clima/climacoupler-ci/builds/3068#_)**
`atmos_ρe_tot`
```
julia> abs(maximum(cpu_atmos_state[:,1] .- gpu_atmos_state[:,1]))
0.005922795255173696
```
`atmos_ρq_tot`
```
julia> abs(maximum(cpu_atmos_state[:,2] .- gpu_atmos_state[:,2]))
1.5653893756352792e-10
```
`atmos_ρ`
```
julia> abs(maximum(cpu_atmos_state[:,3] .- gpu_atmos_state[:,3]))
2.7223778786833464e-7
```
`atmos_uₕ[1]`
```
julia> abs(maximum(cpu_atmos_state[:,4] .- gpu_atmos_state[:,4]))
0.17414431556971977
```
`atmos_uₕ[2]`
```
julia> abs(maximum(cpu_atmos_state[:,5] .- gpu_atmos_state[:,5]))
0.0863372049825557
```

experiments/AMIP/compare_cpu_gpu.jl
```diff
@@ -0,0 +1,111 @@
+import DelimitedFiles as DLM
+using Statistics
+import ClimaCoupler
```
juliasloan25, line 3: don't need this

juliasloan25:
this has been resolved in https://github.com/CliMA/ClimaCoupler.jl/pull/733

Issue #619, opened 2024-02-13
Enhance interactive debugging


LenkaNovak:
In the coming weeks fast interactive debugging tools will be essential. This PR builds on the existing debug_plots.jl in the following ways. 

- [x] add remaining coupler fields to the plots
- [x] enable plotting anomalies of coupler fields wrt their reference counterparts (e.g., initial conditions)

Therefore, plots of all exchanged quantities can now be generated using the single line
```
debug(cs, "<optional-directory-name>")
```
If anomaly plots are needed, the coupler fields state needs to be copied and inserted into the debug function when needed, e.g.:
```
csf0 = deepcopy(cs.fields)
...
debug(cs, "<optional-directory-name>", csf0)
```

PR #620, opened 2024-02-13
Enhance interactive debugging tools

LenkaNovak:
closes https://github.com/CliMA/ClimaCoupler.jl/issues/619

----

### DIFF
--- a/experiments/AMIP/user_io/debug_plots.jl
+++ b/experiments/AMIP/user_io/debug_plots.jl
@@ -5,33 +5,54 @@ using ClimaCoupler.Interfacer: ComponentModelSimulation, SurfaceModelSimulation
 
 # plotting functions for the coupled simulation
 """
-    debug(cs::CoupledSimulation, dir = "debug")
+    debug(cs::CoupledSimulation, dir = "debug", cs_fields_ref = nothing)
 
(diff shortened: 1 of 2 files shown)


experiments/AMIP/user_io/debug_plots.jl
```diff
@@ -5,33 +5,54 @@ using ClimaCoupler.Interfacer: ComponentModelSimulation, SurfaceModelSimulation
 
 # plotting functions for the coupled simulation
 """
-    debug(cs::CoupledSimulation, dir = "debug")
+    debug(cs::CoupledSimulation, dir = "debug", cs_fields_0 = nothing)
```
juliasloan25, line 8: `cs_fields_init` might be a more clear name than `cs_fields_0`. At first I thought the 0 was a type of indexing

juliasloan25 (approved):
Looks good! I just had a naming suggestion

experiments/AMIP/user_io/debug_plots.jl
LenkaNovak: Good point, but maybe let's change it to `cs_fields_ref`, as these fields may not necessarily be the initial fields.

PR #639, opened 2024-02-22
up dependencies (ClimaAtmos v0.21, ClimaLand v0.10)

juliasloan25:
## Purpose 
update for most recent dependency updates

note: The dataset used for the temporal map albedo has been updated in ClimaLand v0.10. The previous file had NaNs in high latitudes, and the dataset has been cleaned to replace these with 1. This may not be the best solution, but we figured it would be better than having NaNs

note 2: In the AMIP paperplots, net TOA fluxes look different as of this PR. We believe this is due to ClimaAtmos using ConstantAlbedo, which doesn't get reset with land albedo in the coupler. We've opened an issue to fix this so the atmosphere uses the albedo provided from land: https://github.com/CliMA/ClimaCoupler.jl/issues/653

## To-do
- [x] fix https://github.com/CliMA/ClimaAtmos.jl/issues/2709 in Atmos
  - not exactly a bug and not introduced in this PR, but would be nice to include fix in Atmos v0.21.0 release :)
- [x] use ClimaAtmos@0.21 instead of ClimaAtmos#main after release is merged into main https://github.com/CliMA/ClimaAtmos.jl/pull/2694
- [x] check buildkite plots

## Content
updates to:
- ClimaLand v0.10
- ClimaCore v0.12
- CLIMAParameters v0.9
- Thermodynamics 0.12
- ClimaAtmos v0.21

## PRs replaced by this one
- #615 
- #621 
- #622 
- #623
- #640 
- #618 

----


perf/Project.toml
```diff
@@ -15,6 +15,7 @@ CloudMicrophysics = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 Dierckx = "39dd38d3-220a-591b-8e3c-4c3a8c710a94"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+FFMPEG = "c87230d0-a227-11e9-1b43-d7ebe4e7570a"
```
LenkaNovak, line 18: note here, otherwise gtg! Thank you!

LenkaNovak (approved):

PR #642, opened 2024-02-26
remove kappa_4 config

szy21:
## Purpose 

Removes kappa_4 from the config files as it is not supported anymore.

## Content

----


LenkaNovak (approved):
LGTM

PR #644, opened 2024-02-27
initial Interfacer cleanup

juliasloan25:
## Purpose 
First steps of Interfacer cleanup. Some of it (e.g. restructuring) will be done in separate PRs so this one doesn't get too big.

closes #339 

### get_field
- [x] add get_field(surfacestub, :air_density)
- [x] remove get_field(atmossim, :cv_m) and get_field(atmossim, :gas_constant_air)
- [x] get_field, update_field: implement default for each variable we expect, output warning with maxlog = 1 for each
- [x] alphabetize get_field and update_field functions for readability

### surfacestub
- [x] add stub_init function (just wrap SurfaceStub constructor)

### renaming variables
- [x] rename co2_gm -> co2 (explain in docs global mean)
- [x] rename albedo -> surface_albedo (atmos update_field, surfaces get_field)
- [x] rename clean_sst, clean_sic -> scale_sst, scale_sic
- [x] rename `get_model_state_vector` -> `get_model_prog_state`

### restructuring
- [x] differentiate_turbulent_fluxes: dispatch on eisenman sea ice type, move to eisenman file (specific to that model)

### docs
- [x] note that update_field(::atmossim, turbulent_fluxes) required for partitionedfluxes option
- [x] note: update_sim! can be optionally overwritten, but not required
- [x] update_turb_fluxes_point required for surface model if using partitioned sf
- [x] get_model_state_vector required for checkpointing
- [x] note in docs: surface model precip only needs update, get not required
- [x] explain that CO2 from atmos is expected to be global mean
- [x] add units to docs

## QA
- [x] test update_field! warnings
- [x] check docs rendering

## To Do in future PRs
- [ ] component model files: rename each to `<model>.jl`, combine bucket_init and bucket_utils -> bucket.jl; separate required vs model-specific (internal) functions and objects in component model files - see #647 
- [ ] separate surfacestub from Interfacer/FieldExchanger https://github.com/CliMA/ClimaCoupler.jl/issues/669
  - move surfacestub to separate file outside of Interfacer.jl, and include from Interfacer.jl so it's still in that module
  - same for FieldExchanger.jl

## Remaining questions
- to run with partitioned fluxes, does the atmos model need to extend `update_turbulent_fluxes_point`? or just surface models as it is currently?
[...]

PR #647, opened 2024-02-28
restructure component models

juliasloan25:
## Purpose 
separate required vs model-specific (internal) functions and objects in component model files

follows off of #644 

## Content
- [x] distinguish between coupler-required and model-specific functions in component model files
- [x] merge bucket_init.jl and bucket_utils.jl
- [x] rename component model files to `<model>.jl`
- [x] alphabetize `get_field` and `update_field!` methods for easier comparison/readability

----


experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -110,17 +75,78 @@ function atmos_init(::Type{FT}, atmos_config_dict::Dict) where {FT}
...
+function get_model_prog_state(sim::ClimaAtmosSimulation)
+    return sim.integrator.u
+end
+
+"""
+    get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux_toa})
+
+Extension of Interfacer.get_field to get the net TOA radiation, which is a sum of the
+upward and downward longwave and shortwave radiation.
+"""
```
LenkaNovak, line 92: I noticed recently that sometimes the docstrings of extensions can sometimes cause the extensions to be ignored. It doesn't seem to be problematic here, but we should probably decide on a convention (Clima wide) as to how to deal with this. (not as part of this PR)

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -110,17 +75,78 @@ function atmos_init(::Type{FT}, atmos_config_dict::Dict) where {FT}
...
+function get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux_toa})
+    FT = eltype(atmos_sim.integrator.u)
+    # save radiation source
```
LenkaNovak, line 95: redundant comment

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -110,17 +75,78 @@ function atmos_init(::Type{FT}, atmos_config_dict::Dict) where {FT}
...
+get_field(sim::ClimaAtmosSimulation, ::Val{:air_density}) =
+    TD.air_density.(thermo_params, sim.integrator.p.precomputed.ᶜts)
+get_field(sim::ClimaAtmosSimulation, ::Val{:air_temperature}) =
+    TD.air_temperature.(thermo_params, sim.integrator.p.precomputed.ᶜts)
+get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation}) = sim.integrator.p.precipitation.col_integrated_rain # kg/m^2/s
 get_field(sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux_sfc}) =
     ClimaCore.Fields.level(sim.integrator.p.radiation.ᶠradiation_flux, half)
-get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation}) = sim.integrator.p.precipitation.col_integrated_rain # kg/m^2/s
 get_field(sim::ClimaAtmosSimulation, ::Val{:snow_precipitation}) = sim.integrator.p.precipitation.col_integrated_snow  # kg/m^2/s
 get_field(sim::ClimaAtmosSimulation, ::Val{:turbulent_energy_flux}) =
```
LenkaNovak, line 143: again we can remove the units here

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -298,64 +359,6 @@ function calculate_surface_air_density(atmos_sim::ClimaAtmosSimulation, T_S::Cli
     extrapolate_ρ_to_sfc.(Ref(thermo_params), ts_int, swap_space!(ones(axes(ts_int.ρ)), T_S))
```
LenkaNovak, line 359: `calculate_surface_air_density` one is required

LenkaNovak (approved):
It looks like the rebase removed some of the changes in the previous PR, e.g. `Spaces -> ClimaCore.Spaces` , and perhaps some of the imports. It may be good to look over the changes [here](https://github.com/CliMA/ClimaCoupler.jl/pull/668/files#diff-459c85f7df7d2d54eb1fb4a9fd1d5c1f7aaae748bc77a474c201d9581f04b0e1) just to double check. It would be good to check that there are no behavior changes both for slabplanet and AMIP before merging. Otherwise LGTM, thank you for doing this!  Approving, but I'm happy to take a look again if needed.

experiments/AMIP/components/atmosphere/climaatmos.jl
juliasloan25: good point!

experiments/AMIP/components/atmosphere/climaatmos.jl
juliasloan25: Ah we missed this in the docs - I'll update them

Do we need to require both `calculate_surface_air_density` and `get_field(:air_density)`?

juliasloan25: update: as discussed offline, these two functions return different values, which the docs have been updated to reflect

PR #657, opened 2024-03-01
add GPU AMIP runs with albedo from file

### LINKED ISSUE #583: run GPU AMIP sims with bucket albedo from file
We should add test cases to buildkite that exercise each of the bucket albedo options, since the file reading is a known pain point with GPU compatibility. We should probably add these cases for both slabplanet and AMIP, but maybe just adding for AMIP is enough.

part of #390
closed by #657 

### cases
- [x] albedo from function #573
- [ ] albedo from static global file
- [ ] albedo from temporal global file

juliasloan25:
## Purpose 
Builds off of #573 - copies the config file introduced there and adds two new runs based on it, using albedo from static map and temporal map instead of function.

The goal of the runs introduced in this PR is to ensure that the coupled simulation can run in the AMIP setup with both types of file reading. Because of this, the simulation length of these runs is only two timesteps. More complex, physically significant setups will be added to the longruns.

closes #583

## Content
- [x] simple AMIP on GPU with bucket albedo from static map (run for 2 timesteps)
- [x] simple AMIP on GPU with bucket albedo from temporal map (run for 2 timesteps)

----

### DIFF
--- a/config/model_configs/gpu_amip_albedo_static_map.yml
+++ b/config/model_configs/gpu_amip_albedo_static_map.yml
@@ -8,14 +8,14 @@ dt_save_to_sol: "1days"
 dz_bottom: 30
 dz_top: 3000
 h_elem: 4
-job_id: "gpu_amip_albedo_function"
-land_albedo_type: "function"
+job_id: "gpu_amip_albedo_static_map"
+land_albedo_type: "map_static"
 mode_name: "amip"
 moist: "equil"
 precip_model: "0M"
 rad: "gray"
 rayleigh_sponge: true
-run_name: "gpu_amip_albedo_function"
+run_name: "gpu_amip_albedo_static_map"
 t_end: "300secs"
 vert_diff: "true"
 z_elem: 50
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -396,6 +396,24 @@ steps:
           slurm_mem: 20GB
           slurm_gpus: 1
 
+      - label: "GPU AMIP: albedo from static map"
+        key: "gpu_amip_albedo_static_map"
+        command: "julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_amip_albedo_static_map.yml"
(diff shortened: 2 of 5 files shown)


config/model_configs/gpu_amip_albedo_static_map.yml
```diff
@@ -0,0 +1,22 @@
+alpha_rayleigh_uh: 0
+alpha_rayleigh_w: 10
+apply_limiter: false
+dt: "150secs"
+dt_cpl: 150
+dt_rad: "1hours"
+dt_save_to_sol: "1days"
+dz_bottom: 30
+dz_top: 3000
+h_elem: 16
```
LenkaNovak, line 10: Could we mimic the [functional albedo resolution config](https://github.com/CliMA/ClimaCoupler.jl/blob/js/gpu-amip-albedo/config/model_configs/gpu_amip_albedo_function.yml), or do these break for the lower resolution? 

Also, I noticed that the [temporal albedo slabplanet](https://github.com/CliMA/ClimaCoupler.jl/blob/js/gpu-amip-albedo/config/model_configs/gpu_slabplanet_albedo_temporal_map.yml) has a higher resolution than the other two albedo runs. Is it because it was breaking for helem=4?

LenkaNovak (approved):
Good idea to bring these in, thank you, @juliasloan25 . I just had some questions on why the runs for the different albedo types have different resolutions (unless they're breaking, we should make these consistent, and I would choose the lowest possible resolution, except the target run).

config/model_configs/gpu_amip_albedo_static_map.yml
juliasloan25: Good catch! I'll lower them to 4, I think it should be fine.

The temporal albedo slabplanet has a higher h_elem because I copied these from the non-GPU slabplanet runs, where the functional and static map albedo have `h_elem: 4` and temporal has `h_elem: 6`. I'm not sure why this is, and happy to change them all to 4.

LenkaNovak: Hmm, maybe it is unstable. 🤔 Worth trying though.

juliasloan25: they all work with the lower resolution :)

PR #659, opened 2024-03-01
add GPU DYAMOND runs

### LINKED ISSUE #658: run DYAMOND on GPU
We want to run the DYAMOND setup (introduced in #457) on GPU. This will be one of the target runs for our profiling work. Following our previous GPU-compatibility work, it's possible this run will be successful without much additional effort.

juliasloan25:
## Purpose 
closes #658

Only adding a longrun, no shortrun.
This run exceeds the memory available on P100s. Caltech's V100s have 16GB and 32 GB options, neither of which is large enough for this job, according to https://www.hpc.caltech.edu/resources. Instead of running on central like the rest of the longruns, this job will run on clima (which has A100s with 80GB of memory). I've opened an issue to address the allocations seen in this run: https://github.com/CliMA/ClimaCoupler.jl/issues/683

view run on buildkite here: https://buildkite.com/clima/climacoupler-longruns/builds/480#_

## Content
- [x] add config file based on `config/longrun_configs/dyamond_target.yml` for longrun
  - set `anim: false` for gpu-compatibility
- [x] add longrun using GPU

----

### DIFF
--- a/config/longrun_configs/gpu_dyamond_target.yml
+++ b/config/longrun_configs/gpu_dyamond_target.yml
@@ -1,0 +1,15 @@
+anim: false
+atmos_config_file: "config/longrun_configs/longrun_aquaplanet_dyamond.yml"
+dt_cpl: 50
+dt_save_state_to_disk: "0.5days"
+dt_save_to_sol: "0.5days"
+energy_check: false
+job_id: "gpu_dyamond_target"
+land_albedo_type: "map_temporal"
+mode_name: "amip"
+mono_surface: false
+monthly_checkpoint: false
+run_name: "gpu_dyamond_target"
+start_date: "19790301"
+t_end: "1days"
+turb_flux_partition: "CombinedStateFluxes"
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -45,6 +45,31 @@ steps:
       JULIA_NUM_PRECOMPILE_TASKS: 8
       JULIA_MAX_NUM_PRECOMPILE_FILES: 50
 
+  # DYAMOND AMIP: 1 day (convection resolving)
+  - label: "GPU AMIP SUPERFINE: dyamond_target"
+    key: "gpu_dyamond_target"
+    command:
+      - "echo $$JULIA_DEPOT_PATH"
+
+      - echo "--- Instantiate AMIP env"
+      - "julia --project=experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia --project=experiments/AMIP/ -e 'using Pkg; Pkg.precompile()'"
+      - "julia --project=experiments/AMIP/ -e 'using Pkg; Pkg.status()'"
+
+      - echo "--- Download artifacts"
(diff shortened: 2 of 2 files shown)


LenkaNovak (commented):
Is there a way we could request an H100 for this job only? I don't think the allocation enhancements will be addressed anytime soon. If that's possible, I would suggest commenting out the regular CI job for now, but retaining the longrun one, which we only run once a week on Sundays.

juliasloan25:
> Is there a way we could request an H100 for this job only? I don't think the allocation enhancements will be addressed anytime soon. If that's possible, I would suggest commenting out the regular CI job for now, but retaining the longrun one, which we only run once a week on Sundays.

ClimaAtmos has a separate buildkite pipeline that runs target GPU simulations on clima (see the [runs](https://buildkite.com/clima/climaatmos-target-gpu-simulations) and the [pipeline.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/gpu_pipeline/pipeline.yml) itself). I can implement the same thing for us

LenkaNovak:
> > Is there a way we could request an H100 for this job only? I don't think the allocation enhancements will be addressed anytime soon. If that's possible, I would suggest commenting out the regular CI job for now, but retaining the longrun one, which we only run once a week on Sundays.
> 
> ClimaAtmos has a separate buildkite pipeline that runs target GPU simulations on clima (see the [runs](https://buildkite.com/clima/climaatmos-target-gpu-simulations) and the [pipeline.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/gpu_pipeline/pipeline.yml) itself). I can implement the same thing for us

Does this allow us to specify the hardware though for just one run?

juliasloan25:
> > > Is there a way we could request an H100 for this job only? I don't think the allocation enhancements will be addressed anytime soon. If that's possible, I would suggest commenting out the regular CI job for now, but retaining the longrun one, which we only run once a week on Sundays.
> > 
> > 
> > ClimaAtmos has a separate buildkite pipeline that runs target GPU simulations on clima (see the [runs](https://buildkite.com/clima/climaatmos-target-gpu-simulations) and the [pipeline.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/gpu_pipeline/pipeline.yml) itself). I can implement the same thing for us
> 
> Does this allow us to specify the hardware for just one run though?

No, it would be a separate pipeline where this job would be run. I think this will be useful for GPU scaling runs too

config/longrun_configs/gpu_dyamond_target.yml
```diff
@@ -0,0 +1,15 @@
...
+dt_save_to_sol: "0.5days"
+energy_check: false
+job_id: "gpu_dyamond_target"
+land_albedo_type: "map_temporal"
+mode_name: "amip"
+mono_surface: false
+monthly_checkpoint: false
+run_name: "gpu_dyamond_target"
+start_date: "19790301"
+t_end: "1days"
```
LenkaNovak, line 14: If it's a long run, could we run it for longer (e.g. 50 days) or does the simulation crash? 👀

LenkaNovak (approved):
LGTM, thank you, @juliasloan25. Just had a question about the sim length.

PR #671, opened 2024-03-05
update GPU AMIP longrun agents

### LINKED ISSUE #662: fix GPU job agent requirements in pipeline
In our buildkite longruns pipeline, we specify agent requirements for the GPU AMIP runs (introduced in https://github.com/CliMA/ClimaCoupler.jl/pull/632) that are very difficult (impossible?) to satisfy. Because of this, these jobs are waiting for an agent for a very long time (see [here](https://buildkite.com/clima/climacoupler-longruns/builds/463#018dfc53-a65f-456c-8902-ac2d2a0e6e79)). We need to update [these jobs](https://github.com/CliMA/ClimaCoupler.jl/blob/385ee23fb3089d57f1319201fe3a918f0d7b3c37/.buildkite/longruns/pipeline.yml#L242-L260) to request GPUs and an appropriate amount of memory, but not multiple nodes/tasks. We should also figure out what the appropriate amount of memory is.

Update: the jobs take < 5GB of memory when run for a few timesteps, so the existing limits of 16-20GB should be sufficient (checked using `CUDA.memory_status`).

juliasloan25:
## Purpose 
fixes buildkite agent specification for GPU AMIP longruns

closes #662

note: running these jobs on clima shows memory usage around 5GB, so I've set the limit to 16GB

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -241,23 +241,19 @@ steps:
 
       - label: "GPU AMIP FINE: new target amip: topo"
         key: "gpu_amip_target_topo"
-        command: "mpiexec julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_amip_target_topo.yml"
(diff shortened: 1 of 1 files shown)


LenkaNovak (approved):
LGTM, thanks!

Sbozzolo (approved):

PR #673, opened 2024-03-06
add GPU AMIP scaling runs

### LINKED ISSUE #663: add GPU AMIP scaling runs
ClimaAtmos has scaling runs for the "aquaplanet chap" setup introduced in https://github.com/CliMA/ClimaAtmos.jl/pull/2724. These use 1, 2, and 4 GPUs each, and all run with MPI. We should setup analogous runs in ClimaCoupler using the same atmos setup/config files.

note: atmos prints SYPD at end of simulation. We'll need this too to plot SYPD vs # GPUs

This also requires setting up a new buildkite pipeline for coupler performance testing, analogous to ClimaAtmos' [Target GPU Simulations](https://buildkite.com/clima/climaatmos-target-gpu-simulations) pipeline. This pipeline will be triggered manually when we want to do performance testing, unlike the automated shortrun and longrun pipelines. I've set it up here: https://buildkite.com/clima/climacoupler-target-gpu-simulations

part of #390

juliasloan25:
## Purpose 
closes #663 

note: we want to use the `slurm_exclusive:` flag to get a GPU that's being used for only the performance job, but this seems not to be working when used in Atmos

Need to use ClimaAtmos#fdf1df4 commit to include PR containing atmos config files. Can't use main because of dependency incompatibilities.

## Content
- [x] add GPU performance buildkite pipeline
- [x] add SYPD calculation to driver (use calculation from atmos so we're consistent)
- [x] add configs for 1-, 2-, and 4-GPU AMIP runs based on ClimaAtmos [chap runs](https://github.com/CliMA/ClimaAtmos.jl/pull/2724/files#diff-fb685ae1505fe3b112706c060678cabc5fc09cb400ea82bcd435fda9112de3d5)
  - do for both strong and weak scaling
- [x] add runs in buildkite/gpu/pipeline.yml
- [ ] manually trigger pipeline to check that these runs succeed on clima

## Status
2 GPU error is preventing us from getting scaling results - see https://github.com/CliMA/ClimaCoupler.jl/issues/687. This needs to be addressed before we can set up reliable scaling runs

----


Sbozzolo:
See here:

https://github.com/CliMA/ClimaAtmos.jl/pull/2646/commits/448d22c25e66fa310c14b48ebb1659d6e17c5113

```diff
-  SLURM_GPU_BIND: none # https://github.com/open-mpi/ompi/issues/11949#issuecomment-1737712291
+  SLRUM_GRES_FLAGS: "allow-task-sharing"
```

src/Regridder.jl
```diff
@@ -507,6 +507,9 @@ function update_surface_fractions!(cs::CoupledSimulation)
     cs.surface_fractions.ice .= max.(min.(ice_d, FT(1) .- land_s), FT(0))
     cs.surface_fractions.ocean .= max.(FT(1) .- (cs.surface_fractions.ice .+ land_s), FT(0))
 
+    comms_ctx = axes(land_s).grid.topology.context
```
charleskawczynski, line 510:
```suggestion
    comms_ctx = ClimaComms.context(land_s)
```
I think we have a helper for this, so that you don't need to use internals.

.buildkite/gpu/pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+  #         slurm_mem: 32G
+
+  - group: "DYAMOND GPU weak scaling"
+    steps:
+
+      - label: "GPU AMIP DYAMOND - weak scaling - 1 GPU"
+        key: "gpu_amip_dyamond_ws"
+        command:
+          - >
+            julia --threads=3 --color=yes --project=experiments/AMIP experiments/AMIP/coupler_driver.jl
```
charleskawczynski, line 95: Can we add an `nsys profile` to the non-MPI jobs?

LenkaNovak:
@charleskawczynski @dennisYatunin does increasing the ref counter seem like a good way forward here, or is the behavioral change worrying here?

juliasloan25:
CHAP is no longer our target; we want to set up scaling runs using the benchmarks pipeline and setups in the future

PR #680, opened 2024-03-07
use NCDatasets nomissing

### LINKED ISSUE #679: fix NCDatasets missing error
Sometimes, data read from a NetCDF file is of type `Union{Missing, FT}`, even though the data doesn't actually contain any missings. NCDatasets provides a function `nomissing` to convert Union{Missing, FT} type to FT when reading from input datasets. This fixes `ERROR: cannot convert MissingtoFloat64` which occasionally pops up on buildkite

This was done previously in ClimaLand (see [here](https://github.com/CliMA/ClimaLand.jl/pull/482/files#diff-f9e76ad8c8a7afdffe40057e6e9cefae9e0dca3fe839c98f5ffae5088c7fe4e6R297-R298))

juliasloan25:
## Purpose 
closes #679

### DIFF
--- a/src/Regridder.jl
+++ b/src/Regridder.jl
@@ -207,7 +207,8 @@ function hdwrite_regridfile_rll_to_cgll(
     # read the remapped file with sparse matrices
     offline_outvector, coords = NCDataset(datafile_cgll, "r") do ds_wt
         (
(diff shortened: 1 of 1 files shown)


akshaysridhar:
I've tested this fix locally (as discussed offline yesterday). Thanks for adding the fix!

akshaysridhar (approved):

PR #681, opened 2024-03-07
add amip requirements to docs

### LINKED ISSUE #652: add checklist of AMIP requirements to docs
Before our release, we should make sure that our docs contain everything we want them to, including a checklist of AMIP requirements

szy21:
## Purpose 

Closes #652 

## Content

----

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -24,6 +24,51 @@ or coupling components (using MPI) that are running on different processes.
 |:--:|
 | *ClimaCoupler.jl allows for independent development of interchangeable component models.* |
 
+
+## Current AMIP components
+
+### Atmosphere
+
+Dynamical core:
+- Equation: non-hydrostatic and fully compressible
+- Prognostic variables: Density, velocity components, total energy, total specific humidity
+- Spatial discretization: Spectral element in the horizontal, finite difference in the vertical, cubed sphere
+- Time stepping: Implicit-explicit additive Runge–Kutta method
+
+Radiation: A scheme based on RRTM for General circulation model applications—Parallel (RRTMGP)
+
+Convection and turbulence: Diagnostic Eddy-diffusivity Mass-Flux scheme with prognostic turbulent kinetic energy
+
+Microphysics: 0-moment scheme, where cloud condensate is removed with a constant timescale
+
+Surface fluxes: A scheme based on Monin-Obukhov similarity theory, with constant roughness lengths over land and ocean
+
+Orographic gravity wave drag: None
+
+Non-orographic gravity wave drag: None
+
+Aerosols and chemistry: None
+
+### Land
+
+Dynamical core:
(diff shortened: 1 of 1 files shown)


szy21:
@juliasloan25 @LenkaNovak Here is a draft of a brief amip description. I wonder:
1. Would you like it to be more detailed?
2. Would you like to have references?
3. For components with `None`, would you like them to be removed?

docs/src/index.md
```diff
@@ -24,6 +24,51 @@ or coupling components (using MPI) that are running on different processes.
...
+Orographic gravity wave drag: None
+
+Non-orographic gravity wave drag: None
+
+Aerosols and chemistry: None
+
+### Land
+
+Dynamical core:
+- Equation: Following Manabe bucket hydrology scheme
```
juliasloan25, line 55: Could be good to add a citation here!

docs/src/index.md
```diff
@@ -24,6 +24,51 @@ or coupling components (using MPI) that are running on different processes.
...
+- Prognostic variables: Temperature, water content, snow water content
+- Spatial discretization: Finite difference, single column
```
juliasloan25, line 57: is single column correct? I thought for global runs we have many columns over the sphere

juliasloan25 (commented):
This looks good to me, Zhaoyi! Now that I see the details, this might fit better in the experiments/AMIP/README.md file since that's where the AMIP driver lives and the coupler supports other setups too.

docs/src/index.md
szy21: Ah, yes, I meant there is no horizontal transport between each column. I'll think about a better description.

LenkaNovak (commented):
Looks great, thank you, @szy21 ! 

Shall we add more details on insolation and EDMF? I've been notionally thinking something like [this](https://pcmdi.llnl.gov/mips/amip/home/Documentation/20gfdl.html#RTFToC2), but maybe being somewhat less specific on the details that are changing is fine. :)

docs/src/index.md
```diff
@@ -24,6 +24,51 @@ or coupling components (using MPI) that are running on different processes.
...
+- Time stepping: Explicit additive Runge–Kutta method
+
+Land surface albedo: 
+- Bare ground: Prescribed from files
+- Snow: Constant
```
LenkaNovak, line 62: There is a simple prognostic snow in the bucket model.

juliasloan25:
to answer your questions...

> Would you like it to be more detailed?
Would you like to have references?
For components with None, would you like them to be removed?

1. I think the detail level is good, but I'm less familiar with the physics so will default to you two on that
2. References would be nice if possible, but if we don't have them it's okay too
3. To me it seems like it's good to keep "none" for components we don't have so it's clear what is and isn't included

szy21:
hmm for some reason my branch is updated but this PR is not. I'll try close and reopen.

PR #684, opened 2024-03-09
add toml for amip_target_topo_diagedmf

### LINKED ISSUE #641: Fix amip with diagnostic edmf longrun
The amip target with diagnostic edmf longrun is currently failing in the most recent [build](https://buildkite.com/clima/climacoupler-longruns/builds/453#018de1db-cc9d-4c59-bcdb-130b0c99c440). We should fix it soon. Maybe try increasing hyperdiffusion and/or sponge.

szy21:
## Purpose 

Adds a toml file for `amip_target_topo_diagedmf` longrun (which increases `alpha_rayleigh_w`), and sets `use_reference_state` to true. Closes #641 

## Content

----

### DIFF
--- a/config/longrun_configs/amip_target_topo_diagedmf.yml
+++ b/config/longrun_configs/amip_target_topo_diagedmf.yml
@@ -2,6 +2,7 @@ FLOAT_TYPE: "Float32"
 anim: true
 apply_limiter: false
 atmos_config_file: "config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_allsky_diagedmf_0M.yml"
+coupler_toml_file: "toml/amip_target_topo_diagedmf.toml"
 dt: "100secs"
 dt_cloud_fraction: "1hours"
 dt_cpl: 100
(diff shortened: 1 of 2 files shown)


szy21:
Working build: https://buildkite.com/clima/climacoupler-longruns/builds/479#018e1f4f-1794-4506-b5f5-356290548732

LenkaNovak (approved):
LGTM, thanks! Btw, why don't we want to have an equivalent run in atmos?

szy21:
> LGTM, thanks! Btw, why don't we want to have an equivalent run in atmos?

Just to avoid too many longruns on the clima cluster and to save resources in general, my current hierarchy for most of the physical parameterizations is aquaplanet dycore --> aquaplanet with idealized insolation and physics --> atmos + land with topography, realistic insolation and physics. We could add more to this hierarchy, e.g. aquaplanet with topography, or atmos + land without topography (which you have as a target in the coupler now), and with idealized or realistic insolation. But I think there will be too many experiments to track if we add everything, so I choose to run other simulations when we see instabilities. Once we have more experience on which hierarchy would be most useful, we can adjust the longruns accordingly.

No matter whether we have an equivalent run in atmos, I think we may need a different toml file in coupler, as it is possible that we need slightly different parameters for stable simulations with and without land.

PR #686, opened 2024-03-11
Add missing docstrings

### LINKED ISSUE #676: Add missing docstrings
Postprocessor and other modules have some missing docstrings (see docs).

LenkaNovak:
closes #676 

----

### DIFF
--- a/docs/src/conservation.md
+++ b/docs/src/conservation.md
@@ -2,15 +2,20 @@
 
 If the model is a physically closed system (e.g., in the `slabplanet` configuration with free slip conditions), it should conserve mass (including water), energy and momentum. The conservation checker logs global conservation.
 
-Only energy and water are currently implemented. 
+Only energy and water are currently implemented.
 
-Note that kinetic energy is not included in the calculation of the global energy, reflecting the formulation on `ClimaAtmos`, which assumes that kinetic energy is negligible in comparison with the moist static energy components. 
+Note that kinetic energy is not included in the calculation of the global energy, reflecting the formulation on `ClimaAtmos`, which assumes that kinetic energy is negligible in comparison with the moist static energy components.
 
-## Coupler Object API
+## ConservationChecker API
 
 ```@docs
 ClimaCoupler.ConservationChecker.EnergyConservationCheck
 ClimaCoupler.ConservationChecker.WaterConservationCheck
 ClimaCoupler.ConservationChecker.check_conservation!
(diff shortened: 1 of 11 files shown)


src/Diagnostics.jl
```diff
@@ -13,8 +13,18 @@ using ClimaComms
 
 export get_var, init_diagnostics, accumulate_diagnostics!, save_diagnostics, TimeMean
 
+"""
+    AbstractOutputGroup
+
+Abstract type for ClimaCoupler's output diagnostics groups.
```
juliasloan25, line 19: What is an output diagnostic group meant to do?

src/Diagnostics.jl
```diff
@@ -23,8 +33,18 @@ struct DiagnosticsGroup{S, NTO <: NamedTuple} <: AbstractOutputGroup
...
+    AbstractDiagnosticsOperations
+
+Abstract type for operations to be performed on ClimaCoupler's diagnostics.
+"""
 abstract type AbstractDiagnosticsOperations end
 
+"""
+    TimeMean{C}
+
+Defines a concrete operation type for time-averaged diagnostics. The counter `ct` is used to accumulate the sum of the diagnostics.
```
juliasloan25, line 46: It would be nice to limit line length to 80-ish characters

docs/src/conservation.md
```diff
@@ -13,4 +13,5 @@ ClimaCoupler.ConservationChecker.EnergyConservationCheck
 ClimaCoupler.ConservationChecker.WaterConservationCheck
 ClimaCoupler.ConservationChecker.check_conservation!
 ClimaCoupler.ConservationChecker.plot_global_conservation
+ClimaCoupler.ConservationChecker.surface_water_gain_from_rates
```
juliasloan25, line 16: This function looks like it's internal to the ConservationChecker and not exported, so maybe it would be good to put it under a section `# ConservationChecker Internal Functions` in the docs

juliasloan25 (approved):
Looks good, thank you Lenka!

src/Diagnostics.jl
LenkaNovak: Sure, I added "Each diagnostic group should
contain fields that are of the same type and size, so the extended methods for the group's
operation functions work in the same way for all the fields."

docs/src/conservation.md
LenkaNovak: Good catch!

PR #699, opened 2024-03-18
remove closures in driver

juliasloan25:
## Purpose 
Remove closures over global variables in `coupler_driver.jl` to improve performance.

This includes changes for:
- comms_ctx
- mono_surface
- turbulent_fluxes
- thermo_params

### DIFF
--- a/experiments/AMIP/coupler_driver.jl
+++ b/experiments/AMIP/coupler_driver.jl
@@ -142,7 +142,7 @@ then `ClimaComms` automatically selects the device from which this code is calle
 
 using ClimaComms
 comms_ctx = get_comms_context(parsed_args)
-const pid, nprocs = ClimaComms.init(comms_ctx)
+ClimaComms.init(comms_ctx)
 
 #=
 ### I/O Directory Setup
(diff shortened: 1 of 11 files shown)


config/model_configs/interactive_debug.yml
```diff
@@ -17,6 +17,6 @@ rad: "gray"
 run_name: "interactive_debug_run"
 start_date: "19790301"
 surface_setup: "PrescribedSurface"
-t_end: "20days"
+t_end: "200secs"
```
LenkaNovak, line 20: can we change this back to 20days when it's ready?

experiments/AMIP/coupler_driver.jl
```diff
@@ -573,32 +589,22 @@ depend on initial conditions of other component models than those in which the v
...
 
-# 4.surface vapor specific humidity (`q_sfc`): step surface models with the new surface density to calculate their respective `q_sfc` internally
+# 3.surface vapor specific humidity (`q_sfc`): step surface models with the new surface density to calculate their respective `q_sfc` internally
 ## TODO: the q_sfc calculation follows the design of the bucket q_sfc, but it would be neater to abstract this from step! (#331)
 step!(land_sim, Δt_cpl)
 step!(ocean_sim, Δt_cpl)
 step!(ice_sim, Δt_cpl)
 
-# 5.turbulent fluxes: now we have all information needed for calculating the initial turbulent surface fluxes using the combined state
+# 4.turbulent fluxes: now we have all information needed for calculating the initial turbulent surface fluxes using the combined state
```
LenkaNovak, line 607: `turbulent_fluxes` and `thermo_params` can also be called from cs in this section.

LenkaNovak (commented):
Thanks for doing this, @juliasloan25. So far, only a couple of minor comments and happy to approve once CI passes.

config/model_configs/interactive_debug.yml
juliasloan25: oops yes, I didn't mean to include this

src/Interfacer.jl
```diff
@@ -65,6 +67,8 @@ struct CoupledSimulation{
     diagnostics::TD
     callbacks::NTC
     dirs::NTP
+    turbulent_fluxes::TF
+    thermo_params::TP
```
charleskawczynski, line 71: We could make this a keyword struct, if that simplifies how it's called. Either way is fine, though.

charleskawczynski (approved):

src/Interfacer.jl
juliasloan25: I think I'll just keep it as it is, but I'll keep this in mind!

PR #703, opened 2024-03-19
Add concurrent group for init

Sbozzolo:
Ensures that only one init job is run at the same time across pipelines.


LenkaNovak:
Looks good, but how do I know it's working? 🤔

Sbozzolo:
Just start two pipelines at the same time (e.g., from the web interface). You should see that one will be waiting for an agent until the other one has finished (the init step)

LenkaNovak (approved):
Sounds good, tried and tested. Thank you, @Sbozzolo !

Sbozzolo:
Thanks for testing!

PR #706, opened 2024-03-20
add coupler summary table

### LINKED ISSUE #705: add coupler summary table
We want to output a table displaying some information about the current state of the coupler. This table/info should be output to the #coupler-report channel on slack.

The runs will include:
- a CPU atmos-only run
- a GPU atmos-only run
  - config file: [config/gpu_configs/gpu_aquaplanet_diagedmf.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/gpu_configs/gpu_aquaplanet_diagedmf.yml)
- a CPU coupled (AMIP) run
- a GPU coupled (AMIP) run

For each of these jobs, we will output the following quantities:
- SYPD
- CPU vs GPU state (max, median, and mean of difference, as is calculated in [land](https://github.com/CliMA/ClimaLand.jl/blob/main/experiments/standalone/Bucket/compare_gpu_cpu_output.jl)
  - update: I'll add these in a separate PR, since this is currently being implemented in #614
- allocations at the end of the simulation
  - note that there is no equivalent calculation between CPU and GPU, so these numbers won't be directly comparable

part of https://github.com/CliMA/ClimaCoupler.jl/issues/693

## To Do
- [x] add config files for the 2 coupled runs
  - based on atmos [gpu_aquaplanet_diagedmf.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/gpu_configs/gpu_aquaplanet_diagedmf.yml)
- [x] add config files for the 2 atmos-only runs 
  - it might be nice to just access these from ClimaAtmos.jl, but I can't find a cpu run that's equivalent to the gpu config
- [x] set up atmos_driver to run from different config files
- [x] add runs to benchmarks pipeline - [ClimaCoupler CPU/GPU Benchmarks](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks) (from #673)
- [x] add in SYPD calculation (from #673)
- [x] add in allocation printing  (from #691)
- [x] save these to txt files (as is done with states in #614)
- [x] use PrettyTables.jl to create table displaying info read in from these files
- [x] update benchmarks pipeline to send this table to slack channel

## To Do in separate issues/PRs
- [ ] add atmos-only runs without diagnostic edmf
- [ ] add gpu scaling (1, 2, 4 gpus)
- [ ] add cpu/gpu state comparison columns
- [ ] save output as .pdf rather than .txt (use latex backend for PrettyTables)

juliasloan25:
## Purpose 
closes #705 

#### status 4/23
view passing build in coupler benchmarks pipeline [here](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/92)
view output to slack [here](https://climate-machine.slack.com/archives/C04S6MBSDC4/p1713907568438739)

### To do
- [x] get CPU runs working
- [x] check slack output
- [x] verify SYPD calculations (make sure all units are seconds, compare to estimate from buildkite timing)
- [x] add individual rows for CPU allocations vs GPU allocations (shouldn't be compared directly)
- [x] include buildkite id in output
- [x] use 4 GPUs

### To do in separate PR(s)
- save output as .pdf rather than .txt
  - use latex backend for PrettyTables.jl?
- add atmos-only run without diagnostic edmf
  - follow ClimaAtmos [gpu_aquaplanet_dyamond](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/gpu_configs/gpu_aquaplanet_dyamond_ss_1process.yml)
- add cpu/gpu state comparison columns
- pass in config file to benchmark script
  - use this to get run names (instead of passing these in), resolution & dt (instead of hardcoding these in table)

note: I was using `CUDA.memory_status` to get the GPU memory usage, but this varied wildly between runs (3-30GB). I'm not sure why this is, but it doesn't seem to be a reliable metric so I won't be using it going forward. Here is the code block calling that function, in case we want to return to it in the future:
```julia
    if comms_ctx.device isa ClimaComms.CUDADevice
        # If `io` is `stdout`, print the memory status, otherwise store in buffer to return
        CUDA.memory_status(io)
        if io != stdout
            # Get the memory usage in GB by parsing the output of `CUDA.memory_status`
            str = String(take!(io))
            println(str)
            gpu_allocs_GB = "GPU: " * split(split(str, '(')[2], '/')[1]
        end
    end
```


.buildkite/longruns/pipeline.yml
```diff
@@ -300,11 +299,11 @@ steps:
 
   # DYAMOND AMIP: 1 day (convection resolving)
   - label: "GPU AMIP SUPERFINE: dyamond_target"
-    key: "gpu_dyamond_target"
+    key: "gpu_longrun_amip_dyamond"
```
juliasloan25, line 302: I renamed this to be consistent with the atmos equivalent ([longrun_aquaplanet_dyamond.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/longrun_configs/longrun_aquaplanet_dyamond.yml))

config/model_configs/amip_diagedmf.yml
```diff
@@ -0,0 +1,15 @@
+anim: false
+atmos_config_file: "config/gpu_configs/gpu_aquaplanet_diagedmf.yml"
```
juliasloan25, line 2: This is using a config file from atmos. It's duplicated in our `config/atmos_configs/gpu_aquaplanet_diagedmf.yml`, but maybe it would be better for this to use the version that's in ClimaCoupler, so that even if atmos changes their setup, we continue to compare analogous atmos/coupled simulations

config/model_configs/amip_diagedmf.yml
```diff
@@ -0,0 +1,15 @@
...
+dt_cpl: 100
+dt_save_state_to_disk: "12hours"
```
juliasloan25, line 4: This is 12hours for the AMIP setups, and Inf for the aquaplanet setups. Which of these do you think we would want? @LenkaNovak

juliasloan25:
In some runs, atmos outputs an SYPD that's much larger than the SYPD calculated here by timing the coupling loop. For example, in the "GPU slabplanet: albedo from function" run [here](https://buildkite.com/clima/climacoupler-ci/builds/3468#018f0bdc-948c-451a-8b64-2bb6de990648), atmos prints an SYPD around 17, and we calculate it to be 8.15. I'm not sure where this difference is coming from since I'm using `CA.EfficiencyStats` and `CA.simulated_years_per_day` to calculate SYPD in the coupler driver

config/atmos_configs/aquaplanet_diagedmf.yml
```diff
@@ -0,0 +1,29 @@
+job_id: "aquaplanet_diagedmf"
```
juliasloan25, line 1: I will alphabetize the aquaplanet configs so we can more easily compare to the amip ones

config/model_configs/amip_diagedmf.yml
juliasloan25: I could copy [this line](https://github.com/CliMA/ClimaCoupler.jl/blob/69a396915161bea19c0abe6919d7be0e3062e3f8/experiments/AMIP/components/atmosphere/climaatmos.jl#L252) to use ClimaCoupler instead of ClimaAtmos, and keep support for both options. I'm not sure of the best way to do this - maybe we add another option to the coupler configs (something like `coupler_atmos_config_file`, but that might be confusing)

test/component_model_tests/climaatmos_standalone/atmos_driver.jl
```diff
@@ -1,12 +1,119 @@
...
+converted to text to read (e.g. via a call`write(file, alloc_info)`).
+In the CPU case, show and return the memory footprint of the provided object(s) in GB.
+Note that these two cases provide different information, and should not be
+directly compared.
+# Arguments
+`comms_ctx`: the communication context being used to run the model
+`objects`: Dict mapping objects whose memory footprint is displayed in the CPU case to their names
+`io`: The input/output stream to use. Currently, `stdout` and `IOBuffer` types are supported.
+"""
+function show_memory_usage(comms_ctx, objects, io::Union{Base.TTY, IOContext, IOBuffer} = stdout)
```
LenkaNovak, line 38: Is this the same as the one above? If so, could we add this to Utilities and use it from both drivers?

LenkaNovak (commented):
Thanks so much for implementing this, @juliasloan25 ! A super useful PR! 🎉 I just had a few comments / suggestions.

LenkaNovak:
I'm also a little surprised about the SYPD for all runs reported on Slack. For example, this [build](https://buildkite.com/clima/climacoupler-longruns/builds/645#018f082d-1b27-4321-bbda-b7dcbb5f30a0) finished running 0.25 sim years in ~0.25 wallclock days (including initialization). I think the config should be the same, but the slack results imply 100x slower. 🤔

config/longrun_configs/gpu_longrun_amip_dyamond.yml
```diff
@@ -4,12 +4,12 @@ dt_cpl: 50
...
-job_id: "gpu_dyamond_target"
+job_id: "gpu_longrun_amip_dyamond"
 land_albedo_type: "map_temporal"
 mode_name: "amip"
 mono_surface: false
 monthly_checkpoint: false
-run_name: "gpu_dyamond_target"
+run_name: "gpu_longrun_amip_dyamond"
 start_date: "19790301"
 t_end: "1days"
```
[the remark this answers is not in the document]
juliasloan25: I think 1 day is okay for the longrun! This isn't run in the benchmarks pipeline so it doesn't need the same t_end that those runs have

test/component_model_tests/climaatmos_standalone/atmos_driver.jl
juliasloan25: good point!

.buildkite/benchmarks/pipeline.yml
```diff
@@ -0,0 +1,111 @@
...
+  JULIA_MAX_NUM_PRECOMPILE_FILES: 100
+  OPENBLAS_NUM_THREADS: 1
+  OMPI_MCA_opal_warn_on_missing_libcuda: 0
+  SLURM_KILL_BAD_EXIT: 1
+  SLURM_GRES_FLAGS: "allow-task-sharing"
+  GPU_DYAMOND_CONFIG_PATH: "config/gpu_configs/gpu_dyamond"
+  GPU_DYAMOND_WS_CONFIG_PATH: "config/gpu_configs/gpu_dyamond_ws"
+  CLIMAATMOS_GC_NSTEPS: 10
+  ATMOS_CONFIG_PATH: "config/atmos_configs"
+  MODEL_CONFIG_PATH: "config/model_configs"
```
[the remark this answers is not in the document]
juliasloan25: Good point! I made the following changes:

- model_configs, mpi_configs -> ci_configs
- atmos_configs, benchmark-related model_configs -> benchmark_configs
- perf_configs and longrun_configs stay the same

.buildkite/benchmarks/pipeline.yml
```diff
@@ -0,0 +1,111 @@
...
+    env:
+      JULIA_NUM_PRECOMPILE_TASKS: 8
+      JULIA_MAX_NUM_PRECOMPILE_FILES: 50
+
+  - wait
+
+  - group: "CPU benchmarks"
+    steps:
+      - label: "CPU aquaplanet with diagnostic EDMF"
+        key: "aquaplanet_diagedmf"
```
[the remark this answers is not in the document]
juliasloan25: I renamed to just `climaatmos_diagedmf`, which I agree is more clear :)

config/model_configs/amip_diagedmf.yml
juliasloan25: Thank you for such clear instructions! I added a new flag `use_coupler_diagnostics` and it's used in the driver to toggle the diagnostics

juliasloan25:
> In some runs, atmos outputs an SYPD that's much larger than the SYPD calculated here by timing the coupling loop. For example, in the "GPU slabplanet: albedo from function" run [here](https://buildkite.com/clima/climacoupler-ci/builds/3468#018f0bdc-948c-451a-8b64-2bb6de990648), atmos prints an SYPD around 17, and we calculate it to be 8.15. I'm not sure where this difference is coming from since I'm using `CA.EfficiencyStats` and `CA.simulated_years_per_day` to calculate SYPD in the coupler driver

After speaking with Charlier, it sounds like this might be because we're using `Base.@elapsed` to time the coupling loop, and this doesn't work well with CUDA kernels. I've changed it to use `ClimaComms.@elapsed`, which will call `CUDA.@elapsed` when we're running on GPU and should give more accurate numbers

config/model_configs/amip_diagedmf.yml
juliasloan25: That makes sense! I added a flag that's only used by the coupled benchmark runs, which specifies that we want to use an atmos config file from ClimaCoupler instead of ClimaAtmos

experiments/AMIP/coupler_driver.jl
```diff
@@ -671,7 +712,8 @@ function solve_coupler!(cs)
 
     ClimaComms.iamroot(comms_ctx) && @info("Starting coupling loop")
     ## step in time
-    walltime = @elapsed for t in ((tspan[begin] + Δt_cpl):Δt_cpl:tspan[end])
+    ## Use ClimaComms.@elapsed to time the simulation on both CPU and GPU
+    walltime = ClimaComms.@elapsed comms_ctx.device for t in ((tspan[begin] + Δt_cpl):Δt_cpl:tspan[end])
```
juliasloan25, line 716: could wrap for loop in `begin` for readability

config/benchmark_configs/climaatmos_diagedmf.yml
```diff
@@ -0,0 +1,29 @@
+approximate_linear_solve_iters: 2
```
juliasloan25, line 1: add comment - duplicated from ClimaAtmos [gpu_aquaplanet_diagedmf.yml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/gpu_configs/gpu_aquaplanet_diagedmf.yml) in April 2024

experiments/AMIP/coupler_driver.jl
```diff
@@ -757,18 +797,46 @@ function solve_coupler!(cs)
...
 
 ## exit if running performance anaysis #hide
 if haskey(ENV, "CI_PERF_SKIP_COUPLED_RUN") #hide
     throw(:exit_profile_init) #hide
 end #hide
 
-## run the coupled simulation
-solve_coupler!(cs);
+## run the coupled simulation for one timestep to precompile everything before timing
+cs.tspan[2] = Δt_cpl
```
LenkaNovak, line 810: I presume `Base.precompile(solve_coupler!, (typeof(cs),))` doesn't work here?

LenkaNovak (commented):
Looks good, thank you! 🚀  I have a few comments and the parsed args need a fix. Then, presuming all CI runs are passing without change from the main, I think we can merge this.

experiments/AMIP/coupler_driver.jl
juliasloan25: I haven't tried it! I can give that a go

juliasloan25: It works :)

LenkaNovak: Although I've just noticed that the latest SYPD is still outputting the older estimate (including the precompilation), so we may need to revert previous version.

LenkaNovak (approved):
I'm approving this since the table works, but please do check the discrepancy between atmos/coupler allocations and that we're not seeing any behavioral changes on Buildkite once CI is fixed. Thanks for the great work, @juliasloan25 ! 🚀

experiments/AMIP/coupler_driver.jl
juliasloan25: I looked into this some more and chatted with Charlie (see conversation on slack [here](https://climate-machine.slack.com/archives/C0102PWRSGJ/p1715280052543379)), and it sounds like `precompile` only works if it can infer all the types, and even then still may not actually compile all the code. I opened an issue for us to fix the type inference and change to using `precompile` (https://github.com/CliMA/ClimaCoupler.jl/issues/775), but for now I'll just keep it as running for 2 timesteps before running the full simulation

juliasloan25:
> I'm approving this since the table works, but please do check the discrepancy between atmos/coupler allocations and that we're not seeing any behavioral changes on Buildkite once CI is fixed. Thanks for the great work, @juliasloan25 ! 🚀

I fixed CI and all the plots look the same as main 😄

I also changed the functions we're using to calculate allocations. Now there's one metric that's the same between CPU and GPU runs, which calculates the maximum CPU memory used over the course of the simulation (max RSS). The numbers don't look exactly like what I would expect (i.e. there are more allocations for atmos-only than coupled), but I think this is something we can easily change later on if we decide this isn't a reliable metric, and I don't want it to hold up this whole PR. I'll see if I can figure out what's going on with the allocation measurement tomorrow, but if nothing comes up I'm inclined to merge this as-is and investigate later on as we trigger more runs

PR #709, opened 2024-03-20
Longrun pipline fix

LenkaNovak:
[passing build](https://buildkite.com/clima/climacoupler-longruns/builds/501)

----


LenkaNovak:
@trontrytel after this PR you should be able to run the longruns without them tripping over the instantiate on Buildkite. Looking at the recent errors, you may need to update for some new atmos args (that are presumably on the main but not in the latest release). Lmk if it's still breaking after that.

trontrytel (approved):

PR #717, opened 2024-04-01
Truncate data to regrid only necessary dates

anastasia-popova:
## Purpose 

This PR aims to truncate the data sets that are regridded in the coupler driver to significantly reduce processing time and memory

## Content

----

### DIFF
--- a/artifacts/artifact_funcs.jl
+++ b/artifacts/artifact_funcs.jl
@@ -59,3 +59,25 @@ function pr_obs_data_path()
     )
     return AW.get_data_folder(pr_obs_data)
 end
+
+"""
+    artifact_data(datapath_full, name)
+
+Returns input dataset at datapath_full
+"""
+function artifact_data(datapath_full, name)
+    datafile_truncated = joinpath(datapath_full, string(name, ".nc"))
+    return datafile_truncated
+end
+
+"""
+    artifact_data(datapath_full, name, datapath_trunc, date0, time_start, time_end, comms_ctx)
+
+Truncates given data set, and constructs a new dataset containing only the dates needed and stores it in datapath_trunc
+"""
+function artifact_data(datapath_full, name, datapath_trunc, date0, time_start, time_end, comms_ctx)
+    datafile = joinpath(datapath_full, string(name, ".nc"))
+    datafile_truncated =
+        Regridder.truncate_dataset(datafile, name, datapath_trunc, date0, time_start, time_end, comms_ctx)
+    return datafile_truncated
+end
--- a/experiments/AMIP/coupler_driver.jl
+++ b/experiments/AMIP/coupler_driver.jl
@@ -168,10 +168,10 @@ original sources.
 =#
 
 include(joinpath(pkgdir(ClimaCoupler), "artifacts", "artifact_funcs.jl"))
-sst_data = joinpath(sst_dataset_path(), "sst.nc")
-sic_data = joinpath(sic_dataset_path(), "sic.nc")
-co2_data = joinpath(co2_dataset_path(), "mauna_loa_co2.nc")
-land_mask_data = joinpath(mask_dataset_path(), "seamask.nc")
+sst_data = artifact_data(sst_dataset_path(), "sst", REGRID_DIR, date0, t_start, t_end, comms_ctx)
+sic_data = artifact_data(sic_dataset_path(), "sic", REGRID_DIR, date0, t_start, t_end, comms_ctx)
+co2_data = artifact_data(co2_dataset_path(), "mauna_loa_co2", REGRID_DIR, date0, t_start, t_end, comms_ctx)
+land_mask_data = artifact_data(mask_dataset_path(), "seamask")
 
 #=
 ## Component Model Initialization
--- a/src/Regridder.jl
+++ b/src/Regridder.jl
@@ -636,5 +636,85 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
     return new_data, coords
 end
 
+"""
+    truncate_dataset(datafile, name, datapath_trunc, date0, time_start, time_end, comms_ctx)
+
+Truncates given data set, and constructs a new dataset containing only the dates that are used in the simulation
+"""
+function truncate_dataset(
+    datafile,
+    name,
+    datapath_trunc,
+    date0,
+    time_start,
+    time_end,
+    comms_ctx::ClimaComms.AbstractCommsContext,
+)
+    date_start = date0 + Dates.Second(time_start)
+    date_end = date0 + Dates.Second(time_start + time_end)
+
+    file_name = replace(string(name, "_truncated_data_", string(date_start), string(date_end), ".nc"), r":" => "")
+    datafile_truncated = joinpath(datapath_trunc, file_name)
+
+    if ClimaComms.iamroot(comms_ctx)
+        ds = NCDatasets.NCDataset(datafile, "r")
+        dates = ds["time"][:]
+
+        (start_id, end_id) = find_idx_bounding_dates(dates, date_start, date_end)
+
+        ds_truncated = NCDatasets.NCDataset(datafile_truncated, "c")
+        ds_truncated = NCDatasets.write(ds_truncated, NCDatasets.view(ds, time = start_id:end_id))
+
+        close(ds)
+        close(ds_truncated)
+
+        return datafile_truncated
+    end
+end
+
+"""
+    find_idx_bounding_dates(dates, date_start, date_end) 
+
+Returns the index range from dates that contains date_start to date_end
+"""
+function find_idx_bounding_dates(dates, date_start, date_end)
+    # if the simulation start date is before our first date in the dataset
+    # leave the beginning of the truncated dataset to be first date available
+    if date_start < dates[1]
+        start_id = 1
+        # if the simulation start date is after the last date in the dataset
+        # start the truncated dataset at its last possible date
+    elseif date_start > last(dates)
+        start_id = length(dates)
+        # if the simulation start date falls within the range of the dataset
+        # find the closest date to the start date and truncate there 
+    else
+        (~, start_id) = findmin(x -> abs(x - date_start), dates)
+        # if the closest date is after the start date, add one more date before 
+        if dates[start_id] > date_start
+            start_id = start_id - 1
+        end
+    end
+
+    # if the simulation end date is before our first date in the dataset
+    # truncate the end of the dataset to be the first date
+    if date_end < dates[1]
+        end_id = 1
+        # if the simulation end date is after the last date in the dataset
+        # leave the end of the dataset as is 
+    elseif date_end > last(dates)
+        end_id = length(dates)
+        # if the simulation end date falls within the range of the dataset
+        # find the closest date to the end date and truncate there 
+    else
+        (~, end_id) = findmin(x -> abs(x - date_end), dates)
+        # if the closest date is before the end date, add one more date after
+        if dates[end_id] < date_end
+            end_id = end_id + 1
+        end
+    end
+
+    return (; start_id, end_id)
+end
 
 end # Module
--- a/test/regridder_tests.jl
+++ b/test/regridder_tests.jl
@@ -6,6 +6,7 @@ import Dates
 import NCDatasets
 import ClimaComms
 import ClimaCore as CC
+import ClimaCoupler
 import ClimaCoupler: Interfacer, Regridder, TestHelper, TimeManager
 
 REGRID_DIR = @isdefined(REGRID_DIR) ? REGRID_DIR : joinpath("", "regrid_tmp/")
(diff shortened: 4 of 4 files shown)


src/Regridder.jl
```diff
@@ -633,5 +633,49 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
     return new_data, coords
 end
 
+"""
+    truncate_dataset(date0, timeStart, timeEnd, datafile)
+
+Truncates given data set to only regrid the dates that are used in the simulation
+"""
+function truncate_dataset(date0, timeStart, timeEnd, datafile, datapath)
```
juliasloan25, line 641: it would be good to use `time_start` and `time_end` instead of `timeStart` and `timeEnd` so we're consistent in using snake case throughout our code

src/Regridder.jl
```diff
@@ -633,5 +633,49 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
...
+    date_start = date0 + Dates.Second(timeStart)
+    date_end = date0 + Dates.Second(timeStart + timeEnd)
+
+    ds = NCDataset(datafile, "r")
+    dates = ds["time"][:]
+
+    if date_start < dates[1]
```
juliasloan25, line 648: could you add some comments in these if/else statements explaining the logic in words please?

src/Regridder.jl
```diff
@@ -633,5 +633,49 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
     return new_data, coords
 end
 
+"""
+    truncate_dataset(date0, timeStart, timeEnd, datafile)
+
+Truncates given data set to only regrid the dates that are used in the simulation
```
juliasloan25, line 639: It's more accurate to avoid saying anything about regridding in this function, since we're just truncating the dataset but not regridding yet. Maybe you could say something like `Truncates given dataset, and constructs a new dataset containing only the dates that are used in this simulation`

src/Regridder.jl
```diff
@@ -633,5 +633,52 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
...
+    else                                # if the simulation end date falls within the range of the dataset, we find the closest date to the end date and truncate there 
+        (~, end_id) = findmin(x -> abs(x - date_end), dates)
+        if dates[end_id] < date_end         # if the closest date is before the end date, we add one more date after it to the dataset 
+            end_id = end_id + 1
+        end
+    end
+
+    file_name = string("truncated_data_", string(date_start), string(date_end), ".nc")
+    datafile_truncated = joinpath(datapath, file_name)
+    truncated_ds = NCDataset(datafile_truncated, "c")
```
juliasloan25, line 672: This line is producing errors in buildkite: `ERROR: LoadError: NetCDF error: Permission denied (NetCDF error code: 13)`. From searching the error code online, it seems like this means either the file exists but we don't have permission to write to it, or it doesn't exist but we don't have permission to create new files in that directory. It would be good to figure out which of these two cases we're in. It sounds like you were able to run this locally, so maybe we don't have write permission for the folder on the cluster where we try to create the file in buildkite runs.

src/Regridder.jl
```diff
@@ -633,5 +633,52 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
...
+Truncates given data set, and constructs a new dataset containing only the dates that are used in the simulation
+"""
+function truncate_dataset(date0, time_start, time_end, datafile, datapath)
+    date_start = date0 + Dates.Second(time_start)
+    date_end = date0 + Dates.Second(time_start + time_end)
+
+    ds = NCDataset(datafile, "r")
+    dates = ds["time"][:]
+
+    if date_start < dates[1]        # if the simulation start date is before our first date in the dataset, we leave the beginning of the truncated dataset to be the first date available
```
juliasloan25, line 648: Could you add the comments in their own line above the if/else statements please? It's nice to keep the line length < ~80 characters when we can

test/regridder_tests.jl
```diff
@@ -314,3 +315,45 @@ for FT in (Float32, Float64)
...
+    sst_data_all = joinpath(sst_dataset_path(), "sst.nc")
+    ds = NCDataset(sst_data_all, "r")
+    dates = ds["time"][:]
+    first_date = dates[1]
+    last_date = last(dates)
+
+    # values for the truncations 
+    time_start = 0.0
+    time_end = 1.728e6
+    date0test = ["18690101", "18700101", "19790228", "20220301", "20230101"]
```
juliasloan25, line 331: it would be helpful to add a comment explaining which case each of these start dates tests

test/regridder_tests.jl
```diff
@@ -314,3 +315,45 @@ for FT in (Float32, Float64)
         end
     end
 end
+# test dataset truncation 
+@testset "test dataset truncation" begin
+    # Get the original dataset set up 
+    include(joinpath(pkgdir(ClimaCoupler), "artifacts", "artifact_funcs.jl"))
+    sst_data_all = joinpath(sst_dataset_path(), "sst.nc")
+    ds = NCDataset(sst_data_all, "r")
```
juliasloan25, line 323: this dataset needs to be closed when we're done with it

test/regridder_tests.jl
```diff
@@ -314,3 +315,45 @@ for FT in (Float32, Float64)
...
+    last_date = last(dates)
+
+    # values for the truncations 
+    time_start = 0.0
+    time_end = 1.728e6
+    date0test = ["18690101", "18700101", "19790228", "20220301", "20230101"]
+    for date in date0test
+        date0 = DateTime(date, dateformat"yyyymmdd")
+        sst_data = Regridder.truncate_dataset(date0, time_start, time_end, sst_data_all, sst_dataset_path())
+        nds = NCDataset(sst_data, "r")
```
juliasloan25, line 335: `nds` also needs to be closed

test/regridder_tests.jl
```diff
@@ -314,3 +315,54 @@ for FT in (Float32, Float64)
...
+
+    # values for the truncations 
+    time_start = 0.0
+    time_end = 1.728e6
+    # tests five cases: 
+    # 1. both start and end are earlier than in datafile 
+    # 2. start is early, end is within datafine 
+    # 3. both start and end within datafile 
+    # 4. start is in datafile end is not 
+    # 5. both start and end are later than in datafile
```
LenkaNovak, line 336: Nice tests! Could we also add a test that the sst data is correctly indexed (as well as the time)? I see from running this locally that sst has the correct dimensions, but it could be good to test that `original_sst[index_of_start_date] == truncated_sst[1]`

LenkaNovak (commented):
Great job, @anastasia-popova ! This is going to be super helpful. I just had a couple of small suggestions.

src/Regridder.jl
anastasia-popova: Would this overwrite all of the files to the same path, since they would all be in REGRID_DIR.truncated_data....nc?

test/regridder_tests.jl
anastasia-popova: Would this need my function to return the start index? It is all calculated in the function so I don't think there would be a way to know what the index_of_start_date is other than just looking at the index of the first value in trucated_sst defeating the purpose of the test

experiments/AMIP/coupler_driver.jl
```diff
@@ -176,11 +176,16 @@ original sources.
...
+sst_data_all = joinpath(sst_dataset_path(), "sst.nc")
+sic_data_all = joinpath(sic_dataset_path(), "sic.nc")
+co2_data_all = joinpath(co2_dataset_path(), "mauna_loa_co2.nc")
 land_mask_data = joinpath(mask_dataset_path(), "seamask.nc")
 
+# truncate data sets to needed dates 
+sst_data = Regridder.truncate_dataset(date0, t_start, t_end, sst_data_all, sst_dataset_path())
+sic_data = Regridder.truncate_dataset(date0, t_start, t_end, sic_data_all, sic_dataset_path())
+co2_data = Regridder.truncate_dataset(date0, t_start, t_end, co2_data_all, co2_dataset_path())
+
```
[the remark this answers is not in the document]
anastasia-popova: Should the tests stay where they are then? Or is there a new file that I should put them into, I did not see one for the artifacts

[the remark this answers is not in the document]
anastasia-popova: I tried to move this but then I have to add NCDatasets as a dependency in the atifact_funcs file. Should I do that? Or is it better to leave it in the regridder then?

test/regridder_tests.jl
```diff
@@ -314,3 +315,54 @@ for FT in (Float32, Float64)
...
+    date0test = ["18690101", "18700101", "19790228", "20220301", "20230101"]
+    for date in date0test
+        date0 = DateTime(date, dateformat"yyyymmdd")
+        sst_data = Regridder.truncate_dataset(date0, time_start, time_end, sst_data_all, sst_dataset_path())
```
juliasloan25, line 340: We should pass in a REGRID_DIR here instead of sst_dataset_path, so the windows tests don't try to access `runneradmin/` (as discussed in our meeting)

artifacts/artifact_funcs.jl
```diff
@@ -59,3 +60,28 @@ function pr_obs_data_path()
...
+"""
+function artifact_data(
+    datapath_full,
+    name,
+    datapath_trunc = nothing,
+    date0 = nothing,
+    time_start = nothing,
+    time_end = nothing,
+)
+    if datapath_trunc === nothing || date0 === nothing || time_start === nothing || time_end === nothing
```
juliasloan25, line 77: you can also dispatch off these types in the function signature using `::Nothing`, and have two separate methods instead of the if/else

test/regridder_tests.jl
```diff
@@ -314,3 +315,72 @@ for FT in (Float32, Float64)
...
+    # set up comms_ctx 
+    device = ClimaComms.device()
+    comms_ctx = ClimaComms.context(device)
+    ClimaComms.init(comms_ctx)
+
+    # make path for truncated datasets 
+    COUPLER_OUTPUT_DIR = joinpath("experiments/AMIP/output", "tests")
+    mkpath(COUPLER_OUTPUT_DIR)
+
+    REGRID_DIR = joinpath(COUPLER_OUTPUT_DIR, "regrid_tmp/")
```
juliasloan25, line 337: Including the "/" in the path might cause problems on windows since it uses the backslash "\" instead. I'm not sure if this would be handled by joinpath though

experiments/AMIP/coupler_driver.jl
```diff
@@ -177,10 +177,19 @@ original sources.
...
 include(joinpath(pkgdir(ClimaCoupler), "artifacts", "artifact_funcs.jl"))
-sst_data = joinpath(sst_dataset_path(), "sst.nc")
-sic_data = joinpath(sic_dataset_path(), "sic.nc")
-co2_data = joinpath(co2_dataset_path(), "mauna_loa_co2.nc")
-land_mask_data = joinpath(mask_dataset_path(), "seamask.nc")
+#sst_data_all = joinpath(sst_dataset_path(), "sst.nc")
+#sic_data_all = joinpath(sic_dataset_path(), "sic.nc")
+#co2_data_all = joinpath(co2_dataset_path(), "mauna_loa_co2.nc")
+#land_mask_data = joinpath(mask_dataset_path(), "seamask.nc")
+sst_data, = artifact_data(sst_dataset_path(), "sst", REGRID_DIR, date0, t_start, t_end, comms_ctx)
```
juliasloan25, line 184: extra comma after dataset variable ?

test/regridder_tests.jl
```diff
@@ -314,3 +315,74 @@ for FT in (Float32, Float64)
...
+    # 2. start is early, end is within datafine 
+    # 3. both start and end within datafile 
+    # 4. start is in datafile end is not 
+    # 5. both start and end are later than in datafile
+    date0test = ["18690101", "18700101", "19790228", "20220301", "20230101"]
+    for date in date0test
+        date0 = DateTime(date, dateformat"yyyymmdd")
+        (sst_data, start_id, end_id) =
+            Regridder.truncate_dataset(date0, time_start, time_end, sst_data_all, REGRID_DIR, "test", comms_ctx)
+        nds = NCDataset(sst_data, "r")
```
juliasloan25, line 356: could we call this `ds_truncated` please? I think it will be more clear

artifacts/artifact_funcs.jl
```diff
@@ -59,3 +60,26 @@ function pr_obs_data_path()
...
+function artifact_data(datapath_full, name, datapath_trunc, date0, time_start, time_end, comms_ctx)
+    datafile = joinpath(datapath_full, string(name, ".nc"))
+    println(date0) 
+    println(time_start)
+    println(time_end) 
+    println(datafile) 
+    println(datapath_trunc) 
+    println(name) 
+    println(comms_ctx)
+    datafile_truncated = Regridder.truncate_dataset(date0, time_start, time_end, datafile, datapath_trunc, name, comms_ctx)
```
juliasloan25, line 83: it would be nice for `artifact_data` and `truncate_dataset` to have arguments in the same order so we don't have to reorder them between calls to these functions

src/Regridder.jl
LenkaNovak: you should be able to set up a `run_name` specific path. The point was to write this somewhere outside the `.depot` directory.

artifacts/artifact_funcs.jl
```diff
@@ -59,3 +60,20 @@ function pr_obs_data_path()
     )
     return AW.get_data_folder(pr_obs_data)
 end
+
+"""
+    artifact_data(datapath_full, name; datapath_trunc, date0, time_start, time_end, comms_ctx)
```
juliasloan25, line 65: Maybe we should add docstrings for both methods of this function and explain what they do differently (i.e. one truncates the dataset and the other just returns the input dataset)

artifacts/artifact_funcs.jl
```diff
@@ -1,4 +1,5 @@
 import ArtifactWrappers as AW
+#using NCDatasets
```
juliasloan25, line 2: can delete this :)

src/Regridder.jl
```diff
@@ -633,5 +633,74 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
     return new_data, coords
 end
 
+"""
+    truncate_dataset(date0, timeStart, timeEnd, datafile, datapath, name)
```
juliasloan25, line 637: docstring doesn't match function signature

test/regridder_tests.jl
```diff
@@ -2,6 +2,7 @@
     Unit tests for ClimaCoupler Regridder module
 =#
 
+using ClimaCoupler
```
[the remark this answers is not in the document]
juliasloan25: we might need it to be able to call `pkgdir(ClimaCoupler)`, but it would be nice to use `import` instead of `using` to follow our new code loading convention

test/regridder_tests.jl
```diff
@@ -314,3 +315,67 @@ for FT in (Float32, Float64)
...
+    mkpath(COUPLER_OUTPUT_DIR)
+
+    REGRID_DIR = joinpath(COUPLER_OUTPUT_DIR, "regrid_tmp", "")
+    mkpath(REGRID_DIR)
+
+    # values for the truncations 
+    time_start = 0.0
+    time_end = 1.728e6
+    # tests five cases: 
+    # 1. both start and end are earlier than in datafile
```
juliasloan25, line 344: could you put these alongside the if/else cases before so it's more clear which lines correspond to which case? thank you!!

juliasloan25 (commented):
This is looking great! I just went through and left a final review, it's all minor stuff :)

juliasloan25 (approved):
This looks great to me! If Lenka doesn't have any additional comments, feel free to merge :)

src/Regridder.jl
```diff
@@ -633,5 +633,74 @@ function cgll2latlonz(field; DIR = "cgll2latlonz_dir", nlat = 360, nlon = 720, c
...
+    date_end = date0 + Dates.Second(time_start + time_end)
+
+    file_name = replace(string(name, "_truncated_data_", string(date_start), string(date_end), ".nc"), r":" => "")
+    datafile_truncated = joinpath(datapath_trunc, file_name)
+
+    if ClimaComms.iamroot(comms_ctx)
+        ds = NCDataset(datafile, "r")
+        dates = ds["time"][:]
+
+        # if the simulation start date is before our first date in the dataset
```
LenkaNovak, line 660: Could you please split these if else statements (L661 - L694) into a separate function (e.g. `find_bounding_dates`)? This way this can be used for different contexts, and it'll be easier to test.

test/regridder_tests.jl
```diff
@@ -314,3 +315,67 @@ for FT in (Float32, Float64)
...
+    # values for the truncations 
+    time_start = 0.0
+    time_end = 1.728e6
+    date0test = ["18690101", "18700101", "19790228", "20220301", "20230101"]
+    for date in date0test
+        date0 = DateTime(date, dateformat"yyyymmdd")
+        sst_data = Regridder.truncate_dataset(sst_data_all, "test", REGRID_DIR, date0, time_start, time_end, comms_ctx)
+        ds_truncated = NCDataset(sst_data, "r")
+        new_dates = ds_truncated["time"][:]
+
```
LenkaNovak, line 349: This is a great test for checking the dates in the truncated dataset, but it would still be good to add a test that checks that the variable in the truncated dataset is aligned with the new truncated dates. If we split `truncate_dataset` as suggested above, we can add something like:

```
all_data = ds["SST"][:, :, :]
new_data = ds_truncated["SST"][:, :, :]
first_index = find_bounding_dates(...)
@test new_data[:,:,1] ≈ all_data data[:,:,first_index]  
```

LenkaNovak (commented):
This looks very neat, thanks, @anastasia-popova. As I mentioned in the comments, it would be good to include a test that checks the truncation of the variable (in your case "SST"), as well as the dates. I propose splitting `truncate_dataset`, which allows us to separate the functionality to 1) find the date indices and 2) perform the actual truncation. This will clarify the functions' scope and make the separate functions more broadly useful.  :)

LenkaNovak (approved):
That's perfect, thank you, @anastasia-popova! No change in output as expected, and the initialization of our target GPU AMIP runs is reduced by 5mins (which can be 20-30% of the standard tests we use for development)! Great job!

PR #718, opened 2024-04-01
move SurfaceStub to its own file

### LINKED ISSUE #669: move SurfaceStub to its own file
Currently, SurfaceStub is defined in src/Interfacer.jl, and it has methods defined for it in Interfacer.jl and FieldExchanger.jl. We should separate it out into its own file, and include it from the Interfacer so that it's still part of that module.

juliasloan25:
## Purpose 
closes #669 

## To-do
- move SurfaceStub struct and functions from `src/Interfacer` to `src/surface_stub.jl`
- move additional SurfaceStub functions from `src/FieldExchanger` and `src/FluxCalculator` to `src/surface_stub.jl`
- update docs

## Questions
- Maybe we want to create a folder `src/Interfacer/` that contains `surface_stub.jl` and `Interfacer.jl`. On one hand, this might be nicer to organize `src/` and make it clear which module `surface_stub.jl` belongs to. On the other hand, it could be nice to maintain the flat structure we currently have in `src/`.

----

### DIFF
--- a/experiments/AMIP/coupler_driver.jl
+++ b/experiments/AMIP/coupler_driver.jl
@@ -60,7 +60,7 @@ using ClimaCoupler.FluxCalculator:
     MoninObukhovScheme,
     partitioned_turbulent_fluxes!,
     water_albedo_from_atmosphere!
-using ClimaCoupler.Interfacer: CoupledSimulation, SurfaceStub, get_field, update_field!
+using ClimaCoupler.Interfacer: CoupledSimulation, SurfaceStub, get_field, update_field!, step!
 using ClimaCoupler.Regridder
 using ClimaCoupler.Regridder: update_surface_fractions!, combine_surfaces!, binary_mask
 using ClimaCoupler.TimeManager:
--- a/docs/src/fieldexchanger.md
+++ b/docs/src/fieldexchanger.md
@@ -13,8 +13,6 @@ The component models are updated by broadcasting the coupler fields, via the `up
 
 If an `update_field!` function is not defined for a particular component model, it will be ignored.
 
-Each component model is also required to define its own `step!` and `reinit!` functions, otherwise this will be ignored.
-
 ## FieldExchanger API
 
 ```@docs
@@ -24,12 +22,4 @@ Each component model is also required to define its own `step!` and `reinit!` fu
     ClimaCoupler.FieldExchanger.update_sim!
     ClimaCoupler.FieldExchanger.reinit_model_sims!
     ClimaCoupler.FieldExchanger.step_model_sims!
-```
-
-
-## FieldExchanger Internal Functions
-
-```@docs
-    ClimaCoupler.FieldExchanger.step!
-    ClimaCoupler.FieldExchanger.reinit!
 ```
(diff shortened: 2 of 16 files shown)


experiments/AMIP/coupler_driver.jl
```diff
@@ -60,7 +60,7 @@ using ClimaCoupler.FluxCalculator:
     MoninObukhovScheme,
     partitioned_turbulent_fluxes!,
     water_albedo_from_atmosphere!
-using ClimaCoupler.Interfacer: CoupledSimulation, SurfaceStub, get_field, update_field!
+using ClimaCoupler.Interfacer: CoupledSimulation, SurfaceStub, get_field, update_field!, step!
```
LenkaNovak, line 63: Do we need `step!` here?

LenkaNovak (approved):
LGTM, thank you, @juliasloan25 ! 🚀

experiments/AMIP/coupler_driver.jl
juliasloan25: Yes, otherwise when we call `step!` on `ocean_sim` or `ice_sim` when they're `SurfaceStub`s, we don't have access to the `SurfaceStub` method of `step!`

LenkaNovak: Ahh, I see. It was defined locally in the FieldExchanger previously. Could we make `step!` and `reinit!` functions of the`Interfacer`, which throw an error if not defined? Then then we don't have to expose the `step!` function in the driver, where it is not directly used.

LenkaNovak (approved):

experiments/AMIP/coupler_driver.jl
juliasloan25: It looks like `step!` is used directly in the driver, during our initial component model exchange ([here](https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/AMIP/coupler_driver.jl#L625-L627)). `step!` and `reinit!` are functions of the Interfacer even though they're in `surface_stub.jl`, since that file is included by `Interfacer.jl`. I think we didn't explicitly import these functions in the driver before because they were already imported by files we include by the driver (e.g. atmos_init), but we use `step!` in the driver too.

I can add fallback methods to the Interfacer that error if they aren't extended

LenkaNovak: I meant `step!` of the stub isn't directly used. But yeah, it would be nice for all of the `step!` functions to be extensions of the Interfacer fallbacks. :) I guess we were shortcutting this before with the `step!` of the stub being defined inside the FieldExchanger module, where all the `step!`s are called within the src code. 👀

juliasloan25: I think `step!` of a stub is used in the driver, in line 626 that I linked, since `ocean_sim` is a `SurfaceStub` in AMIP sims. Maybe I'm missing something?

LenkaNovak: Nope, that was me. You're right, it is used. :)

juliasloan25: Okay! Good call with adding the fallback methods for these functions :) Are you okay with this PR as it is now?

LenkaNovak: Yeah, looks great. After the import fix, feel free to merge. Thanks, Julia!

PR #725, opened 2024-04-04
Use output dates from atmos.integrator

Sbozzolo:
This PR ensures that the leaderboard uses the output dates from atmos.


experiments/AMIP/coupler_driver.jl
```diff
@@ -858,20 +858,22 @@ if ClimaComms.iamroot(comms_ctx)
         # Compare against observations
         if t_end > 84600
             @info "Error against observations"
+            output_dates = cs.dates.date0[] .+ Second.(atmos_sim.integrator.sol.t)
```
LenkaNovak, line 861: It would be good to add a comment that `dt_save_to_sol` needs to be consistent with the required averaging. For most of our runs `dt_save_to_sol` is `10days` but the seasonal average using this method would require `1days` (for daily averaged data).

LenkaNovak (approved):
LGTM, thank you, @Sbozzolo! Just testing if two seasons produce different outputs [here](https://buildkite.com/clima/climacoupler-longruns/builds/557). Once checked, feel free to merge.

LenkaNovak:
Confirmed that MAM and DJF differ. 👍

juliasloan25:
@Sbozzolo can this be merged?

Sbozzolo:
Was cherry picked in https://github.com/CliMA/ClimaCoupler.jl/commit/79594b1ac40e6e123ff2aab8feca27bc9aaa0ff2

PR #743, opened 2024-04-18
Add tagbot

Sbozzolo:
ClimaCoupler 0.0.1 was released, but no tag was added. Tagbot will take care of automatically tagging everything

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,0 +1,14 @@
+name: TagBot
+on:
+  issue_comment:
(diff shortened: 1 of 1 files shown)


LenkaNovak (approved):
LGTM! Do we want to add a `changelog` as well like in [ClimaAtmos](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.github/workflows/TagBot.yml)? If so, we should probably unify/rethink our labeling system. What do you think, @juliasloan25 ?

PR #744, opened 2024-04-18
Target EDMF run on clima

### LINKED ISSUE #740: Use 4xA100 to achieve > 1SYPD for target EDMF AMIP
Since [gpu_amip_topo_target_diagedmf](https://github.com/CliMA/ClimaCoupler.jl/blob/main/.buildkite/longruns/pipeline.yml#L291) is our current target, we want to run it on the faster nodes: either clima's A100 or new-central's V100 / H100.

# Results
Running on Clima should be sufficient.

## clima A100:
- current driver, 4 GPUs [build](https://buildkite.com/clima/climacoupler-longruns/builds/630) SYPD ~ 1.51
- current driver  w/o coupler diagnostics (still with atmos diags), 4 GPUs [build](https://buildkite.com/clima/climacoupler-longruns/builds/631#018eef5f-81f0-4396-abf8-f9c93924a2b8) SYPD ~ 1.63

## new-central P100
- current driver, 1 GPU [build](https://buildkite.com/clima/climacoupler-longruns/builds/627#018eeed0-d717-4342-a92d-d85509ea6eb4) SYPD ~ 0.23
- current driver, 4 GPUs [build](https://buildkite.com/clima/climacoupler-longruns/builds/628) - memory timeout

## Note
- currently this run crashes (saturation adjustment) after ~90-140 days (non deterministic).

## Components in PR
- [x] move gpu_amip_topo_target_diagedmf to clima
- [x] increase t_end to 90d
- [x] spilt clima pipeline init from individual runs, so we can run the multiple runs on clima
- [x] remove concurrency block, which was slowing down the clima init, even though the depots are in different locations

LenkaNovak:
closes #740 

Corresponding [build](https://buildkite.com/clima/climacoupler-longruns/builds/635)

----

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -39,14 +39,8 @@ steps:
       JULIA_NUM_PRECOMPILE_TASKS: 8
       JULIA_MAX_NUM_PRECOMPILE_FILES: 50
 
-    concurrency: 1
-    concurrency_group: 'depot/climacoupler-longruns'
-
-  - wait
-
-  # DYAMOND AMIP: 1 day (convection resolving) - running on clima, so need to instantiate again
-  - label: "GPU AMIP SUPERFINE: dyamond_target"
-    key: "gpu_dyamond_target"
+  - label: "init clima :computer:"
+    key: "init_cpu_env_clima"
     command:
       - "echo $$JULIA_DEPOT_PATH"
 
@@ -61,15 +55,14 @@ steps:
       - "julia --project=artifacts -e 'using Pkg; Pkg.status()'"
       - "julia --project=artifacts artifacts/download_artifacts.jl"
 
-      - echo "--- Run simulation"
-      - "julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_dyamond_target.yml"
-    artifact_paths: "experiments/AMIP/output/amip/gpu_dyamond_target_artifacts/*"
     agents:
       queue: clima
-      slurm_mem: 20GB
-      slurm_gpus: 1
       modules: common
-    soft_fail: true
+    env:
+      JULIA_NUM_PRECOMPILE_TASKS: 8
+      JULIA_MAX_NUM_PRECOMPILE_FILES: 50
+
+  - wait
 
   - group: "Coupler integration and conservation tests"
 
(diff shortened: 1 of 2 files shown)


.buildkite/longruns/pipeline.yml
```diff
@@ -39,14 +39,8 @@ steps:
       JULIA_NUM_PRECOMPILE_TASKS: 8
       JULIA_MAX_NUM_PRECOMPILE_FILES: 50
 
-    concurrency: 1
```
juliasloan25, line 42: Is it okay to remove this? It seemed like it was needed in [#709](https://github.com/CliMA/ClimaCoupler.jl/pull/709)

.buildkite/longruns/pipeline.yml
```diff
@@ -289,13 +282,35 @@ steps:
...
-          slurm_gpus: 1
-          slurm_mem: 16GB
+          queue: clima
+          slurm_ntasks: 4
+          slurm_gpus: 4
+          slurm_mem: 20GB
+          modules: common
         soft_fail: true
 
+  # DYAMOND AMIP: 1 day (convection resolving) - running on clima, so need to instantiate again
```
juliasloan25, line 300: I think we can remove the "running on clima, so need to instantiate again" part of this comment

juliasloan25 (commented):
Looks good! I just had a couple questions

.buildkite/longruns/pipeline.yml
LenkaNovak: This depends on how often we want to run longruns. Sometimes (though not always) the pipeline init fails if two builds are triggered at the same time (i.e., the depot is updated by both at the same time). Setting up these concurrency groups forces one of the pipeline `init`s to wait, until the first one is finished. So in our normal CI, where we trigger concurrent builds every day, this is needed. In the longruns this can still be useful (on the offchance that we're triggering several runs), but this also forces the `new-central` and `clima` inits to wait for each other each time (adding 10 mins or more to each build). Since we don't trigger the longruns too frequently, I think it's more useful to be able to trigger both pipelines simultaneously. There might be a way to set up concurrency groups to separate clima and new-central environment inits, but not sure how complicated this would be. @Sbozzolo do you know?

Sbozzolo: Longruns are not using a dedicated depot and different builds will have their personal depot (the different compared to normal runs is that we don't set JULIA_DEPOT_PATH). So, I don't think that this concurrency group is needed.

.buildkite/longruns/pipeline.yml
```diff
@@ -289,13 +282,35 @@ steps:
...
+        env:
+          CLIMACOMMS_CONTEXT: "MPI"
         agents:
-          slurm_gpus: 1
-          slurm_mem: 16GB
+          queue: clima
+          slurm_ntasks: 4
+          slurm_gpus: 4
+          slurm_mem: 20GB
+          modules: common
```
Sbozzolo, line 297:
```suggestion
          queue: clima
          slurm_gpus_per_task: 1
          slurm_cpus_per_task: 4
          slurm_ntasks: 4
          slurm_mem: 20GB
          modules: common
```

.buildkite/longruns/pipeline.yml
```diff
@@ -289,13 +282,35 @@ steps:
 
       - label: "GPU AMIP FINE: new target amip: topo + diagedmf"
         key: "gpu_amip_target_topo_diagedmf"
-        command: "srun julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_amip_target_topo_diagedmf.yml"
+        command:
+          - echo "--- Run simulation"
+          - "srun julia --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_amip_target_topo_diagedmf.yml"
```
Sbozzolo, line 287:
```suggestion
          - "  srun --cpu-bind=threads --cpus-per-task=4 julia --threads=3 --color=yes --project=experiments/AMIP/ experiments/AMIP/coupler_driver.jl --config_file $CONFIG_PATH/gpu_amip_target_topo_diagedmf.yml"
```

Sbozzolo: It seems that having 4 threads per GPU while leaving one idle (= for OS communcation) leads to better performance/scaling

juliasloan25 (approved):
LGTM!

PR #745, opened 2024-04-20
Remove ternary operator with nothing

### LINKED ISSUE #719: `condition ? something : nothing` is not an optimal pattern
Often in `ClimaCoupler` one sees something like:

https://github.com/CliMA/ClimaCoupler.jl/blob/861d17ee8c3cea2f9fef76fd012666826f896288/src/Regridder.jl#L632

The intention here is to execute `rm` if `clean_dir` is `true`.

There's nothing fundamentally wrong with this pattern, but it obfuscates the intention a little bit. Why is a `nothing` there?

The preferred way to do this, as far as I can tell is
```diff
-clean_dir ? rm(DIR; recursive = true) : nothing
+clean_dir && rm(DIR; recursive = true) 
```

Sbozzolo:
Thanks to the magic of regular expressions:
```
find . -type f -name "*.jl" -exec sed -i -E 's/(.+)\\s?\\?\\s?(.*)\\s?:\\s?nothing/\1 \&\& \2/g' {} \;
```

Closes #719

### DIFF
--- a/experiments/AMIP/coupler_driver.jl
+++ b/experiments/AMIP/coupler_driver.jl
@@ -158,7 +158,7 @@ dir_paths = (; output = COUPLER_OUTPUT_DIR, artifacts = COUPLER_ARTIFACTS_DIR)
 
 if ClimaComms.iamroot(comms_ctx)
     @info(COUPLER_OUTPUT_DIR)
-    config_dict["print_config_dict"] ? @info(config_dict) : nothing
+    config_dict["print_config_dict"] && @info(config_dict)
 end
 
 #=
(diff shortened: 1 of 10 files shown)


LenkaNovak (approved):
LGTM, I'm also seeing a similar pattern [here](https://github.com/CliMA/ClimaCoupler.jl/blob/897515ef3f4df06459eef9ad6138857f857cd93d/src/PostProcessor.jl#L134) and [here](https://github.com/CliMA/ClimaCoupler.jl/blob/897515ef3f4df06459eef9ad6138857f857cd93d/experiments/AMIP/components/atmosphere/climaatmos.jl#L263). Could they also be replaced?

Sbozzolo:
They are different: you assigning the value of something.

Also, did you really mean to write `@error` or did you want to write `error` in the first of your links? They are different

LenkaNovak:
Oh yeah, just the first link then ;-) it just conditionally throws an error, so could we do:
`package.tag == RawData() && @error("Cannot perform horizontal slicing on raw model data. Specify :regrid")`?

LenkaNovak:
As for the `@error` vs `error`, I don't have a preference. `@error` does what we wanted it to at the time. Would `error` be preferable?

Sbozzolo:
Thanks, I misread the function.

I changed the first link too.

***@***.***` and `error` and just different. If you want to just print a
message, the first is the correct one.

PR #749, opened 2024-04-23
tests on GPU - plots

### LINKED ISSUE #750: tests on GPU - plots
We want to be able to produce plots in GPU runs. This was partially implemented in a branch that wasn't merged into main (see [here](https://github.com/CliMA/ClimaCoupler.jl/pull/614/files#diff-f01ec1107a04223a7409af03497ce828cdf644af9d40d69a8110831c2142cf27R51-R83)). It's accomplished by making a new space on the CPU, then copying fields from GPU to CPU. This will be slow and allocate a lot, but might still be worth it for debugging and postprocessing.

Maybe we should add a flag to optionally enable this in GPU runs, to avoid accidentally converting all the fields for plotting and slowing down GPU runs

part of https://github.com/CliMA/ClimaCoupler.jl/issues/571

juliasloan25:
## Purpose 
We want to be able to produce plots in GPU runs. The easiest way to do this is to create a CPU-based space and copy all of the data over to that, then plot them. This will allocate a lot, but will still be useful for debugging and postprocessing

ncep_visualizer.jl and amip_visualizer.jl already work on GPU.

closes #750 

Note: I think this same approach could be applied to the animations in `viz_explorer`, but I think it would involve creating a new space/field on CPU for each field at each timestep we want to plot. This will be really resource intensive, and I don't think it's worth doing until we have a cleaner way of transferring fields from GPU to CPU.

This was initially implemented in https://github.com/CliMA/ClimaCoupler.jl/pull/614 but not merged there

## To-do
- [x] plot_helper.jl
- [x] debug plots
- [x] ci_plots.jl

## To test
- [x] run debug_plots test
- [x] run driver with ci_plots = true (e.g. slabplanet_atmos_diags)

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -403,6 +403,14 @@ steps:
           slurm_mem: 20GB
           slurm_gpus: 1
 
+      - label: "GPU Slabplanet: extra atmos diagnostics"
+        key: "gpu_slabplanet_atmos_diags"
+        command: "julia --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/run_amip.jl --config_file $CONFIG_PATH/gpu_slabplanet_atmos_diags.yml"
+        artifact_paths: "experiments/ClimaEarth/output/slabplanet/gpu_slabplanet_atmos_diags_artifacts/*"
+        agents:
+          slurm_mem: 20GB
+          slurm_gpus: 1
+
       # GPU RUNS: AMIP
       - label: "GPU AMIP test: albedo from function"
         key: "gpu_amip_albedo_function"
--- a/config/ci_configs/gpu_slabplanet_atmos_diags.yml
+++ b/config/ci_configs/gpu_slabplanet_atmos_diags.yml
@@ -1,0 +1,22 @@
+anim: false
+apply_limiter: false
+ci_plots: true
+dt: "200secs"
+dt_cpl: 200
+dt_save_to_sol: "9days"
+energy_check: true
+h_elem: 4
+job_id: "gpu_slabplanet_atmos_diags"
+mode_name: "slabplanet"
+moist: "equil"
+mono_surface: true
+output_default_diagnostics: false
+precip_model: "0M"
+rad: "gray"
+run_name: "gpu_slabplanet_atmos_diags"
+t_end: "10days"
+vert_diff: "true"
+diagnostics:
+  - short_name: [mse, lr, ediff, ts, mass_strf, stab, vt, egr]
+    reduction_time: average
+    period: 1days
--- a/experiments/ClimaEarth/user_io/debug_plots.jl
+++ b/experiments/ClimaEarth/user_io/debug_plots.jl
@@ -1,9 +1,12 @@
 import Plots
 import Printf
+import ClimaComms
 import ClimaCore as CC
 import ClimaCorePlots
 import ClimaCoupler: Interfacer
 import ClimaAtmos as CA
+
+include("plot_helper.jl")
 
 # plotting functions for the coupled simulation
 """
@@ -12,7 +15,7 @@ import ClimaAtmos as CA
 Plot the fields of a coupled simulation and save plots to a directory.
 """
 function debug(cs::Interfacer.CoupledSimulation, dir = "debug", cs_fields_ref = nothing)
-    mkpath(dir)
+    isdir(dir) || mkpath(dir)
     @info "plotting debug in " * dir
     for sim in cs.model_sims
         debug(sim, dir)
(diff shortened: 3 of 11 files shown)


.buildkite/pipeline.yml
```diff
@@ -410,6 +410,14 @@ steps:
           slurm_mem: 20GB
           slurm_gpus: 1
 
+      - label: "GPU Slabplanet: extra atmos diagnostics"
```
juliasloan25, line 413: This run tests that `ci_plots.jl` can run on GPU

config/ci_configs/gpu_slabplanet_atmos_diags.yml
```diff
@@ -0,0 +1,22 @@
+anim: false
```
juliasloan25, line 1: This config is copied from `config/ci_configs/slabplanet_atmos_diags.yml`. Once we remove `job_id` from inside the config file, we won't need to have both the CPU and GPU versions (though we'll have to make sure we always use `anim: false` for GPU runs)

experiments/ClimaEarth/Manifest.toml
```diff
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.3"
 manifest_format = "2.0"
-project_hash = "9e4203990c8bcfa6c89ab13b8fa1ae6772387ffe"
+project_hash = "c4d564fc1d4d9f904168c6e8bfd0dc2be15a558e"
```
juliasloan25, line 5: I think this can be reverted ?

experiments/ClimaEarth/user_io/debug_plots.jl
```diff
@@ -48,9 +51,24 @@ function debug(cs_fields::NamedTuple, dir, cs_fields_ref = nothing)
...
+        if (field_name == :T_S) && (@isdefined debug_csf0)
+            push!(
+                all_plots,
+                Plots.plot(
+                    cpu_field .- debug_csf0.T_S,
+                    title = string(field_name) * "_anom" * print_extrema(field),
+                    color = :viridis,
+                ),
+            )
+        end
```
juliasloan25, line 71: I'm not sure where this part comes from - maybe it can be removed here?

experiments/ClimaEarth/user_io/plot_helper.jl
```diff
@@ -90,3 +92,66 @@ function Plots.contourf(
...
+        )
+        cpu_field = CC.Fields.ones(cpu_space)
+
+        parent(cpu_field) .= Array(parent(field))
+        return cpu_field
+    end
+end
+
+make_cpu_field(arr::Array) = arr
+make_cpu_field(arr::CUDA.CuArray) = Adapt.adapt(Array, arr)
```
juliasloan25, line 157: If we didn't support converting CuArrays to Arrays, we could remove the Adapt and CUDA dependencies, but I think it's good to include this since sometimes we plot fields and sometimes we plot arrays

Maybe this should be renamed to `make_cpu_object` instead of `make_cpu_field`

perf/Project.toml
```diff
@@ -1,7 +1,9 @@
 [deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
```
juliasloan25, line 2: Adapt and CUDA have to be in the perf environment now because they're required by `plot_helper.jl`, which is included by `amip_visualizer.jl` and `ncep_visualizer.jl`, which are included in `run_amip.jl` and then by `flame.jl` They also have to be in the test environment because `test/debug/debug_amip_plots.jl` includes `debug_plots.jl`, which includes `plot_helper.jl`

config/ci_configs/gpu_slabplanet_atmos_diags.yml
LenkaNovak: We can always dispatch the `plot_anim` on device and do nothing for GPU.

.buildkite/pipeline.yml
```diff
@@ -410,6 +410,14 @@ steps:
...
+        key: "gpu_slabplanet_atmos_diags"
```
LenkaNovak, line 414: Could be include the debug plot output in this run (or preferably in all runs as a default) at the end of the simulation? I think we need to modify 
```
    comms_ctx.device == ClimaComms.CPUSingleThreaded() &&
        comms_ctx isa ClimaComms.SingletonCommsContext &&
        debug(cs, joinpath(dir_paths.artifacts, "endstates_"))
```
at the end of the driver.

LenkaNovak (approved):
LGTM, thank you, @juliasloan25 ! It would be good to include the debug plot output in the CI runs, otherwise it's good to go.

config/ci_configs/gpu_slabplanet_atmos_diags.yml
juliasloan25: Good idea!

PR #754, opened 2024-04-24
Generalize atmos_init to dry and no radiation cases

### LINKED ISSUE #753: Generalize atmos_init to dry and no radiation cases
Currently, we are assuming that we are always running with a radiation and moisture models. As part of  #392 we need to be able to relax these assumptions. 

Components
- [x] add `hasradiation` and `hasmoisture` functions and update the relevant `get_field` functions
- [x] add buildkite driver with conservation checks (but asserts set to softfail due to division by zero for water conservation)
- [x] move `time_to_seconds` to `user_io/io_helpers.jl`, as it is not dependent on cli_options.  
- [x] renmed `get_atmos_config` to `get_atmos_config_dict` to differentiate between the config `Dict` and `AtmosConfig` objects
- [x] generalize flux calculation
- [x] generalize `dir_paths` so they can be easily manipulated for different experiments
- [x] update driver

# QA
- [x] ensure energy conservation [build](https://buildkite.com/clima/climacoupler-ci/builds/3510#018f1e4c-7953-48c9-826b-3dcae9ec00df)

Screen Shot 2024-04-29 at 11 30 02 AM

LenkaNovak:
closes #753

----


experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -21,6 +21,18 @@ struct ClimaAtmosSimulation{P, Y, D, I} <: Interfacer.AtmosModelSimulation
...
 
+hasradiation = function (integrator)
+    return !isnothing(integrator.p.atmos.radiation_mode)
+end
+
+hasmoisture = function (integrator)
+    return !(integrator.p.atmos.moisture_model isa CA.DryModel)
+end
+
+zero_scalar_half(integrator) = zeros(axes(CC.Fields.level(integrator.p.scratch.ᶠtemp_scalar, CC.Utilities.half)))
```
juliasloan25, line 32: Could you add comments explaining each of these? e.g. `_half` will return the first center (I think), `_one` will return the first face, and `_center` is the whole center space

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -35,27 +47,30 @@ function atmos_init(::Type{FT}, atmos_config_dict::Dict) where {FT}
         @warn("Running with ρe_int in coupled mode is not tested yet.", maxlog = 1)
     end
 
-    # define shorter references for long variable Interfacer.names to increase readability
+    # define shorter references for long variable Interfacer.names to increase readability, and set to zero
```
juliasloan25, line 50: oops - I think I changed "names" here to "Interfacer.names" when I was updating our import/using syntax... would you mind removing the "Interfacer."?

experiments/AMIP/components/atmosphere/climaatmos.jl
juliasloan25, line 32: Also, these functions will allocate since they're calling `zeros`. We should avoid this since they're called in `get_field`. Maybe we could have temp fields on each of these spaces and use them here?

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -100,48 +115,68 @@ function Interfacer.get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:radiative_
...
+        thermo_params = get_thermo_params(atmos_sim)
+        return integrator.u.c.ρe_tot .-
+               ᶜS_ρq_tot .* CA.e_tot_0M_precipitation_sources_helper.(Ref(thermo_params), ᶜts, ᶜΦ) .* integrator.dt
     else
-        return atmos_sim.integrator.u.c.ρe_tot
+        return integrator.u.c.ρe_tot
     end
 end
 
+# helpers for get_field extensions
```
juliasloan25, line 140: It might be helpful to add a comment that we have two cases for moisture (dry vs moist) and two for radiation (without or with), which is why we have two versions of each function here

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -162,10 +197,6 @@ function Interfacer.update_field!(atmos_sim::ClimaAtmosSimulation, ::Val{:co2},
     end
 end
 # extensions required by the Interfacer
-function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:surface_temperature}, csf)
-    sim.integrator.p.radiation.radiation_model.surface_temperature .= CA.RRTMGPI.field2array(csf.T_S)
-end
```
juliasloan25, line 167: Do we no longer require atmos models to provide `update_field!` for `surface_temperature`? If so, we should update the interfacer docs

juliasloan25 (commented):
Looks good! I especially like the directory setup change :) I noticed that we're using `COUPLER_OUTPUT_DIR/COUPLER_REGRID_DIR` in [test/regridder_tests.jl](https://github.com/CliMA/ClimaCoupler.jl/blob/main/test/regridder_tests.jl#L331-L335) too, would you mind changing that to use this new structure as well?

I was also concerned about some of the helper functions allocating new fields. If you want, we can brainstorm together how to avoid that

experiments/AMIP/components/atmosphere/climaatmos.jl
LenkaNovak: Atmos is now using the surface temperature in the surface thermo state to set `radiation_model.surface_temperature` internally. But now you mention it, this update is dependent on atmos calculating turbulent surface fluxes (basically one parameterization depends on the output of the other 😓), so without turbulent fluxes, radiation won't get its update. I.e. we should keep this. I will also comment on this in the code.

experiments/AMIP/components/atmosphere/climaatmos.jl
```diff
@@ -21,6 +21,31 @@ struct ClimaAtmosSimulation{P, Y, D, I} <: Interfacer.AtmosModelSimulation
...
+function hasradiation(integrator)
+    return !isnothing(integrator.p.atmos.radiation_mode)
+end
+
+function hasmoisture(integrator)
+    return !(integrator.p.atmos.moisture_model isa CA.DryModel)
+end
+
+# return zero fields at the first face model level (half), the first center model level (one), and all center model levels (center)
+function zero_scalar_half(integrator)
```
juliasloan25, line 33: I think we can delete these functions now

juliasloan25 (approved):
Looks good! Just one note about removing the `zero_scalar_...` functions, but after that I think this is good to go

PR #757, opened 2024-04-29
Add additional diagnostics to CAD extensions

### LINKED ISSUE #756: add additional diagnostics to CAD extensions
We need better diagnostics for large scale circulation features, which will help demonstrate the usefulness of the hierarchical modelling approach across different complexities.

This PR adds standard atmospheric diagnostics: 
- mass streamfunction for the zonal-mean integrated meridional mass flux (showing overturning circulations)
- static stability relating to the vertical thermal stratification of the atmosphere
- v*T, used for calculating large-scale eddies
- maximum Eady growth rate, which is a theoretical estimate of the growth rate of the most baroclinically unstable modes.

LenkaNovak:
closes #756 

----


experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
 
     end,
 )
+
+CAD.add_diagnostic_variable!(
+    short_name = "mass_strf",
+    long_name = "Meridional Mass Streamfunction = 2πa cos(lat) ∫_0^z ρv dz",
+    standard_name = "meridional_mass_streamfunction",
+    units = "kg m^-1 s^-1",
+    compute! = (out, state, cache, time) -> begin
```
juliasloan25, line 92: Could you add a comment explaining this computation please?

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
...
+"""
+    static_stability(cache)
+
+This function computes the static stability, or the squared root of the Brunt-Väisälä
+frequency N = √(g/θ dθ/dz) using the precomputed temperature field.
+"""
+function static_stability(cache)
+    thermo_params = CAP.thermodynamics_params(cache.params)
+    ᶜts = cache.precomputed.ᶜts
+    θ_virt = TD.virtual_pottemp.(thermo_params, ᶜts)
```
juliasloan25, line 168: Could you add small comments explaining what `θ_virt` and `Nsq` represent? I'm sure it's defined somewhere in atmos, but it would be nice to see it here without having to go digging

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
...
+    dθ_virtdz = cache.scratch.ᶜtemp_scalar
+    @. dθ_virtdz = CC.Geometry.WVector.(ᶜgradᵥ_.(ᶠinterp_(θ_virt))).components.data.:1
+    Nsq = cache.scratch.ᶜtemp_scalar_2
+    parent(Nsq) .= parent(CAP.grav(cache.params) ./ θ_virt .* dθ_virtdz)
```
juliasloan25, line 172: Do we need to use `parent` here? Can we just do these calculations directly on the fields? Same for line 175 below

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
...
+    units = "kg m^-1 s^-1",
+    compute! = (out, state, cache, time) -> begin
+        FT = eltype(state)
+        lat = CC.Fields.coordinate_field(axes(state.f)).lat
+        ρv = CC.Geometry.UVVector.(state.c.uₕ).components.data.:2 .* state.c.ρ
+        ᶠ∫_0_z_ρv = cache.scratch.ᶠtemp_scalar
+        CC.Operators.column_integral_indefinite!(ᶠ∫_0_z_ρv, ρv)
+        twoπa = FT(2π * 6371e3)
+        if isnothing(out)
+            return ᶠ∫_0_z_ρv .* twoπa .* cosd.(lat)
```
juliasloan25, line 100: Should we use `copy(ᶠ∫_0_z_ρv)` here since it's using `cache.scratch`?

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
...
+        end
+    end,
+)
+
+CAD.add_diagnostic_variable!(
+    short_name = "egr",
+    long_name = "(zonal) max. Eady growth rate = 0.31 f/N du/dz",
+    standard_name = "max_eady_growth_rate",
+    units = "s^-1",
+    compute! = (out, state, cache, time) -> begin
```
juliasloan25, line 143: A comment explaining the computation here would be helpful too

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
```diff
@@ -80,3 +83,95 @@ CAD.add_diagnostic_variable!(
...
+        Ω = CAP.Omega(cache.params)
+        φ = CC.Fields.coordinate_field(state.c).lat
+        u = CC.Geometry.UVVector.(state.c.uₕ).components.data.:1
+        du_dz = cache.scratch.ᶜtemp_scalar
+        @. du_dz = CC.Geometry.WVector.(ᶜgradᵥ_.(ᶠinterp_(u))).components.data.:1
+        egr = cache.scratch.ᶜtemp_scalar_2
+        parent(egr) .= parent(FT(0.31) .* 2Ω .* sind.(φ) ./ static_stability(cache) .* du_dz)
```
juliasloan25, line 150: Do we need to use `parent` here, or can we do the computation directly on fields?

juliasloan25 (commented):
Thank you Lenka! I left some comments

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
LenkaNovak: You're right. As long as all gradient vectors are converted to Fields with `.components.data.:1`, we should be able to avoid parent, except the conditional 
```
        mask(N²) = N² > 0
        N = cache.scratch.ᶜtemp_scalar_3
        parent(N) .= sqrt.((parent(N²) .* mask.(parent(N²))))
```
which doesn't work with Fields.

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
LenkaNovak: It should be ok, since after the multiplication the result becomes a new object. Would you agree, @dennisYatunin ?

dennisYatunin: Yeah, it looks like the value in `ᶠ∫_0_z_ρv` is used to set the diagnostic field immediately after it is computed with `column_integral_indefinite!`. Since there's no way its value can change before the diagnostic field is set, you don't need to allocate a copy of it.

LenkaNovak: Nice, thank you for clarifying!

experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl
juliasloan25: Why doesn't the conditional work with Fields?

dennisYatunin: Not completely sure, but it's probably because you're relying on automatic type promotion of `mask.(N²)` from `Bool`s to floating point numbers. The `Field` broadcasting machinery is still a bit brittle when it comes to inferring type promotions certain cases. I would suggest rewriting this in a simpler way to avoid using `parent`; e.g.,
```
@. N = sqrt(max(N², 0))
```

LenkaNovak: Great point! :)

juliasloan25 (approved):
LGTM!

PR #769, opened 2024-05-06
Reducing allocations through temp fields

### LINKED ISSUE #584: add temp field on boundary space to cache
It might be useful to add a field to the cache where we can put temporary fields on the boundary space. For example, a land mask over the surface. This could prevent allocating new fields (masks) at each timestep.

We already have two fields defined on the boundary space: `:temp1` and `:temp2` in `coupler_fields`. We should use these more throughout the code where we're using fields for calculations but don't necessarily need to keep the field itself. Atmos also stores some temp fields in its cache, which we can use in functions accessing the AtmosSimulation.

Note: `Fields.ones` and `Fields.zeros` allocate a new field. `parent` and `level` both return views and do not allocate. `axes` also doesn't allocate, but creating a new field on the space returned by it would

### places to use temp fields
- `import_combined_surface_fields`
- `swap_space` (double check that it allocates as it is now) - maybe we can change this to take in a space instead of calling `axes(field)`. Then we won't have to create a field on a space just to access the space, e.g. in `Regridder.land_fraction`
- `calculate_surface_air_density` for atmos
- `get_field(bucket_sim::BucketSimulation, ::Val{:energy})`
- `check_conservation!(cc::WaterConservationCheck, coupler_sim::Interfacer.CoupledSimulation)`
- anything else using Fields.ones or Fields.zeros

# Some useful findings
- pointers
`@btime p.precomputed.ᶜts > 450.522 ns (4 allocations: 4.66 KiB)`
- operators
`@btime @. ᶜgradᵥ_.(ᶠinterp_(u.c.ρ)) > 16.212 μs (12 allocations: 120.62 KiB)`
- ...

### notes: 
- import_combined_surface_fields!(): `@btime FieldExchanger.import_combined_surface_fields!(cs.fields, cs.model_sims, cs.boundary_space, cs.turbulent_fluxes) 122.500 μs (207 allocations: 119.08 KiB)` if all predefined: `@btime FieldExchanger.import_combined_surface_fields!(fields, model_sims, boundary_space, turbulent_fluxes) 121.709 μs (200 allocations: 102.61 KiB)`
    - using temp field reduced allocs from 209 -> 207 
    - `@btime Regridder.combine_surfaces!(combined_field, model_sims, Val(:surface_temperature)) 8.333 μs (28 allocations: 11.23 KiB)`
        - `@btime eltype(field1) 11.094 ns (0 allocations: 0 bytes)`
        - `@btime Interfacer.get_field(sim, Val(:area_fraction)) 54.272 ns (1 allocation: 32 bytes)`
        - `@btime nans_to_zero.(Interfacer.get_field(sim, field_name)) 1.467 μs (5 allocations: 12.39 KiB)` (1 from get_field()).
    - `@btime Regridder.dummmy_remap!(csf.T_S, combined_field) 290.280 ns (1 allocation: 32 bytes)` vs `@btime Regridder.dummmy_remap!(T_S, combined_field) 233.447 ns (0 allocations: 0 bytes`
        - `@btime parent(cs.fields.T_S) 338.352 ns (3 allocations: 3.73 KiB)`
- swap_space!(): `@btime Utilities.swap_space!(cs.fields.temp1, cs.fields.temp2)  759.009 ns (7 allocations: 7.50 KiB)` if fields preallocated: `@btime Utilities.swap_space!(field1, field2)  56.968 ns (1 allocation: 32 bytes)`

anastasia-popova:
## Purpose 

This PR aims to close #584 by reducing the allocations within the coupler. 

## To-do

- [x] import_combined_surface_fields (temp fields)
- [x] swap_space (updated function to take in space to swap to instead of field)
- [x] calculate_surface_air_density for atmos (swap_space! update to not need allocations)
- [x] get_field(bucket_sim::BucketSimulation, ::Val{:energy}) (e_per_area added to integrator fields)
- [x] check_conservation!(cc::WaterConservationCheck, coupler_sim::Interfacer.CoupledSimulation) (using new swap_space!)
- [x] anything else using Fields.ones or Fields.zeros (left mostly in init statements and tests)

## Content

----

### DIFF
--- a/experiments/AMIP/components/land/climaland_bucket.jl
+++ b/experiments/AMIP/components/land/climaland_bucket.jl
@@ -26,6 +26,20 @@ struct BucketSimulation{M, Y, D, I, A} <: Interfacer.LandModelSimulation
 end
 Interfacer.name(::BucketSimulation) = "BucketSimulation"
 
+"""
+    get_new_cache(p, Y, energy_check)
+
+Returns a new `p` with an updated field to store e_per_area if energy conservation 
+    checks are turned on. 
+"""
+function get_new_cache(p, Y, energy_check)
+    if energy_check
+        e_per_area_field = CC.Fields.zeros(axes(Y.bucket.W))
+        return merge(p, (; e_per_area = e_per_area_field))
+    else
+        return p
+    end
+end
 
 """
     bucket_init
@@ -46,6 +60,7 @@ function bucket_init(
     stepper = CTS.RK4(),
     date_ref::Dates.DateTime,
     t_start::Float64,
+    energy_check::Bool,
 ) where {FT}
     if config != "sphere"
         println(
@@ -93,6 +108,7 @@ function bucket_init(
 
     # Initial conditions with no moisture
     Y, p, coords = CL.initialize(model)
+    p = get_new_cache(p, Y, energy_check)
 
     # Get temperature anomaly function
     T_functions = Dict("aquaplanet" => temp_anomaly_aquaplanet, "amip" => temp_anomaly_amip)
@@ -148,7 +164,7 @@ Extension of Interfacer.get_field that provides the total energy contained in th
 """
 function Interfacer.get_field(bucket_sim::BucketSimulation, ::Val{:energy})
     # required by ConservationChecker
-    e_per_area = zeros(axes(bucket_sim.integrator.u.bucket.W))
+    e_per_area = bucket_sim.integrator.p.e_per_area .= 0
     CC.Operators.column_integral_definite!(
         e_per_area,
         bucket_sim.model.parameters.ρc_soil .* bucket_sim.integrator.u.bucket.T,
--- a/src/Utilities.jl
+++ b/src/Utilities.jl
@@ -12,16 +12,17 @@ import ClimaCore as CC
 export swap_space!
 
 """
-    swap_space!(field_out::CC.Fields.Field, field_in::CC.Fields.Field)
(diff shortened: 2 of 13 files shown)


juliasloan25 (commented):
Thank you Anastasia, this looks great! The `swap_space!` change will be really nice.

I sent you a message on slack about an idea to get rid of the allocations in the bucket `get_field` function - it would be nice to address that too since it's called at every timestep when we run with conservation checks (so all slabplanet runs). Let me know if you have any questions!

Also, in line 560 of src/Regridder.jl, it would be good to change the docstring example from`zeros(boundary_space)` to `temp_field`, so that we don't promote any poor coding practices in the future :)

experiments/AMIP/components/land/climaland_bucket.jl
```diff
@@ -27,6 +27,15 @@ end
 Interfacer.name(::BucketSimulation) = "BucketSimulation"
 
 
+function get_new_cache(p, Y, energy_check)
```
juliasloan25, line 30: It would be good to add a docstring to this function to explain what it's doing

src/Utilities.jl
```diff
@@ -12,16 +12,17 @@ import ClimaCore as CC
 export swap_space!
 
 """
-    swap_space!(field_out::CC.Fields.Field, field_in::CC.Fields.Field)
+    swap_space!(space_out, field_in::CC.Fields.Field)
```
juliasloan25, line 15: docstring doesn't match function signature

juliasloan25 (approved):
This looks great! I just had a coupler minor comments about docstrings, but once those are done feel free to merge :)

PR #770, opened 2024-05-08
Fix current longrun

### LINKED ISSUE #774: Fix current longrun
Currently, the Rayleigh sponge makes the longrun unstable after [140d](https://buildkite.com/clima/climacoupler-longruns/builds/679#018f4cc7-6aeb-4fd2-8d49-eed80bae7c97). Removing the sponge allows running the same config for [300d+](https://buildkite.com/clima/climacoupler-longruns/builds/680#018f501d-c80a-4cdd-a395-51faee06db92)). 

Note, in a recent PR we assumed that adding the sponge would not lead to instability, so we had only tested for a few tens of days. 

We remove the sponge for now, so we can maintain the 300d benchmark. Nevertheless, it is clear that we need to use some dissipation in the upper atmosphere to achieve a more realistic upper level circulation, so the correct dissipation configuration should be explored as part of a separate issue. (#773)

LenkaNovak:
Closes #774

[resulting build](https://buildkite.com/clima/climacoupler-longruns/builds/680)

----

### DIFF
--- a/config/longrun_configs/gpu_amip_target_topo_diagedmf.yml
+++ b/config/longrun_configs/gpu_amip_target_topo_diagedmf.yml
@@ -18,7 +18,7 @@ mode_name: "amip"
 mono_surface: false
 netcdf_output_at_levels: true
 output_default_diagnostics: true
-rayleigh_sponge: true
+rayleigh_sponge: false
 run_name: "gpu_amip_target_topo_diagedmf"
 start_date: "19790301"
 surface_setup: "PrescribedSurface"


juliasloan25:
In the [linked build](https://buildkite.com/clima/climacoupler-longruns/builds/680#_), the MPI AMIP FINE runs with topo and with topo + diagedmf fail after 3 days and 1 week respectively. Aren't these the runs that we expected to run for ~300 days?

juliasloan25 (approved):

Issue #788, opened 2024-05-16
Refactor ClimaEarth experiment dir


LenkaNovak:
We need to refactor the `experiments` directory to accommodate for new drivers that are needed for the hierarchies suite. 

## Components
- [x] rename coupler_driver.jl
- [x] add a new README that lists the experiments

## QA
- [x] no behavior changes on Buildkite

## Note
- eventually we may want to split the AMIP and slabplanet drivers, but this is not necessary for the hierarchies suite, so we will do this as a separate PR. https://github.com/CliMA/ClimaCoupler.jl/issues/787

PR #789, opened 2024-05-16
Refactor ClimaEarth experiment dir

LenkaNovak:
closes https://github.com/CliMA/ClimaCoupler.jl/issues/788

----


juliasloan25 (approved):
This looks good! I assume that we'll later on use this structure to have separate AMIP/slabplanet/etc drivers, as in the issue you mentioned here?

LenkaNovak:
> This looks good! I assume that we'll later on use this structure to have separate AMIP/slabplanet/etc drivers, as in the issue you mentioned here?

Definitely. It's just something that'll require more intrusive changes to the driver scripts and isn't super urgent right now. Thanks for the review!

PR #791, opened 2024-05-16
Move precipitation to artifact

### LINKED ISSUE #781: Move precipitation observation to ClimaArtifacts

Sbozzolo:
Closes #781


Sbozzolo:
I don't like this solution. The observed precipitation should not be an artifact in ClimaCoupler, it should belong to the experiment.

PR #799, opened 2024-05-17
Add RMSE leaderboard

Sbozzolo:
Looks like this

image

### DIFF
--- a/experiments/ClimaEarth/user_io/leaderboard/compare_with_obs.jl
+++ b/experiments/ClimaEarth/user_io/leaderboard/compare_with_obs.jl
@@ -1,15 +1,35 @@
 const OBS_DS = Dict()
 const SIM_DS_KWARGS = Dict()
+const OTHER_MODELS_RMSEs = Dict()
 
 function preprocess_pr_fn(data)
     # -1 kg/m/s2 -> 1 mm/day
     return data .* Float32(-86400)
 end
 
+Base.@kwdef struct RMSEs
+    model_name::String
+    ANN::Union{<:Real, ClimaAnalysis.OutputVar} = 0.0
+    DJF::Union{<:Real, ClimaAnalysis.OutputVar} = 0.0
+    JJA::Union{<:Real, ClimaAnalysis.OutputVar} = 0.0
+    MAM::Union{<:Real, ClimaAnalysis.OutputVar} = 0.0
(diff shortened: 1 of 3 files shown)


szy21 (approved):

experiments/ClimaEarth/user_io/leaderboard/compare_with_obs.jl
```diff
@@ -1,16 +1,35 @@
...
+    return val_or_rmse.([r.ANN, r.DJF, r.JJA, r.MAM, r.SON])
+end
+
 OBS_DS["pr"] =
     ObsDataSource(; path = joinpath(pr_obs_data_path(), "gpcp.precip.mon.mean.197901-202305.nc"), var_name = "precip")
 
 SIM_DS_KWARGS["pr"] = (; preprocess_data_fn = preprocess_pr_fn, new_units = "mm / day")
 
+OTHER_MODELS_RMSEs["pr"] = [RMSEs(; model_name = "AM4.0", ANN = 0.5, DJF = 1.0, JJA = 1.5, MAM = 0.5, SON = 1.0)]
+
```
LenkaNovak, line 32: Please can you clarify how exactly is this obtained from the source below? Is this dataset saved in the Box?

LenkaNovak (approved):
LGTM, thank you @Sbozzolo! I wondered how the AM4 was estimated (i.e. why does it need to be "eyeballed" as shown in the plot of the PR)? Also the legend obscures a part of the plot. Otherwise, it's good to go.

experiments/ClimaEarth/user_io/leaderboard/compare_with_obs.jl
Sbozzolo: It's a placeholder at the moment. 

I read the data off 
image

But this PR is mostly to set up the infrastructure to do this

LenkaNovak: Ahh ok, so it's using a different observed data source too. I think that's totally fine, but it would be good to include this in the comment. :)

szy21: It uses the same observation I think, just a different time period.

LenkaNovak: That would be also good to add, thanks!

PR #801, opened 2024-05-20
Add moist Held-Suarez run script 

### LINKED ISSUE #760: Hierarchies: Moist Held-Suarez driver
Following [Thatcher and Jablonowski 16](https://gmd.copernicus.org/articles/9/1263/2016/)

# components
- [x] add new driver
- [x] specify types of fluxes (we may want to use `Bulk` fluxes instead of `MOST` in the future)
- [x] fix dictionary passing to `atmos_init`

# QA
- [x] no behavioural chenges for previous AMIP and slabplanet runs in CI
- [x] resulting circulation compared with literature:
## Original paper result
Screen Shot 2024-05-22 at 3 34 26 PM

## Our model
Screen Shot 2024-05-22 at 3 22 56 PM
Screen Shot 2024-05-22 at 3 23 21 PM
Screen Shot 2024-05-22 at 3 30 18 PM

# Note
- for a better agreement in a future PR, we can try to increase the resolution, use bulk fluxes and use instantaneous precipitation

LenkaNovak:
closes #760

----


.buildkite/hierarchies/pipeline.yml
```diff
@@ -50,22 +50,18 @@ steps:
...
-          - "julia --project=artifacts -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-          - "julia --project=artifacts -e 'using Pkg; Pkg.precompile()'"
-          - "julia --project=artifacts -e 'using Pkg; Pkg.status()'"
-          - "julia --project=artifacts artifacts/download_artifacts.jl"
-
-          - echo "--- Run simulation"
           - "julia --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/run_dry_held_suarez.jl"
         artifact_paths:  "dry_held_suarez/dry_held_suarez/clima_atmos/*"
+        agents:
+          queue: clima
```
juliasloan25, line 56: If everything in this pipeline will run on clima, you can also change `queue` in line 2 of this file to `clima` instead of `new-central`, then everything will run on clima by default

.buildkite/pipeline.yml
```diff
@@ -387,6 +387,15 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":construction: Moist Held Suarez"
+        key: "moist_held_suarez"
+        command:
+          - sed 's/t_end = "1000days"/t_end = "1days"/' experiments/ClimaEarth/run_moist_held_suarez.jl > experiments/ClimaEarth/run_moist_held_suarez_modified.jl
```
juliasloan25, line 393: how are the modified drivers different from the unmodified ones?

docs/src/fluxcalculator.md
```diff
@@ -3,7 +3,7 @@
 This modules contains abstract types and functions to calculate surface fluxes in the coupler, or to call flux calculating functions from the component models.
 
 Fluxes over a heterogeneous surface (e.g., from a gridpoint where atmospheric cell is overlying both land and ocean) can be handled in two different ways:
-1. **Combined fluxes** (called with `CombinedStateFluxes()`)
+1. **Combined fluxes** (called with `CombinedStateFluxesMOST()`)
```
juliasloan25, line 6: Why did you add the `MOST` suffix? It would be nice to document what that refers to

experiments/ClimaEarth/run_dry_held_suarez.jl
```diff
@@ -13,15 +17,11 @@ import YAML
 ## ClimaESM packages
 import ClimaAtmos as CA
 using ClimaCore
```
juliasloan25, line 19: could change ClimaCore from `using` to `import` too

experiments/ClimaEarth/run_moist_held_suarez.jl
```diff
@@ -0,0 +1,354 @@
...
+        ),
+    ],
+    # held-suarez specific
+    "forcing" => "held_suarez", # Newtonian cooling already modified for moisture internally in Atmos
+    "precip_model" => "0M",
+    "moist" => "equil",
+    "prognostic_surface" => "PrescribedSurfaceTemperature",
+    "turb_flux_partition" => "CombinedStateFluxesMOST",
+)
+# TODO: may need to switch to Bulk fluxes
```
juliasloan25, line 120: why would we need to switch?

juliasloan25 (commented):
Looks good, thanks Lenka! I had some questions throughout

.buildkite/hierarchies/pipeline.yml
LenkaNovak: Good point, but I set it up on new-central intentionally as we may need to run some of the sims on it  once we add all of them (like we do with the longruns), so we don't take up all of the processes on clima. 👀

.buildkite/pipeline.yml
LenkaNovak: this is just changing the length of the simulation for CI (so it doesn't run for 1000d like in the hierarchy pipeline). There will be more elegant ways of doing it, but it works for our purposes here.

docs/src/fluxcalculator.md
LenkaNovak: This refers to the Monin Obukhov Similarity Theory (the acronym is common in the field and is used in Atmos). I can clarify this in the comments though. :)

experiments/ClimaEarth/run_dry_held_suarez.jl
LenkaNovak: Good catch!

experiments/ClimaEarth/run_moist_held_suarez.jl
LenkaNovak: Though less accurate, the bulk fluxes are what's used in the original paper. We use MOST, so we might want to switch to the bulk fluxes if we want to follow the exact setup as in the paper. I want to leave this decision for the refinement stage of this project though.

.buildkite/hierarchies/pipeline.yml
juliasloan25: sounds good!

.buildkite/pipeline.yml
juliasloan25: could we rename it from `_modified` to `_short` to be more descriptive?

docs/src/fluxcalculator.md
juliasloan25: Thank you!

experiments/ClimaEarth/run_moist_held_suarez.jl
juliasloan25: sounds good, thank you for clarifying!

juliasloan25 (approved):
Looks good! I think it might be nice to rename the `_modified` to `_short`, but other than that I think this is good to go

.buildkite/pipeline.yml
LenkaNovak: of course, good point!

LenkaNovak:
Awesome, thank you for the review! 🚀

PR #805, opened 2024-05-20
rm bycolumn

### LINKED ISSUE #736: tests on GPU - remove Fields.bycolumn
This is supposed to improve CPU performance when using threading. However, the gains are most likely marginal and this is not GPU compatible when other operators are being performed on the column (i.e. `colidx` is ignored on GPU). See #735 for an example of this problem and suggested solutions

## Tests on GPU
`bycolumn` is also preventing GPU compatibility in `flux_calculator_tests.jl` and `eisenman_seaice_tests.jl`. Removing it should make those GPU compatible.

# Solution
We can either remove the `bycolumn` loop completely, or replace this with `Operators.column_mapreduce!`. If it's not performant critical code, we could use `Spaces.column`

# QA
- make note of performance changes, both for CPU and GPU. Buildkite total job time suffices

juliasloan25:
## Purpose 
closes #736

## Results
- SYPD for partitioned turbulent fluxes run increased from ~2.5 to ~3.7 on CPU (not running long enough for consistent SYPD estimates, but definitely increased)
- partitioned turbulent fluxes now GPU-compatible
- results are identical to main

## Content: `bycolumn` removed from
- [x] `experiments/ClimaEarth/components/atmosphere/climaatmos_extra_diags.jl`
- [x] `src/FluxCalculator.jl.update_turbulent_fluxes` (used for partitioned turbulent fluxes)
  - and extended methods for component models
- [x] `get_field`, `update_field!` methods for component models
- [x] `experiments/ClimaEarth/components/ocean/eisenman_seaice.jl` `solve_eisenman_model`

----

### DIFF
--- a/src/FluxCalculator.jl
+++ b/src/FluxCalculator.jl
@@ -24,7 +24,7 @@ export PartitionedStateFluxes,
     BulkScheme,
     partitioned_turbulent_fluxes!,
     get_surface_params,
-    update_turbulent_fluxes_point!,
+    update_turbulent_fluxes!,
     water_albedo_from_atmosphere!
 
 """
@@ -107,7 +107,7 @@ end
     partitioned_turbulent_fluxes!(model_sims::NamedTuple, fields::NamedTuple, boundary_space::CC.Spaces.AbstractSpace, surface_scheme, thermo_params::TD.Parameters.ThermodynamicsParameters)
 
 The current setup calculates the aerodynamic fluxes in the coupler (assuming no regridding is needed)
-using adapter function `get_surface_fluxes_point!`, which calls `SurfaceFluxes.jl`. The coupler saves
+using adapter function `get_surface_fluxes!`, which calls `SurfaceFluxes.jl`. The coupler saves
 the area-weighted sums of the fluxes.
 
 Args:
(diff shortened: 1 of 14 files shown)


src/FluxCalculator.jl
```diff
@@ -341,15 +386,27 @@ function get_surface_fluxes_point!(inputs, surface_params::SF.Parameters.Surface
...
-    return (;
-        F_turb_ρτxz = F_turb_ρτxz,
-        F_turb_ρτyz = F_turb_ρτyz,
-        F_shf = F_shf,
-        F_lhf = F_lhf,
-        F_turb_moisture = F_turb_moisture,
-    )
+    F_turb_moisture = SF.evaporation(surface_params, inputs, outputs.Ch)
+
+    if !isnan(F_turb_ρτxz) && !isnan(F_turb_ρτyz) && !isnan(F_shf) && !isnan(F_lhf) && !isnan(F_turb_moisture)
```
juliasloan25, line 391: this works but could be cleaner

Sbozzolo (approved):

src/FluxCalculator.jl
```diff
@@ -248,61 +239,87 @@ Returns the inputs for the surface model simulation `sim`.
...
-        beta,                                   # beta
+        beta_fv,                                # beta
     )
+
+    # Put the result data layout back onto the surface space
+    return CC.Fields.Field(result, boundary_space)
 end
 function surface_inputs(::MoninObukhovScheme, input_args::NamedTuple)
-    (; thermo_state_sfc, thermo_state_int, uₕ_int, z_int, z_sfc, scheme_properties) = input_args
+    (; thermo_state_sfc, thermo_state_int, uₕ_int, z_int, z_sfc, scheme_properties, boundary_space) = input_args
```
charleskawczynski, line 275: Can we instead grab `boundary_space` from `axes(z_sfc)`? Or will that not work?

charleskawczynski (approved):
Looks great!

PR #808, opened 2024-05-21
Add  cloud-less aquaplanet driver

### LINKED ISSUE #761: Hierarchies: Cloud-less aquaplanet driver
Following [Frierson et al 2006](https://journals.ametsoc.org/view/journals/atsc/63/10/jas3753.1.xml)

also see [Frierson 2007](https://journals.ametsoc.org/view/journals/atsc/64/6/jas3935.1.xml) for reference. 

## Components
- [x] add driver
- [x] implement in CI
- [x] implement in hierarchy tests
- [x] enable time varying albedo in slab ocean

## QA
Screen Shot 2024-05-23 at 8 37 10 AM
Screen Shot 2024-05-23 at 8 37 36 AM
Screen Shot 2024-05-23 at 8 37 26 AM

- the above circulation is closer to the 10x moisture case in Frierson, with weak eddies and Ferrel cell, too strong tropical updrafts, lack of the cold point. Also it crashes after 180d. So this will require fine tuning. 
Screen Shot 2024-05-23 at 8 34 35 AM
Screen Shot 2024-05-23 at 8 34 30 AM
Screen Shot 2024-05-23 at 8 34 13 AM
Screen Shot 2024-05-23 at 8 34 07 AM

## Notes for fine tuning
- constant albedo crashes at 180d (with `FriersonDiffusion` and 160d standard diffusion), varying albedo after 100d. Debug plots show strong grid scale storms and not enough mixing, too moist near the surface, so consider 1) increasing the resolution, 2) increasing diffusion coefficient, 3) applying an upper tropo sponge, 4) revisit moisture removal, 5) add Betts Miller, 6) consider using parameters from [this Atmos run](https://github.com/CliMA/ClimaAtmos.jl/blob/eba9f0c29cfbbee9f22f1446e536261e68861333/config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_tvinsol_0M_slabocean.yml#L4).

LenkaNovak:
closes #761

----

### DIFF
--- a/experiments/ClimaEarth/components/ocean/slab_ocean.jl
+++ b/experiments/ClimaEarth/components/ocean/slab_ocean.jl
@@ -89,6 +89,8 @@ function ocean_init(
         thermo_params = thermo_params,
         # add dss_buffer to cache to avoid runtime dss allocation
         dss_buffer = CC.Spaces.create_dss_buffer(CC.Fields.zeros(space)),
+        α_direct = CC.Fields.ones(space) .* params.α,
+        α_diffuse = CC.Fields.ones(space) .* params.α,
     )
 
     ode_algo = CTS.ExplicitAlgorithm(stepper)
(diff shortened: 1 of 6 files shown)


experiments/ClimaEarth/components/ocean/slab_ocean.jl
```diff
@@ -137,6 +139,12 @@ end
 function Interfacer.update_field!(sim::SlabOceanSimulation, ::Val{:turbulent_energy_flux}, field)
     parent(sim.integrator.p.F_turb_energy) .= parent(field)
 end
+function Interfacer.update_field!(sim::SlabOceanSimulation, ::Val{:surface_direct_albedo}, field::CC.Fields.Field)
```
juliasloan25, line 142: we still don't require surface model simulations to extend `update_field!` for these two albedo fields, right? Just wondering if we should update the docs/explanation for FluxCalculator

experiments/ClimaEarth/run_cloudless_aquaplanet.jl
```diff
@@ -0,0 +1,362 @@
+# # Cloudless Aquaplanet
```
juliasloan25, line 1: Could you add a couple sentences at the top here explaining the setup for this simulation, and particularly how it's different from the other drivers? It would be good actually to have this for each of the driver scripts, especially as we're adding more

juliasloan25 (approved):

experiments/ClimaEarth/components/ocean/slab_ocean.jl
LenkaNovak: This is only required if the component in question has time varying albedo calculated outside of the model (e.g. the ocean albedo calculated in the coupler using atmos state). But it's a good idea to clarify this in the docs.

experiments/ClimaEarth/run_cloudless_aquaplanet.jl
LenkaNovak: Of course, good point!

PR #816, opened 2024-05-23
set job_id for interactive runs

juliasloan25:
## Purpose 
#804 checks that `job_id` is set via the command line, but we don't do that for interactive runs. This sets it to a default value for interactive debug runs


LenkaNovak (approved):
LGTM. Thank you @juliasloan25 !

Issue #778, opened 2024-05-13
table improvement: add atmos run without diag EDMF


juliasloan25:
We should add an additional row to the table for an atmos-only run without diagnostic EDMF. This will allow us to see the performance of ClimaAtmos without the dominating effect of diagnostic EDMF.

The goal is to have a config with as similar a setup as possible to [climaatmos_diagedmf.yml](https://github.com/CliMA/ClimaCoupler.jl/blob/main/config/benchmark_configs/climaatmos_diagedmf.yml), but without diagnostic edmf

- should this have `topography: Earth`? climaatmos_diagedmf.yml doesn't have that line either

PR #817, opened 2024-05-23
add atmos to table

juliasloan25:
## Purpose 
closes https://github.com/CliMA/ClimaCoupler.jl/issues/778

benchmark pipeline build: https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/164

## Content
- [x] add a ClimaAtmos run without diagnostic EDMF to the coupler table
- [x] replace "allocations" with "max RSS" to reflect the metric we're reporting now
- [x] add back the diagnostic EDMF toml that I accidentally removed in https://github.com/CliMA/ClimaCoupler.jl/pull/840/

----


config/benchmark_configs/climaatmos.yml
```diff
@@ -0,0 +1,21 @@
...
+dt: 120secs
+dt_cloud_fraction: 1hours
+dt_rad: 1hours
+dt_save_state_to_disk: "Inf"
+dt_save_to_sol: "Inf"
+dz_bottom: 30.0
+dz_top: 3000.0
+h_elem: 30
+moist: equil
+output_default_diagnostics: false
```
Sbozzolo, line 12: Maybe in a next PR we can switch the diagnostics on for our target jobs?  We need to include them if the table has to be representative of realistic runs.

experiments/ClimaEarth/user_io/benchmarks.jl
```diff
@@ -60,97 +64,121 @@ function argparse_settings()
...
     # Read in SYPD info
     sypd_file = open(joinpath(artifacts_dir, "sypd.txt"), "r")
     sypd = round(parse(Float64, read(sypd_file, String)), digits = 4)
 
-    # Read in allocations info
-    cpu_allocs_file = open(joinpath(artifacts_dir, "allocations_cpu.txt"), "r")
-    cpu_allocs = read(cpu_allocs_file, String)
+    # Read in max RSS info
+    cpu_max_rss_file = open(joinpath(artifacts_dir, "max_rss_cpu.txt"), "r")
+    cpu_max_rss = read(cpu_max_rss_file, String)
```
Sbozzolo, line 122: It looks like that this file is never properly closed

test/component_model_tests/climaatmos_standalone/atmos_driver.jl
```diff
@@ -73,7 +73,7 @@ if ClimaComms.iamroot(comms_ctx)
     sypd_filename = joinpath(output_dir, "sypd.txt")
     write(sypd_filename, "$sypd")
 
-    cpu_allocs_GB = Utilities.show_memory_usage(comms_ctx)
-    cpu_allocs_filename = joinpath(output_dir, "allocations_cpu.txt")
-    write(cpu_allocs_filename, cpu_allocs_GB)
+    cpu_max_rss_GB = Utilities.show_memory_usage(comms_ctx)
+    cpu_max_rss_filename = joinpath(output_dir, "max_rss_cpu.txt")
+    write(cpu_max_rss_filename, cpu_max_rss_GB)
```
Sbozzolo, line 78: It looks like that the file is never properly closed here too

Sbozzolo (approved):

experiments/ClimaEarth/user_io/benchmarks.jl
juliasloan25: good point

config/benchmark_configs/climaatmos.yml
juliasloan25: Our goal for the table was not to include diagnostics, since that isn’t included in the performance goals we have. If we’re interested in understanding how long “real runs” will take, I can also add a run with diagnostics

Sbozzolo: The performance goals should include IO: at the end, we want to look at the output.

Issue #764, opened 2024-05-01
Fix land albedo regridding


LenkaNovak:
This is the output of `CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)` when using `map_temporal`
<img width="376" alt="Screen Shot 2024-04-30 at 11 10 12 PM" src="https://github.com/CliMA/ClimaCoupler.jl/assets/38083334/8b5566c5-d9d7-4a7d-8399-96bbd223554d">

We checked the original nc file and this seems to be fine:
<img width="546" alt="Screen Shot 2024-04-30 at 10 45 52 PM" src="https://github.com/CliMA/ClimaCoupler.jl/assets/38083334/b5edf1ff-c9a5-49dc-a0f2-6f3f3b9cac0b">

This seems to be a problem in ClimaLand, but it would be good to check that this is still the case for the latest ClimaLand tag.

Sbozzolo:
A first look in ClimaUtilities shows no clear problems with TempestRemap:
![image](https://github.com/CliMA/ClimaCoupler.jl/assets/9167485/a42ad9bc-b046-46a2-bc31-f5ff48d2dbe9)

https://github.com/CliMA/ClimaUtilities.jl/issues/53

Sbozzolo:
This is from ClimaLand, `surface_albedo`, at the first iteration for our CI experiment
![image](https://github.com/CliMA/ClimaCoupler.jl/assets/9167485/ead271d5-26ad-4fe0-b93e-3b1f9702a99a)

juliasloan25:
I tried to recreate this from ClimaCoupler using the `slabplanet_albedo_temporal_map` configuration. The albedo looks fine after initialization, after 2 timesteps, and after 1 hour of coupled simulation. Here's the plot after one hour:
![albedo_after_1hour](https://github.com/CliMA/ClimaCoupler.jl/assets/51397186/b7703efb-7950-4895-ba35-dee127bb767f)

Was this seen in only the AMIP case before? If not, maybe this was fixed by updating to the most recent release of ClimaLand, which uses ClimaUtilities for regridding

juliasloan25:
I also tried the AMIP case (using the gpu_amip_albedo_temporal_map config, but running on CPU) and the albedo still looks fine after 1 hour:
[...]

PR #818, opened 2024-05-24
debug land albedo regridding

juliasloan25:
## Purpose 
see #764 description and thread for details

what we know:
- changing the `regridder_type` passed to the albedo constructor between InterpolationRegridder and TempestRegridder changes the quality of the plot produced (only tempest is messed up)
- tempest produces an unrealistic plot for both the temporal and static maps,
  - this is true whether we use a space constructed with `spacefillingcurve` in ClimaCoupler or not

## Plots
- AMIP and slabplanet look the same
- plot looks the same after init and after 1day simulation
- mostly using `slabplanet_temporal_map.yml` config with `h_elem: 16`, but tested with AMIP too
### Interpolations regridder
albedo_init_amip

### Tempest regridder
looks the same with or without `spacefillingcurve` in ClimaCoupler
slabplanet_tempest_albedo_before_step

#### Interpolation regridder
static_map_helem16_before_solve

#### Tempest regridder
tempest_static_map_before_solve

----

### DIFF
--- a/src/Regridder.jl
+++ b/src/Regridder.jl
@@ -169,11 +169,9 @@ function hdwrite_regridfile_rll_to_cgll(
     # If doesn't make sense to regrid with GPUs/MPI processes
     cpu_singleton_context = ClimaComms.SingletonCommsContext(ClimaComms.CPUSingleThreaded())
 
-    topology = CC.Topologies.Topology2D(
-        cpu_singleton_context,
-        CC.Spaces.topology(space2d).mesh,
-        CC.Topologies.spacefillingcurve(CC.Spaces.topology(space2d).mesh),
-    )
+    # Note: this topology gives us `space == space_undistributed` in the undistributed
+    # case (as desired), which wouldn't hold if we used `spacefillingcurve` here.
+    topology = CC.Topologies.Topology2D(cpu_singleton_context, CC.Spaces.topology(space2d).mesh)
     Nq = CC.Spaces.Quadratures.polynomial_degree(CC.Spaces.quadrature_style(space2d)) + 1
 
     space2d_undistributed = CC.Spaces.SpectralElementSpace2D(topology, CC.Spaces.Quadratures.GLL{Nq}())
--- a/test/TestHelper.jl
+++ b/test/TestHelper.jl
@@ -41,9 +41,9 @@ function create_space(
     mesh = CC.Meshes.EquiangularCubedSphere(domain, ne)
 
     if comms_ctx isa ClimaComms.SingletonCommsContext
-        topology = CC.Topologies.Topology2D(comms_ctx, mesh, CC.Topologies.spacefillingcurve(mesh))
+        topology = CC.Topologies.Topology2D(comms_ctx, mesh)
     else
-        topology = CC.Topologies.DistributedTopology2D(comms_ctx, mesh, CC.Topologies.spacefillingcurve(mesh))
+        topology = CC.Topologies.DistributedTopology2D(comms_ctx, mesh)
     end
 
     Nq = polynomial_degree + 1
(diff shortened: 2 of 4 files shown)


Sbozzolo:
The problem is in `boundary_space` somehow. I created a horzspace from scratch from the coupler and it works. What I noticed is that the `boundary_space` mesh has a different `elemorder` compared to a fresh horzspace

```
julia> boundary_space.grid.topology.elemorder
2400-element Vector{CartesianIndex{3}}:
 CartesianIndex(1, 1, 1)
.....

julia> horzspace.grid.topology.elemorder
CartesianIndices((20, 20, 6))
```

These are not equivalent.

Interestingly,
```
julia> extrema(parent(good) .- parent(bad))
(0.0, 0.0)
```
where `good` and `bad` are albedo obtained using `horzspace` or `boundary_space` respectively.

juliasloan25:
ClimaAtmos and ClimaCoupler both create spaces using a `spacefillingcurve` (e.g. [here](https://github.com/CliMA/ClimaCoupler.jl/blob/f2f6a6e2f3df584674bb152a7ab2b5f36bcdaf29/src/Regridder.jl#L172-L176) in ClimaCoupler.Regridder), whereas ClimaLand and ClimaUtilities both create spaces without a `spacefillingcurve` (e.g. [here](https://github.com/CliMA/ClimaUtilities.jl/blob/ea7f2b353d2179738ee967d7245ad18cbc112114/ext/TempestRegridderExt.jl#L340-L345) in ClimaUtilities.TempestRegridder). This is why the two spaces have different element orders.

I had to add a TempestRemap GPU workaround in ClimaLand to force the regridding to happen on the CPU (this involves creating a new CPU context and space), and I noticed that if I created the undistributed space with a `spacefillingcurve`, `space_distributed == space_undistributed` was not true. This is why ClimaLand (and ClimaUtilities) don't use the `spacefillingcurve`. We should unify this across all repos, but I'm not sure which approach we should take

juliasloan25:
> The problem is in `boundary_space` somehow. I created a horzspace from scratch from the coupler and it works. What I noticed is that the `boundary_space` mesh has a different `elemorder` compared to a fresh horzspace
> 
> ```
> julia> boundary_space.grid.topology.elemorder
> 2400-element Vector{CartesianIndex{3}}:
>  CartesianIndex(1, 1, 1)
> .....
> 
> julia> horzspace.grid.topology.elemorder
> CartesianIndices((20, 20, 6))
> ```
> 
> These are not equivalent.
> 
> Interestingly,
> 
> ```
> julia> extrema(parent(good) .- parent(bad))
> (0.0, 0.0)
> ```
> 
> where `good` and `bad` are albedo obtained using `horzspace` or `boundary_space` respectively.

ClimaAtmos and ClimaCoupler both create spaces using a `spacefillingcurve` (e.g. [here](https://github.com/CliMA/ClimaCoupler.jl/blob/f2f6a6e2f3df584674bb152a7ab2b5f36bcdaf29/src/Regridder.jl#L172-L176) in ClimaCoupler.Regridder), whereas ClimaLand and ClimaUtilities both create spaces without a `spacefillingcurve` (e.g. [here](https://github.com/CliMA/ClimaUtilities.jl/blob/ea7f2b353d2179738ee967d7245ad18cbc112114/ext/TempestRegridderExt.jl#L340-L345) in ClimaUtilities.TempestRegridder). This is why the two spaces have different element orders.

I had to add a TempestRemap GPU workaround in ClimaLand to force the regridding to happen on the CPU (this involves creating a new CPU context and space), and I noticed that if I created the undistributed space with a `spacefillingcurve`, `space_distributed == space_undistributed` was not true. This is why ClimaLand (and ClimaUtilities) don't use the `spacefillingcurve`. We should unify this across all repos, but I'm not sure which approach we should take

juliasloan25:
resolved - see https://github.com/CliMA/ClimaCoupler.jl/issues/764

Issue #713, opened 2024-03-26
`/groups/esm/slurm-buildkite/climacoupler-ci` folder


Sbozzolo:
`ClimaCoupler` is creating empty folders in `/groups/esm/slurm-buildkite/climacoupler-ci`.
The `slurm-buildkite` folder in `/groups/esm/` should be considerate private. It is used to manage our pipelines and should not be written to. Instead, you can create a subfolder in `/groups/esm`, or `/central/scratch/esm/slurm-buildkite/`

Sbozzolo:
Related to this, I think `flame_diff.jl` is not comparing differences. Something seems borken in the system to handle references

PR #842, opened 2024-06-07
change buildkite output folder

juliasloan25:
## Purpose 
closes https://github.com/CliMA/ClimaCoupler.jl/issues/713

Removes all references to the `/groups/esm/slurm-buildkite/climacoupler-ci ` folder.

## To Do
- [x] check that `/groups/esm/slurm-buildkite/climacoupler-ci` is still empty after CI runs


juliasloan25:
@Sbozzolo It looks like these were the only places this folder was used in the repo. Do you know if there's anything else that needs to be changed?

Sbozzolo:
More generally, the folder was being used to save some file across runs, but the folder on the cluster is empty, meaning that files are not being saved. So, something is not working as expected. We can still move forward with this PR.

Sbozzolo (approved):

PR #843, opened 2024-06-07
Update amip_target_topo_diagedmf.yml

szy21:
## Purpose 

This yaml file will be for the target simulation we want to show at the VESRI meeting. This PR updates the corresponding atmos yaml file, and removes a few redundant configs. We will add more configs (e.g. aerosols) to this file when they are ready.

## Content

----


Sbozzolo:
Do we want to maybe add every single config here? Including the one in default_config. This would isolate the run from possible changes and one has to "opt-in" changing configurations

Sbozzolo:
> Do we want to maybe add every single config here? Including the one in default_config. This would isolate the run from possible changes and one has to "opt-in" changing configurations

Oops, I though this was Atmos. We can do that in atmos (if we want to)

Sbozzolo (approved):

juliasloan25:
Thanks Zhaoyi! Should we change the runs that are output in the coupler table to use this setup too, since this is our target?

szy21:
> Thanks Zhaoyi! Should we change the runs that are output in the coupler table to use this setup too, since this is our target?

We would probably need to lower the resolution of this run in order to run a multi-year simulation for the VESRI meeting, so I think we can keep the one we have in the coupler table. Which is the config file you are using for the coupler table now?

PR #844, opened 2024-06-08
run benchmarks for 24h simulation time

juliasloan25:
## Purpose 
The benchmark runs have been set to run for 12 hours, but it looks like there's still some variability in SYPD at that point. Since we're iterating on the table output itself less now, it's probably worth it to increase the simulation length from 12 hours to 1 day.

The 2 original runs, AMIP and ClimaAtmos with diagnostic EDMF, seem to have a stable SYPD after 12 hours, but the newest ClimaAtmos without diagnostic EDMF seems to still have variability at 12 hours.

## Content

----

### DIFF
--- a/config/benchmark_configs/amip_diagedmf.yml
+++ b/config/benchmark_configs/amip_diagedmf.yml
@@ -11,6 +11,6 @@ mode_name: "amip"
 mono_surface: false
 monthly_checkpoint: false
 start_date: "19790301"
-t_end: "12hours"
+t_end: "1days"
 turb_flux_partition: "CombinedStateFluxesMOST"
 use_coupler_diagnostics: false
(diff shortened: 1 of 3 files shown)


Sbozzolo:
Could you please quantify the variability?

There shouldn't be much for a 12h run and we should understand why things fluctuate if they do.

Sbozzolo:
The difference is less than 2 %. It is very reasonable variability to have and I don't think we should be concerned with reducing it further. Such variability could even be due to the physical temperature of the device or with how processes are distributed by the operating system. 

If we want to have a really accurate measurament, we would have to run a statistically significant sample and take statistics out of it.

szy21:
In the atmos now we run most of the simulations for 12 hours for scaling. The difference I saw is also within 2% so I think it's ok?

juliasloan25:
Okay, I won't change it here then

PR #845, opened 2024-06-10
Add boxes in bias_leaderboard

Sbozzolo:
In this next iteration of the leaderboard, I am now using Makie [boxplot](https://docs.makie.org/v0.21/reference/plots/boxplot).

> Draw a Tukey style boxplot. The boxplot has 3 components:
> - a crossbar spanning the interquartile (IQR) range with a midline marking the median
> - an errorbar whose whiskers span range * iqr
> - points marking outliers, that is, data outside the whiskers

image

I tried plotting individual models to see how they would look like, but I couldn't get something that is properly readable. I can highlight specific models, if some are more useful/important than others.

### DIFF
--- a/experiments/ClimaEarth/user_io/leaderboard/cmip_rmse.jl
+++ b/experiments/ClimaEarth/user_io/leaderboard/cmip_rmse.jl
@@ -1,4 +1,4 @@
-import Statistics: median
+import Statistics: median, quantile
 
 const RMSE_FILE_PATHS = Dict()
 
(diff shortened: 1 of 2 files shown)


Sbozzolo:
image

experiments/ClimaEarth/user_io/leaderboard/cmip_rmse.jl
```diff
@@ -75,7 +75,25 @@ function RSME_stats(vecRMSEs)
...
+    quantile75 = RMSEs(;
+        model_name = "Quantile 0.75",
+        ANN = quantile(ANN, 0.75),
+        DJF = quantile(DJF, 0.75),
+        JJA = quantile(JJA, 0.75),
+        MAM = quantile(MAM, 0.75),
+        SON = quantile(SON, 0.75),
+    )
+
+    (; best_single_model = best_single_model(vecRMSEs), median_model, worst_model, best_model, quantile25, quantile75)
```
szy21, line 96: Do we still need worst_model, best_model, quantile25, quantile75?

szy21 (approved):
Thanks!

experiments/ClimaEarth/user_io/leaderboard/cmip_rmse.jl
Sbozzolo: We don't, but I don't think it hurts having them available?

szy21: Sounds good.

PR #855, opened 2024-06-13
Use divergent colormap in bias plots

Sbozzolo:
Automatically balanced around 0. 

image

Also fixes issue with order of months.

(Thanks @haakon-e )


szy21 (approved):
Do you know what would happen if the biases are positive or negative everywhere?

Sbozzolo:
> Do you know what would happen if the biases are positive or negative everywhere?

I think it should still include 0 as a point, but the levels are going to be onesided.

PR #857, opened 2024-06-14
Fix discard spinup in leaderboad and add best/worst single models

Sbozzolo:
The routine that was discarding the spinup period had a flipped sign which resulted in everything but the spinup being discard. Now that's fixed and the first 6 months are discarded. In fixing this, I added the model that minimize/maximise errors across all the metrics (normalized to the median RMSE). 

I also started decoupling the code to produce the leaderboard from the driver so that it can be run independently. Now, it still lives in `run_amip.jl` but if ones changes these two lines
```
            diagnostics_folder_path = atmos_sim.integrator.p.output_dir
            leaderboard_base_path = dir_paths.artifacts
```
with the correct folders, the code will produce the leaderboard for the data that is found.

This is what the leaderboard looks now (540 days run, 180 discarded)
bias_leaderboard


szy21 (approved):
Thank you!

PR #878, opened 2024-06-27
update driver docs

juliasloan25:
## Purpose 
Improve the clarity and accuracy of documentation in the AMIP driver.

### DIFF
--- a/experiments/ClimaEarth/run_amip.jl
+++ b/experiments/ClimaEarth/run_amip.jl
@@ -191,13 +191,16 @@ Utilities.show_memory_usage(comms_ctx)
 atmos_sim = atmos_init(atmos_config_object);
 Utilities.show_memory_usage(comms_ctx)
 
-thermo_params = get_thermo_params(atmos_sim) # TODO: this should be shared by all models #610
+thermo_params = get_thermo_params(atmos_sim) # TODO: this should be shared by all models #342
 
 #=
 ### Boundary Space
 We use a common `Space` for all global surfaces. This enables the MPI processes to operate on the same columns in both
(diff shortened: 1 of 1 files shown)


experiments/ClimaEarth/run_amip.jl
```diff
@@ -993,19 +1032,19 @@ if ClimaComms.iamroot(comms_ctx)
...
-    !(comms_ctx isa ClimaComms.MPICommsContext) && debug(cs, dir_paths.artifacts)
+    !CA.is_distributed(comms_ctx) && debug(cs, dir_paths.artifacts)
 
-    if isinteractive()
-        ## clean up for interactive runs, retain all output otherwise
-        rm(dir_paths.output; recursive = true, force = true)
-    end
+    if isinteractive() #hide
+        ## clean up for interactive runs, retain all output otherwise #hide
+        rm(dir_paths.output; recursive = true, force = true) #hide
```
Sbozzolo, line 1047: Just a reminder that we should not do this (it's very dangerous)

Sbozzolo (approved):

experiments/ClimaEarth/run_amip.jl
juliasloan25: Why is this very dangerous? It's only being done for interactive runs

Sbozzolo: If you set `COUPLER_OUTPUT_DIR` to be your home directory, this line would remove all your files. Or, if you set `COUPLER_OUTPUT_DIR = "."`, this line will remove your working folder.

We should never remove files and folders we haven't created.

PR #879, opened 2024-06-28
Clean up long-run pipeline, info statements. 

akshaysridhar:
- Simplifies 'info' statements to only print on root proc
- Fixes slabocean, topo+edmf longrun config files following this [broken build](https://buildkite.com/clima/climacoupler-longruns/builds/757). These bring these config files in line with Atmos defaults while we investigate potential issues with our SSPKnoth timestepper.  
	```modified:   experiments/ClimaEarth/components/atmosphere/climaatmos.jl
	modified:   src/Utilities.jl
	modified:   .buildkite/longruns/pipeline.yml
	modified:   config/longrun_configs/amip_target_topo.yml
	modified:   config/longrun_configs/slabplanet_aqua_coupler_sf_evolve_ocn.yml
	```


akshaysridhar:
There is a pending issue following [build 699](https://buildkite.com/clima/climacoupler-longruns/builds/699), in which the same configuration fails on CPU but passes on GPU devices (possibly due to recent `bycolumn` changes in ClimaAtmos / Coupler)

config/longrun_configs/amip_target_topo.yml
```diff
@@ -9,11 +9,16 @@ dt_cpl: 100
...
 dt_save_to_sol: "10days"
+ode_algo: "ARS343"
 energy_check: false
 hourly_checkpoint: true
 land_albedo_type: "map_temporal"
 mode_name: "amip"
 mono_surface: false
+hyperdiffusion: "ClimaHyperdiffusion"
+vorticity_hyperdiffusion_coefficient: 1.0
+scalar_hyperdiffusion_coefficient: 4.0
```
juliasloan25, line 20: these are also specified in the atmos config: https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_0M_earth.yml

It's okay to specify them here too (it will overwrite the atmos options), but it does make the coupler config longer/more complex

config/longrun_configs/slabplanet_aqua_coupler_sf_evolve_ocn.yml
```diff
@@ -9,6 +9,7 @@ h_elem: 6
 land_albedo_type: "function"
 mode_name: "slabplanet_aqua"
 moist: "equil"
+ode_algo: "ARS343"
```
juliasloan25, line 12: I thought we only want to use ARS343 for runs with diagnostic EDMF. Do we want it for this run too?

experiments/ClimaEarth/run_amip.jl
```diff
@@ -108,6 +108,9 @@ job_id = parsed_args["job_id"]
 config_dict = YAML.load_file(parsed_args["config_file"])
 config_dict = merge(parsed_args, config_dict)
 
+comms_ctx = Utilities.get_comms_context(parsed_args)
```
juliasloan25, line 111: should we change the @info calls in this file to run on the root too?

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -305,20 +305,28 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String)
     atmos_config_file = coupler_dict["atmos_config_file"]
     atmos_config_repo = coupler_dict["atmos_config_repo"]
     # override default or specified configs with coupler arguments, and set the correct atmos config_file
+    comms_ctx = ClimaComms.context()
+    @info atmos_config_repo
     if atmos_config_repo == "ClimaCoupler"
         @assert !isnothing(atmos_config_file) "Must specify `atmos_config_file` within ClimaCoupler."
-        @info "Using Atmos configuration from ClimaCoupler in $atmos_config_file"
+        if ClimaComms.iamroot(comms_ctx)
```
juliasloan25, line 312: it would be nice to have a wrapper for this - maybe a function `info_on_root` that takes in a context and a message, and does this check and calls `@info` internally

juliasloan25:
We kept the separate GPU and CPU configs for this run only because we want to run it on GPU for 300 days, but running on CPU for 300 days will take a very long time (30 days already takes 7 hours). We kept the CPU config runs at 30 days to avoid wasting compute resources

juliasloan25:
> There is a pending issue following [build 699](https://buildkite.com/clima/climacoupler-longruns/builds/699), in which the same configuration fails on CPU but passes on GPU devices (possibly due to recent `bycolumn` changes in ClimaAtmos / Coupler)

This configuration has been failing on CPU and passing on GPU for a few weeks (e.g. see [this build](https://buildkite.com/clima/climacoupler-longruns/builds/727) from two weekends ago), so it predates at least the bycolumn changes that were made in the coupler

akshaysridhar:
> We kept the separate GPU and CPU configs for this run only because we want to run it on GPU for 300 days, but running on CPU for 300 days will take a very long time (30 days already takes 7 hours). We kept the CPU config runs at 30 days to avoid wasting compute resources

Ah noted, the SYPD values are  0.3 or 1.4 for CPU/ GPU respectively, I'll revert this change.

config/longrun_configs/slabplanet_aqua_coupler_sf_evolve_ocn.yml
akshaysridhar: We're looking into possible issues when there is precipitation, with the `SSPKnoth` scheme, so I have reset this `ode_algo` to `ARS343` in the meantime.

szy21 (approved):
Just one comment on the timestepper change. Otherwise it looks good, thanks!

szy21:
Also, Could we use vert_diff: "FriersonDiffusion" instead of vert_diff: "true"?

PR #880, opened 2024-07-01
Update climaatmos.yml in benchmark

szy21:
## Purpose 

Updates atmos without edmf benchmark to a more physical setup, while keeping it as a useful comparison with the atmos with edmf benchmark.

## Content

----


Sbozzolo (approved):

szy21:
I will update this config so it is more similar to the one in atmos and closer to what we want.

szy21:
I updated the config. Now this benchmark may be a bit slower, but it is a more fair comparison with the atmos with EDMF. I will keep an eye on the stability of this benchmark.

PR #887, opened 2024-07-03
remove ediff diagnostics

szy21:
## Purpose 

Eddy diffusivity is in atmos diagnostics (`edt` and `evu`) so we don't need extra diagnostics here. The atmos diagnostics are also more correct as we get them directly from the cache.

## Content

----


szy21:
cc @LenkaNovak in case you need those diagnostics for plots

PR #919, opened 2024-08-06
Fix a bug in P_snow in bucket

szy21:
## Purpose 

ClimaLand uses snow water equivalent, so the mass flux should be divided by the density of liquid water instead of ice.

## Content

----


kmdeck (approved):
looks good, thank you!

PR #921, opened 2024-08-07
Add support for plotting limits in leaderboard

Sbozzolo:
Looks like this

image


szy21 (approved):
Thanks! I will add the extrema for other variables.

szy21:
Looks like this:
bias_total

PR #923, opened 2024-08-08
Use Periodic instead of manually shifting lons

Sbozzolo:
`Periodic()` from 0 to 360 is equivalent to manually shifting the coordinates from -180 to 180, but it is less code.


szy21:
looks great
bias_total

szy21 (approved):

PR #926, opened 2024-08-15
Add option to shift ObsDataSources to end of month

Sbozzolo:
This helps with comparing with our model because observations are typically defined on the 15th, but our model is defined at the end of the month. Since we always find the closest data available, this can lead to comparisons across months.


szy21:
merged with https://github.com/CliMA/ClimaCoupler.jl/pull/927

PR #937, opened 2024-08-22
Use ClimaUtilities `Inputs` in `run_amip`

### LINKED ISSUE #916: Add an option to add a constant to prescribed SST
This will be useful for e.g. AMIP4K run.

Sbozzolo:
ClimaUtilities comes with a rich array of tools to read input data. Among these are the `SpaceVaryingInput` and the `TimeVaryingInput` to read data from NetCDF files (and more). These functions are more efficient, versatile, and come with various interpolation options.

This commit moves AMIP to using those. In the process, it moves from using ClimaCoreTempestRemap to Interpolations.

Closes #916


szy21 (approved):
Thanks! I'm not entirely sure what interpolation_daily is doing (I assume it means getting updated once every day), and why land fraction is needed for some of the previous interpolation. But the code change looks good to me.

Sbozzolo:
Yes, interpolate daily updates the data once a day, otherwise data is read once a month.

PR #940, opened 2024-08-24
Clean up some environments

Sbozzolo:
This commits removes some of the unused dependencies in various environments.

It also removes the direct dependency on CUDA and MPI (as for ClimaComms 0.6)


experiments/ClimaCore/sea_breeze/run.jl
```diff
@@ -29,9 +28,6 @@ CC.Operators.allow_mismatched_fd_spaces() = true #hide
 push!(LOAD_PATH, joinpath(@__DIR__, "..", "..", "..")) #hide
 const FT = Float64 #hide
 
-# Set random seed for reproducibility
-Random.seed!(1234)
```
juliasloan25, line 33: @Sbozzolo we need this because the atmos is initialized randomly [here](https://github.com/CliMA/ClimaCoupler.jl/blob/90a016e6fa45d8db63794d01e726e0635115ea40/experiments/ClimaCore/sea_breeze/atmos_rhs.jl#L169). Towards the beginning of the year, we weren't setting the random seeds and ran into problems because the simulation wasn't reproducible

Sbozzolo: Thanks, I missed it!

PR #958, opened 2024-09-20
delete `cgll2latlonz` function

juliasloan25:
## Purpose 
We have a function `Regridder.cgll2latlonz` that is a wrapper around wrappers around ClimaCoreTempestRemap functions. We have no use for it in source, and it is only called in a single test. This commit deletes the function, and calls the CCTR wrappers directly from the test where `cgll2latlonz` was called before.

closes https://github.com/CliMA/ClimaCoupler.jl/issues/833

## Content
- [x] delete function
- [x] update docs
- [x] update test to call CCTR wrappers directly

----

### DIFF
--- a/test/regridder_tests.jl
+++ b/test/regridder_tests.jl
@@ -296,7 +296,11 @@ for FT in (Float32, Float64)
             Base.close(hdfreader)
 
             # regrid back to lat-lon
-            T_rll, _ = Regridder.cgll2latlonz(T_cgll)
+            datafile_latlon = joinpath(REGRID_DIR, "remapped_latlon.nc")
+            nlat = 360
+            nlon = 720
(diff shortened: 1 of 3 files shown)


test/regridder_tests.jl
```diff
@@ -296,7 +296,11 @@ for FT in (Float32, Float64)
             Base.close(hdfreader)
 
             # regrid back to lat-lon
-            T_rll, _ = Regridder.cgll2latlonz(T_cgll)
+            datafile_latlon = REGRID_DIR * "/remapped_latlon.nc"
```
Sbozzolo, line 299:
```suggestion
            datafile_latlon = joinpath(REGRID_DIR, "remapped_latlon.nc")
```

Sbozzolo (approved):

PR #964, opened 2024-09-26
Conditionally use high res version of SST/SIC

Sbozzolo:
Following https://github.com/CliMA/ClimaArtifacts/pull/45, we have a thinned down artifact that can be automatically downloaded if the higher resolution version is not available.

### DIFF
--- a/src/Utilities.jl
+++ b/src/Utilities.jl
@@ -6,11 +6,30 @@ modules in the coupler.
 """
 module Utilities
 
+import Artifacts
 import ClimaComms
 import ClimaCore as CC
 import Logging
 
 export swap_space!
+
+"""
+    artifact_exists(name)
+
+Return whether the artifact with the given name exists and is available to use.
+"""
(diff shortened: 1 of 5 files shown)


src/Utilities.jl
```diff
@@ -6,11 +6,30 @@ modules in the coupler.
...
 import ClimaCore as CC
 
 export swap_space!
 
+"""
+    artifact_exists(name)
+
+Return whether the artifact with the given name exists and is available to use.
+"""
+function artifact_exists(name)
```
juliasloan25, line 20: some tests for this would be good

juliasloan25 (commented):
Awesome, this will be nice to have!

src/Utilities.jl
Sbozzolo: I cannot test for non-existing artifacts because all the artifacts in the test suite are always automatically downloaded.

juliasloan25 (approved):

PR #967, opened 2024-09-26
specify grid in amip config

szy21:
## Purpose 

I think it might be helpful to add a bit redundancy and specify the grid in the amip config, so the amip simulation won't be affected while we modify the grid setup in the corresponding atmos yaml. What do others think?

The coupler yaml file seems alphabetic, so I kept the order. Is this required? I find it a bit annoying as e.g. the grid configs are not together, but I'm fine with it.

## Content

----


Sbozzolo:
Sounds good with me. We can restore it to non-grid version when all the changes are applied

PR #974, opened 2024-09-29
specify sponge in amip config

szy21:
## Purpose 

In #967 the grids are specified in the amip config. The sponge config needs to be consistent with the grid config, so we specify sponge in the amip config too.

## Content

----


szy21:
@juliasloan25 Could you update the tiny amip yaml with these changes and the changes in #967?

akshaysridhar (approved):

PR #976, opened 2024-09-30
use land branch with qsat fix, set psnow to zero

kmdeck:
## Purpose 
Uses ClimaLand branch [kd/bucket_qsat_fix](https://github.com/CliMA/ClimaLand.jl/pull/805) which sets q_sfc solely based on temperature.
Sets P_snow to zero for the land model

## To-do
Run AMIP run

## Content

----


szy21:
I think we can close this.

PR #978, opened 2024-09-30
include bucket diagnostics

juliasloan25:
## Purpose 
To facilitate debugging of AMIP simulations. We can use the existing diagnostics from ClimaLand - defined [here](https://github.com/CliMA/ClimaLand.jl/blob/735590ec50b132de62601b68ad341ec33e64a331/src/diagnostics/default_diagnostics.jl#L52-L85) for the bucket.

## Content
- [x] update `bucket_init` to initialize diagnostics from ClimaLand
  - required interface change: pass artifacts directory to `bucket_init`
- [x] add ClimaDiagnostics to `experiments/ClimaEarth` environment

----


juliasloan25:
see #979

PR #980, opened 2024-10-01
Misc refactoring

charleskawczynski:
This PR applies some miscellaneous refactoring:
 - use a concrete `eltype` in `AtmosSim`, this should improve inference in places
 - `[FT(0)]` -> `FT[0]` guarantees the vector eltype is of type `FT`, whereas the former pattern may not.
 - Call `ones` / `zeros` when working on scalar fields.
 - Add `;` to separate positional and keyword arguments
 -

### DIFF
--- a/experiments/ClimaEarth/components/ocean/slab_ocean.jl
+++ b/experiments/ClimaEarth/components/ocean/slab_ocean.jl
@@ -49,11 +49,11 @@ function slab_ocean_space_init(space, params)
     coords = CC.Fields.coordinate_field(space)
 
     # initial condition
-    T_sfc = CC.Fields.zeros(space) .+ params.T_init # FT(271) close to the average of T_1 in atmos
-    @. T_sfc += temp_anomaly(coords)
+    # FT(271) close to the average of T_1 in atmos
+    T_sfc = params.T_init .+ temp_anomaly.(coords)
 
     # prognostic variable
-    Y = CC.Fields.FieldVector(T_sfc = T_sfc)
+    Y = CC.Fields.FieldVector(; T_sfc = T_sfc)
 
     return Y, space
 end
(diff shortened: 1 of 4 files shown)


experiments/ClimaEarth/components/ocean/slab_ocean.jl
```diff
@@ -49,11 +49,11 @@ function slab_ocean_space_init(space, params)
     coords = CC.Fields.coordinate_field(space)
 
     # initial condition
-    T_sfc = CC.Fields.zeros(space) .+ params.T_init # FT(271) close to the average of T_1 in atmos
-    @. T_sfc += temp_anomaly(coords)
+    # FT(271) close to the average of T_1 in atmos
+    @. T_sfc = params.T_init + temp_anomaly(coords)
```
juliasloan25, line 53: this line fails because T_sfc isn't defined yet

juliasloan25 (approved):
Thanks for fixing these! After the `T_sfc` fix it looks good to me

experiments/ClimaEarth/components/ocean/slab_ocean.jl
charleskawczynski: Ah, right, thanks!

experiments/ClimaEarth/run_amip.jl
```diff
@@ -378,14 +378,14 @@ elseif mode_name in ("slabplanet", "slabplanet_aqua", "slabplanet_terra")
...
-        ρ_sfc = CC.Fields.zeros(boundary_space),
+        T_sfc = ones(boundary_space),
+        ρ_sfc = zeros(boundary_space),
         z0m = FT(0),
         z0b = FT(0),
         beta = FT(1),
-        α_direct = CC.Fields.ones(boundary_space) .* FT(1),
-        α_diffuse = CC.Fields.ones(boundary_space) .* FT(1),
-        area_fraction = CC.Fields.zeros(boundary_space),
+        α_direct = ones(boundary_space) .* FT(1),
```
akshaysridhar, line 386:
```suggestion
        α_direct = ones(boundary_space),
```

experiments/ClimaEarth/run_amip.jl
```diff
@@ -378,14 +378,14 @@ elseif mode_name in ("slabplanet", "slabplanet_aqua", "slabplanet_terra")
...
+        α_diffuse = ones(boundary_space) .* FT(1),
```
akshaysridhar, line 387:
```suggestion
        α_diffuse = ones(boundary_space)
```

experiments/ClimaEarth/run_amip.jl
charleskawczynski: I don't think we need this? `ones` in ClimaCore already inherits the undertype of the space:
```julia
function Base.ones(::Type{FT}, space::AbstractSpace) where {FT}
    field = Field(FT, space)
    data = parent(field)
    fill!(data, one(eltype(data)))
    return field
end
Base.ones(space::AbstractSpace) = ones(Spaces.undertype(space), space)
```

charleskawczynski:
Thanks for fixing this up, @juliasloan25!

PR #982, opened 2024-10-01
change initial condition and parameters for land

szy21:
## Purpose 

Change the initial condition of land water to be more consistent with the literature. Before the land is too wet and with this change, the land is much drier and probably too dry. I think it's still a better initial condition. I set it so that the initial W / W_f is 75%, which is the threshold for beta. Hopefully, the initial condition doesn't have too much effect.

Also changes to mono_surface: false in amip ci configs. This is mostly because one of the simulations becomes unstable with mono_surface: true. But it probably makes sense to unify them as all other amip configs have mono_surface: false.

Also changes some parameters of the land to be more consistent with the literature.

## Content

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_bucket.jl
+++ b/experiments/ClimaEarth/components/land/climaland_bucket.jl
@@ -97,9 +97,9 @@ function bucket_init(
     # the snow would melt in time `τc`. It prevents us from having to specially time step in cases where
     # all the snow melts in a single timestep.
     σS_c = FT(0.2) # critical snow water equivalent
-    W_f = FT(10) # bucket capacity
-    κ_soil = FT(0.7) # soil conductivity
-    ρc_soil = FT(2e8) # soil volumetric heat capacity
+    W_f = FT(0.2) # bucket capacity
+    κ_soil = FT(1.5) # soil conductivity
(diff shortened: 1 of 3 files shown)


kmdeck (approved):
Looks good. for reference, the SLIM model (Lague et al) uses W_f = 0.2 as a global constant also.

Im sure the bucket has some spin up time. we can test that offline forcing the bucket with ERA5

experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -97,9 +97,9 @@ function bucket_init(
     # the snow would melt in time `τc`. It prevents us from having to specially time step in cases where
     # all the snow melts in a single timestep.
     σS_c = FT(0.2) # critical snow water equivalent
-    W_f = FT(10) # bucket capacity
-    κ_soil = FT(0.7) # soil conductivity
-    ρc_soil = FT(2e8) # soil volumetric heat capacity
+    W_f = FT(0.2) # bucket capacity
+    κ_soil = FT(1.5) # soil conductivity
```
kmdeck, line 101: follows SLIM

kmdeck: also for rho_c_soil

juliasloan25 (approved):

PR #990, opened 2024-10-03
Refactor leaderboard code

### LINKED ISSUE #948: Leaderboard test fails due to undownloadable artifact
This test was commented out in #946

### LINKED ISSUE #956: Replace leaderboard with ClimaAnalysis leaderboard
Most of the leaderboard code has been ported and improved in ClimaAnalysis 0.5.8. When that is released, it would be good to switch to using that.

ph-kev:
closes #948 , closes #956 

This PR refactors the leaderboard code using the new features from ClimaAnalysis. This commit also deletes the old leaderboard code and the tests for it. Also, there is an off by one month issue when handling the dates and seasons. This issue arises because the first day of each month in the simulation data represents the monthly average of the previous month. This issue only affect the plots involving seasonal data. This is fixed in this commit.

The commit also moves the code to leaderboard.jl and a line is added to the pipeline to run the script.

One significant difference is the leaderboard which now plot the best and worst single model using only annual rather than averaging the error over the annual and seasonal data.

The leaderboard now looks like this:
bias_leaderboard

To compare, the previous leaderboard looks like this:
bias_leaderboard

The errors for the seasonal data change slightly because of the off by one month issue. The new bias plot for MAM looks like this:
bias_pr_MAM

The old bias plot for MAM looks this:
bias_pr_MAM

In addition to the plots above, there is now a plot with all the seasons for a subset of the variables (see `compare_vars_biases_groups` for which variables are plotted together in data_sources.jl).
bias_pr_all_seasons


Sbozzolo (commented):
Would it be possible to split the `leaderboard.jl` file in such a way that if one wants to add a new variable they would have to only change one file? (Essentially, moving the loaders to a separate file)

ph-kev:
> Would it be possible to split the `leaderboard.jl` file in such a way that if one wants to add a new variable they would have to only change one file? (Essentially, moving the loaders to a separate file)

I moved all the code involving data and preprocessing to `data_sources.jl`.

experiments/ClimaEarth/leaderboard/data_sources.jl
```diff
@@ -0,0 +1,134 @@
...
+    ("rsuscs", "sfc_sw_up_clr_t_mon"),
+    ("rldscs", "sfc_lw_down_clr_t_mon"),
+]
+
+# Add "pr" and the necessary preprocessing
+for (short_name, _) in sim_obs_short_names_no_pr
+    sim_var_dict[short_name] =
+        () -> begin
+            sim_var = get(ClimaAnalysis.SimDir(diagnostics_folder_path), short_name = short_name)
+            sim_var = ClimaAnalysis.shift_to_start_of_previous_month(sim_var)
```
Sbozzolo, line 59: Maybe I am getting confused, but shouldn't this be `shift_to_start_of_month`?

experiments/ClimaEarth/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,168 @@
...
+"""
+function compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
+    @info "Error against observations"
+
+    # Set up dict for storing simulation and observational data after processing
+    sim_obs_comparsion_dict = Dict()
+    seasons = ["ANN", "MAM", "JJA", "SON", "DJF"]
+
+    # Print dates for debugging
+    pr_var = sim_var_dict["pr"]() # it shouldn't matter what short name we use
```
Sbozzolo, line 27: Instead of assuming that `pr` is available, you can pick the first entry in the dict

experiments/ClimaEarth/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,168 @@
...
+                CairoMakie.save(
+                    joinpath(leaderboard_base_path, "bias_$(first(compare_vars_biases))_$season.png"),
+                    fig_bias,
+                )
+            else
+                CairoMakie.save(
+                    joinpath(leaderboard_base_path, "bias_$(first(compare_vars_biases))_total.png"),
+                    fig_bias,
+                )
+            end
```
Sbozzolo, line 101:
```suggestion
                CairoMakie.save(
                    joinpath(leaderboard_base_path, "bias_$(first(compare_vars_biases))_$season.png"),
                    fig_bias,
                )
```
It's fine to change the name to _ANN

Sbozzolo (commented):
I think this looks fantastic. Can you add an entry to the NEWS file to briefly mention to script? And can you add a section somewhere in the docs about "How do I add a new variable to the leaderboard?"

Sbozzolo (approved):
LGTM. Thank you! 

I'd recommend you tag Julia for a second review before merging.

experiments/ClimaEarth/run_amip.jl
```diff
@@ -856,6 +856,53 @@ The postprocessing includes:
 - Error against observations
 - Optional additional atmosphere diagnostics plots
 - Plots of useful coupler and component model fields for debugging
+
+### Add a new variable to compare against observations
```
juliasloan25, line 860: The comments in this section are really helpful! Do you think you could move them to the docs? We should add a new section `Model Output` (or something like that) and add a page about the leaderboard that contains this information.

The AMIP driver (run_amip.jl) is pretty complex right now but we're moving toward slimming it down, so it would be better to move a big comment like this to the docs

experiments/ClimaEarth/leaderboard/leaderboard.jl
```diff
@@ -0,0 +1,165 @@
+import ClimaAnalysis
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
+import GeoMakie
+import CairoMakie
+import Dates
+
+include("data_sources.jl")
+
+# leaderboard_base_path is the path to save the leaderboards
+# diagnostics_folder_path is the path to simulation data
```
juliasloan25, line 10: these comments are duplicates of the docstring below

juliasloan25 (commented):
This looks great, thank you Kevin! I just had a comment about moving some information to the docs, but after that it looks good to merge

juliasloan25 (approved):
Looks good! I'm not sure why the GPU slabplanet runs fail - is this rebased on the latest main?

ph-kev:
> Looks good! I'm not sure why the GPU slabplanet runs fail - is this rebased on the latest main?

I don't think this is rebased on the latest main. I am going to rebase and push right now.

PR #992, opened 2024-10-04
modify nightly amip config

szy21:
## Purpose 

This config runs for 180 days. Now the failure mode looks similar to the one in the target amip. The major change is the smoothing order, which helps with unrealistic surface fluxes. This is not ideal, but I still think it is good to have a setup that runs for longer. @akshaysridhar @juliasloan25 What do you think? Akshay may have ideas why the smoothing order matters for surface fluxes. I also changed dz_bottom as I think it makes sense to make it coarser when z_elem is smaller. This is not necessarily needed, we can explore the grids further once the current setup becomes stable.

## Content

----

### DIFF
--- a/.buildkite/nightly/pipeline.yml
+++ b/.buildkite/nightly/pipeline.yml
@@ -56,7 +56,7 @@ steps:
           - echo "--- Run simulation"
           - "julia --threads=3 --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/run_amip.jl --config_file $CONFIG_PATH/amip_coarse.yml --job_id amip_coarse1"
         artifact_paths: "experiments/ClimaEarth/output/amip/amip_coarse1_artifacts/*"
-        timeout_in_minutes: 1080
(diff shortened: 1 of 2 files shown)


juliasloan25:
180 days is a great improvement from the <1 day we were seeing recently! I also think it's a good sign that the failure is similar to that we see for regular AMIP

I'm not quite sure about the impact of the smoothing order. Maybe we need this because of sharp changes with the coarser resolution? If it seems reasonable to you two I'm happy with this

akshaysridhar (approved):
We can use this as a starting point for debugging these issues; (I'll verify in the meantime whether the issues occur specifically at mask boundaries)

szy21:
I further reduced z_max to 50000 and now it can run for a year: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/42#_. I'll change it in this PR and merge it, and open an issue about what we need to investigate. I may also explore the grid configuration more.

PR #993, opened 2024-10-04
rm extra Random.seed! call

juliasloan25:
## Purpose 
previous random seed update PR didn't remove the original call


Sbozzolo:
I also realized that the print statement for the seed is one per MPI process, could you adjust that since you are here?

Sbozzolo (approved):

PR #996, opened 2024-10-07
run 4 nightly AMIPs

juliasloan25:
## Purpose 
Following #992, our all-main tiny AMIP runs for 1 year successfully. This PR adds 3 additional runs to the pipeline to create a small ensemble. They all use the same config, but set the random seed independently, so should be slightly different.

### DIFF
--- a/.buildkite/nightly/pipeline.yml
+++ b/.buildkite/nightly/pipeline.yml
@@ -64,52 +64,48 @@ steps:
           slurm_cpus_per_task: 4
           slurm_ntasks: 1
           slurm_mem: 30GB
-        soft_fail: true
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Could we remove soft_fail: true?

szy21:
And for the schedule, could we change to Sunday - Thursday? On Friday evening there is atmos longrun that takes all the GPUs.

juliasloan25:
> And for the schedule, could we change to Sunday - Thursday? On Friday evening there is atmos longrun that takes all the GPUs.

Yes, I just updated the pipeline schedule!

PR #998, opened 2024-10-07
fix aquaplanet longruns

juliasloan25:
## Purpose 
See failing longruns [here](https://buildkite.com/clima/climacoupler-longruns/builds/811#019260a2-381a-4dc5-900e-ebe035810e89)

We have some longruns that evolve the atmos and surface states independently without coupling, so that we can independently test the infrastructure in the coupler that constructs and steps component models, and have a hierarchy of increasingly-complex runs. These `nocouple` runs are set up with a large coupling timestep (> t_end) so coupling never occurs.

This caused an error in the bucket diagnostics because the (very large) coupling timestep was being passed to the bucket model and used as its timestep. We actually have separate quantities `dt` and `dt_cpl` in our config files, where one is the component model timestep and the other is coupling timestep, but `dt` isn't being used except for in `ClimaAtmos.get_simulation`

There are a few things that should be done differently here
- We want an option to turn off bucket diagnostics, or compute at a different frequency
- We want each component model to use independent timesteps, so our configs should have options to specify each of them. Right now we just have dt and dt_cpl
- The failing longrun is an "aquaplanet" experiment, which runs the ocean component over the entire surface. This is fine, but the land is actually evaluated everywhere too, and just zeroed out. I imagine this might be why bucket_init was taking in the large dt_cpl value before - it wouldn't actually ever get evaluated using such a large step. If we know that the land model isn't being used, we should be able to avoid initializing it at all. The same applies in reverse for slabplanet_terra runs, which run land over the entire surface and zero out ocean. I'm not sure what the best short-term solution is, but I think in the medium-term we should separate AMIP and slabplanet drivers

The first point is addressed in this PR, the second is partially addressed here, and the third will be addressed separately.

## Content
- [x] add a flag to optionally output bucket diagnostics: set true by default, false for all aquaplanet runs
- [x] parse `dt` from the config dict and pass this to component models instead of `dt_cpl`

## To-do (in separate PRs)
- separate AMIP and slabplanet drivers; don't run land/ocean when they aren't used, instead of just masking them out
- add options for unique dts for each component model (atmos, land, ocean, sea ice)

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_bucket.jl
+++ b/experiments/ClimaEarth/components/land/climaland_bucket.jl
@@ -63,6 +63,7 @@ function bucket_init(
     t_start::Float64,
     energy_check::Bool,
     surface_elevation,
+    use_land_diagnostics::Bool,
 ) where {FT}
     if config != "sphere"
         println(
@@ -132,7 +133,7 @@ function bucket_init(
     @. Y.bucket.T = T_sfc_0 + temp_anomaly(coords.subsurface)
     # `surface_elevation` is a ClimaCore.Fields.Field(`half` level)
     orog_adjusted_T = CC.Fields.field_values(Y.bucket.T) .- lapse_rate .* CC.Fields.field_values(surface_elevation)
-    # Adjust T based on surface elevation (p.bucket.T_sfc is then set using the 
+    # Adjust T based on surface elevation (p.bucket.T_sfc is then set using the
     # set_initial_cache! function)
     parent(Y.bucket.T) .= parent(orog_adjusted_T)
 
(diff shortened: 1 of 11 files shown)


experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -149,12 +150,16 @@ function bucket_init(
     bucket_ode_function = CTS.ClimaODEFunction(T_exp! = exp_tendency!, dss! = CL.dss!)
     prob = SciMLBase.ODEProblem(bucket_ode_function, Y, tspan, p)
 
-    # Add diagnostics
-    netcdf_writer = CD.Writers.NetCDFWriter(domain.space.subsurface, artifacts_dir)
-    scheduled_diagnostics = CL.default_diagnostics(model, date_ref, output_writer = netcdf_writer)
+    # Add diagnostics, which are output daily by default
```
Sbozzolo, line 153: The output frequency depends on the duration of the simulation

juliasloan25:
> Disabling diagnostics feels like a workaround: the root cause of the problem is that we are adding a land model when there's no land. Is there any easy ways to fix that?

We can add two more branches to the big if/else [here](https://github.com/CliMA/ClimaCoupler.jl/blob/58242573bfc73e2f55ee5ba27d6c0a1ad822cbee/experiments/ClimaEarth/run_amip.jl#L264-L445) for the `slabplanet_aqua` and `slabplanet_terra` cases, but I didn't like this solution because it adds a lot of code to the driver for cases that aren't high priority. I think the solution is to create separate drivers for AMIP and for slabplanet runs - then we can add these cases to the slabplanet driver and still keep the AMIP driver clean

juliasloan25:
> > Disabling diagnostics feels like a workaround: the root cause of the problem is that we are adding a land model when there's no land. Is there any easy ways to fix that?
> 
> We can add two more branches to the big if/else [here](https://github.com/CliMA/ClimaCoupler.jl/blob/58242573bfc73e2f55ee5ba27d6c0a1ad822cbee/experiments/ClimaEarth/run_amip.jl#L264-L445) for the `slabplanet_aqua` and `slabplanet_terra` cases, but I didn't like this solution because it adds a lot of code to the driver for cases that aren't high priority. I think the solution is to create separate drivers for AMIP and for slabplanet runs - then we can add these cases to the slabplanet driver and still keep the AMIP driver clean

I think we could also come up with a different component model initialization approach - maybe we have smaller if/else chunks where we check the mode names and come up with flags to pass to generic component model constructors that dispatch off of them. E.g. we could have `land_init(:bucket)`, `land_init(nothing)` that return different land model types

trontrytel:
Would it be possible to merge this as a temporary fix?

juliasloan25:
@Sbozzolo are you okay with merging this as a temporary fix? It would be nice to be able to run the longruns this weekend. I opened an issue about setting up the component models in a more reasonable way #1011 that we can address when we have the time

Sbozzolo:
Okay

trontrytel (approved):

PR #999, opened 2024-10-07
Do not download bucket artifacts for ClimaLand > 0.15.1

Sbozzolo:
ClimaLand moved to ClimaArtifacts for the bucket artifacts, so that the files are automatically downloaded and there is no need anymore to force their download


juliasloan25 (approved):
Good catch! Maybe we should document changes like this so we can keep track of and eventually remove them

Issue #730, opened 2024-04-08
Remove ArtifactsWrappers, document input data


Sbozzolo:
Following [ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts/):

Files downloaded by ArtifactsWrappers:
- [x] sst
  - downloaded from https://gdex.ucar.edu/dataset/158_asphilli.html
- [x] sic
  - downloaded from https://gdex.ucar.edu/dataset/158_asphilli.html
- [x] co2 #1014 
  - values taken from https://gml.noaa.gov/webdata/ccgg/trends/co2/co2_mm_mlo.txt
  - .txt file was converted to a netcdf file of format matching SST and SIC - see https://github.com/CliMA/ClimaCoupler.jl/issues/408
- [ ] land/sea mask #1006 
  - downloaded from https://www.ncl.ucar.edu/Applications/Data/#cdf, file at https://www.ncl.ucar.edu/Applications/Data/cdf/landsea.nc
  - alt: new dataset from https://www.ncei.noaa.gov/products/etopo-global-relief-model
- [x] pr obs #781
- [x] cesm2 albedo
- [x] bareground albedo

PR #1006, opened 2024-10-09
Use higher resolution land-sea mask

### LINKED ISSUE #936: Land-sea mask is low resolution
The land sea mask we are currently using is only 64x64 points (in long-lat).

It is also a different land sea mask than ClimaLand.

### LINKED ISSUE #1057: remove perf/ folder
The performance of ClimaCoupler is as good as we need it to be for the foreseeable future. In addition, the runs produced by drivers/scripts in the `perf/` folder are largely not looked at currently. Because of this, we can remove the `perf/` folder altogether. As we do this, we should note the commit that removes it in our README, so that we can easily bring it back in the future if it becomes helpful again.

Sbozzolo:
As added in https://github.com/CliMA/ClimaArtifacts/pull/49

30 arcseconds is already pretty high resolution

### Issues closed by this PR
High-res land/sea mask closes #936
Artifact removal closes https://github.com/CliMA/ClimaCoupler.jl/issues/730 and https://github.com/CliMA/ClimaCoupler.jl/issues/1016
Manifest removal in seabreeze and perf closes https://github.com/CliMA/ClimaCoupler.jl/issues/1056
perf/ removal closes #1057 

This PR also removes all the artifact infrastructure from ClimaCoupler.


experiments/ClimaEarth/run_amip.jl
```diff
@@ -237,7 +242,13 @@ Note that land-sea area fraction is different to the land-sea mask, which is a b
 (masks are used internally by the coupler to indicate passive cells that are not populated by a given component model).
 =#
 
-land_area_fraction = SpaceVaryingInput(land_mask_data, "LSMASK", boundary_space)
+# Preprocess the file to be 1s and 0s before remapping into onto the grid
+land_area_fraction = SpaceVaryingInput(
+    land_mask_data,
+    "z",
+    boundary_space,
+    file_reader_kwargs = (; preprocess_func = (data) -> data >= 0),
```
juliasloan25, line 250: What does this preprocess_func do? Does it just select the data values >=0 ?

experiments/ClimaEarth/run_amip.jl
```diff
@@ -203,7 +208,7 @@ returns a `ComponentModelSimulation` object (see `Interfacer` docs for more deta
 
 #=
 ### Atmosphere
-This uses the `ClimaAtmos.jl` model, with parameterization options specified in the `atmos_config_object` dictionary.
+This uses the `ClimaAtmos.jl` model, with parametrization options specified in the `atmos_config_object` dictionary.
```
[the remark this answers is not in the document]
Sbozzolo: You are right! My spell checker was marking it wrong, so I thought it add an extra 'e'.

The other spelling seems to be used too:
https://rmets.onlinelibrary.wiley.com/doi/10.1256/qj.03.223

experiments/ClimaEarth/run_amip.jl
Sbozzolo: The dataset contains elevation, so a the land mask is where the elevation is >=0. The preprocessing function converts the dataset into a boolean mask

This is not completely true for places inland that are below sea levels, so we still need to work on this before merging the PR.

szy21:
We can either try using this and see if there is a problem (with our resolution I guess probably not), or use the one @akshaysridhar referenced [here](https://github.com/CliMA/ClimaCoupler.jl/issues/936#issuecomment-2347233326), which is already better than what we have now.

experiments/ClimaEarth/run_amip.jl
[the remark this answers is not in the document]
juliasloan25: probably a US vs UK thing :)

akshaysridhar:
> We can either try using this and see if there is a problem (with our resolution I guess probably not), or use the one @akshaysridhar referenced [here](https://github.com/CliMA/ClimaCoupler.jl/issues/936#issuecomment-2347233326), which is already better than what we have now.

Yeah this reference is in part based on the current `AMIP` documentation: (we reference that 360x180 dataset, but use a coarser 64x64 one in current `main`)

szy21:
> > We can either try using this and see if there is a problem (with our resolution I guess probably not), or use the one @akshaysridhar referenced [here](https://github.com/CliMA/ClimaCoupler.jl/issues/936#issuecomment-2347233326), which is already better than what we have now.
> 
> Yeah this reference is in part based on the current `AMIP` documentation: (we reference that 360x180 dataset, but use a coarser 64x64 one in current `main`). Thanks @Sbozzolo for updating this to use the latest artifacts.

This PR uses the topography artifact. Could you either check if this is ok, or add the higher resolution land sea mask to ClimaArtifact and update it in the coupler?

Sbozzolo:
I updated this PR to remove ArtifactWrappers and all the artifact infrastructure. 

I think this can only be merged after https://github.com/CliMA/ClimaAtmos.jl/pull/3378

NEWS.md
```diff
@@ -13,8 +13,19 @@ low-resolution version of such files is automatically downloaded when a
 higher-resolution version is not available. Please, refer to
 [ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts) for more information.
 
+### A higher resolution land-sea mask is now used and automatically downloaded - PR [#1006](https://github.com/CliMA/ClimaCoupler.jl/pull/1006)
+
+A 60 arcsecond land-sea mask constructed from topographic data is now used.
+Topographic data is automatically downloaded and a land-sea mask is constructed
+by identifying where elevation is greater than 0. Note, this can lead to
+misidentification of ocean in some areas of the globe that are inland but below
+sea level (Dead Sea, Death Valley, ...).
```
juliasloan25, line 22: Do we have a plan to correct this ocean misidentification in the future?

experiments/ClimaEarth/Manifest-v1.11.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.1"
```
juliasloan25, line 3: Maybe we want to stay at 1.11.1 unless there's a reason to go back to 1.11.0

perf/Manifest.toml
```diff
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.5"
```
juliasloan25, line 3: we should update `perf/Manifest-v1.11.toml` as well

juliasloan25 (approved):
The only remaining question is how to handle areas that are inland and below sea level, but besides that this looks good to merge

NEWS.md
akshaysridhar: @juliasloan25 One of two options is feasible here 
(1) Write / use existing tools for closed contour detection to address such areas within continent boundaries, 
(2) Use a merged dataset with the ETOPO2022 information based on the existing NCAR dataset (lower resolution 360x180 npts), which already contains classifications as follows: 
`0=ocean, 1=land, 2=lake, 3=small island, 4=ice shelf`, see here: https://www.ncl.ucar.edu/Document/Functions/Shea_util/landsea_mask.shtml

juliasloan25: Okay! Could you open an issue for it so we can keep track of resolving this in the future?

akshaysridhar: @juliasloan25 Please see #1049

.buildkite/pipeline.yml
```diff
@@ -42,17 +42,11 @@ steps:
       - echo "--- Instantiate ClimaEarth env"
       - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.develop(path=\".\")'"
       - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.add(\"MPI\"); Pkg.add(\"CUDA\")'"
+      # ProfileCanvas is used by the perf jobs
```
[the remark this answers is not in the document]
Sbozzolo: Added to the header of the file

perf/flame.jl
```diff
@@ -1,6 +1,8 @@
 # flame.jl: provides allocation breakdown for individual backtraces for single-process unthredded runs
 # and check for overall allocation limits based on previous runs
 # copied and modified from `ClimaAtmos/perf`
+#
+# To run this, add ProfileCanvas to the ClimaEarth environment
```
juliasloan25, line 5: This should be added to flame_diff.jl and flame_test.jl too

Sbozzolo: They don't seem to need ProfleCanvas

Sbozzolo: Because they import ProfileCanvasDiff (which copies the code of ProfileCanvas)

.buildkite/pipeline.yml
```diff
@@ -48,27 +41,17 @@ steps:
       - echo "--- Instantiate ClimaEarth env"
       - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.develop(path=\".\")'"
       - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.add(\"MPI\"); Pkg.add(\"CUDA\")'"
+      - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.add(\"MPI\"); Pkg.add(\"CUDA\");"
```
juliasloan25, line 44:
```suggestion
      - "julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.add(\"MPI\"); Pkg.add(\"CUDA\");'"
```

Sbozzolo: Thank you!

PR #1008, opened 2024-10-09
Make land diagnostics monthly

Sbozzolo:
Instead of the default daily


szy21 (approved):

Sbozzolo:
> I thought you had said before that the frequency for diagnostics is dependent on the simulation length - or was that for outputting rather than the averaging frequency?

That was for ClimaAtmos. I also thought that being true for ClimaLand as well, but when I checked I realized that it is not and it is left up to the user to pick.

PR #1010, opened 2024-10-10
clipping sea ice tendencies

szy21:
## Purpose 

In Holloway and Manabe 1971 it says

> If the solution to this heat balance equation results in a value of T* above freezing, this value is set back to freezing with the implicit assumption that the excess heat is used for melting pack ice.

The current code sets the tendency to zero if it results in T above freezing, which differs from the paper. I change it to clip the tendency to (T_freeze - T) / dt, which should match the paper more closely. I also remove a comment about sublimation which doesn't seem to be true in the code.

Edit: One out of four simulations break with dt 240 in the latest nightly build: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/73. So I change dt back to 180 in this PR, and change t_end to one and a half years.

## To-do

I will revert the change in nightly pipeline before merging.

## Content

----

### DIFF
--- a/experiments/ClimaEarth/components/ocean/prescr_seaice.jl
+++ b/experiments/ClimaEarth/components/ocean/prescr_seaice.jl
@@ -14,7 +14,7 @@ include("../slab_utils.jl")
 
 Ice concentration is prescribed, and we solve the following energy equation:
 
-    (h * ρ * c) d T_sfc dt = -(F_turb_energy + F_radiativead) + F_conductive
+    (h * ρ * c) d T_sfc dt = -(F_turb_energy + F_radiative) + F_conductive
 
     with
     F_conductive = k_ice (T_base - T_sfc) / (h)
@@ -24,9 +24,7 @@ Ice concentration is prescribed, and we solve the following energy equation:
     as well as a conductive flux that depends on the temperature difference
     across the ice layer (with `T_base` being prescribed).
 
-In the current version, the sea ice has a prescribed thickness, and we assume that it is not
-sublimating. That contribution has been zeroed out in the atmos fluxes.
-
+In the current version, the sea ice has a prescribed thickness.
 """
 struct PrescribedIceSimulation{P, Y, D, I} <: Interfacer.SeaIceModelSimulation
     params::P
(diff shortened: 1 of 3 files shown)


juliasloan25 (approved):

.buildkite/nightly/pipeline.yml
```diff
@@ -65,47 +65,47 @@ steps:
...
+      #     - "julia --threads=3 --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/run_amip.jl --config_file $CONFIG_PATH/amip_coarse.yml --job_id amip_coarse4"
+      #   artifact_paths: "experiments/ClimaEarth/output/amip/amip_coarse4_artifacts/*"
+      #   timeout_in_minutes: 1080
+      #   env:
+      #     CLIMACOMMS_DEVICE: "CUDA"
+      #   agents:
+      #     slurm_gpus_per_task: 1
+      #     slurm_cpus_per_task: 4
+      #     slurm_ntasks: 1
+      #     slurm_mem: 30GB
```
Sbozzolo, line 111: Do you want these commented out?

experiments/ClimaEarth/components/ocean/prescr_seaice.jl
```diff
@@ -174,12 +172,11 @@ function ice_rhs!(du, u, p, _)
...
-
-    # do not count tendencies that lead to temperatures above freezing, and mask out no-ice areas
+    # If tendencies lead to temperature above freezing, set temperature to freezing
+    @. rhs = min(rhs, (T_freeze - Y.T_sfc) / p.dt)
+    # mask out no-ice areas
     area_mask = Regridder.binary_mask.(area_fraction)
-    threshold = zero(FT)
-    unphysical = @. Regridder.binary_mask.(T_freeze - (Y.T_sfc + FT(rhs) * FT(p.dt)), threshold) .* area_mask
-    parent(dY.T_sfc) .= parent(rhs .* unphysical)
+    parent(dY.T_sfc) .= parent(rhs .* area_mask)
```
Sbozzolo, line 179: Would this work?

```suggestion
    dY.T_sfc .= rhs .* area_mask
```

.buildkite/nightly/pipeline.yml
szy21: No, I will revert this before merging.

experiments/ClimaEarth/components/ocean/prescr_seaice.jl
szy21: I think there might be a reason why we have parent before, but let me try.

szy21: it works:)

szy21:
Update: One out of four simulations break with dt 240 in the latest nightly build: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/73. So I change dt back to 180 in this PR, and change t_end to one and a half years.

experiments/ClimaEarth/components/ocean/prescr_seaice.jl
Sbozzolo: I suspected so

PR #1014, opened 2024-10-11
Tr/move co2 to clima artifacts

### LINKED ISSUE #997: Move CO2 artifact to ClimaArtifacts
And keep it as text file. It will need a little bit of parsing, but it should be relatively straightforward

imreddyTeja:
## Purpose 
Reduce usage of ArtifactWrapper
Closes #997 
Relates to ClimaArtifact [PR 51](https://github.com/CliMA/ClimaArtifacts/pull/51)

## To-do
- Check that this does not change results of run_amip.jl 

## Content
- Replace ArtifactWrapper in co2_dataset_path with a clima artifact
- The new co2 artifact contains a text file with co2 monthly means and a text file with daily means
- The old co2 artifact was a NetCDF file generated from the monthly means text file
- run_amip.jl now uses the monthly txt file and parses it into a TimeVaryingInput. 
There are several differences between the old data in the artifact and the new data. First,
the new data does not have lat/lon, but the old Netcdf file did not vary in space even though the dimensions existed.
Second, the dates in the new file all fall exactly on the 15th of the month, while the old file varied by a few hours in each direction.  Third the mean co2 is very slightly off (greatest diff is .01%) due to floating points during unit conversion, but this also seems like an issue in the old data file. 
- Change truncate test to not rely on old NetCDF file

----

### DIFF
--- a/artifacts/artifact_funcs.jl
+++ b/artifacts/artifact_funcs.jl
@@ -1,5 +1,4 @@
 import ArtifactWrappers as AW
-import ClimaUtilities.ClimaArtifacts: @clima_artifact
 
 function mask_dataset_path()
     mask_dataset = AW.ArtifactWrapper(
--- a/artifacts/artifact_funcs.jl
+++ b/artifacts/artifact_funcs.jl
@@ -1,10 +1,6 @@
 import ArtifactWrappers as AW
 import ClimaUtilities.ClimaArtifacts: @clima_artifact
 
-function co2_dataset_path(; context = nothing)
-    return @clima_artifact("co2_dataset", context)
-end
-
 function mask_dataset_path()
     mask_dataset = AW.ArtifactWrapper(
         @__DIR__,
--- a/artifacts/artifact_funcs.jl
+++ b/artifacts/artifact_funcs.jl
@@ -1,15 +1,8 @@
 import ArtifactWrappers as AW
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
 
-function co2_dataset_path()
-    co2_dataset = AW.ArtifactWrapper(
-        @__DIR__,
-        "co2",
-        AW.ArtifactFile[AW.ArtifactFile(
-            url = "https://caltech.box.com/shared/static/xg028wnsn57wam6euwrh98fe43ibei8g.nc",
-            filename = "mauna_loa_co2.nc",
-        ),],
-    )
-    return AW.get_data_folder(co2_dataset)
+function co2_dataset_path(; context = nothing)
+    return @clima_artifact("co2_dataset", context)
 end
 
 function mask_dataset_path()
--- a/experiments/ClimaEarth/run_amip.jl
+++ b/experiments/ClimaEarth/run_amip.jl
@@ -196,7 +196,7 @@ catch error
         "MODEL.ICE.HAD187001-198110.OI198111-202206_lowres.nc",
     )
 end
-co2_data = joinpath(co2_dataset_path(), "co2_mm_mlo.txt")
+co2_data = joinpath(@clima_artifact("co2_dataset", comms_ctx), "co2_mm_mlo.txt")
 land_mask_data = joinpath(mask_dataset_path(), "seamask.nc")
 
 #=
@@ -340,9 +340,11 @@ if mode_name == "amip"
     ## CO2 concentration from temporally varying file
     CO2_text = DelimitedFiles.readdlm(co2_data, Float64; comments = true)
     # The text file only has month and year, so we set the day to 15th of the month
-    CO2_dates = Dates.DateTime.(CO2_text[:, 1], CO2_text[:, 2]) + Dates.Day(14)
+    years = CO2_text[:, 1]
+    months = CO2_text[:, 2]
+    CO2_dates = Dates.DateTime.(years, months) + Dates.Day(14)
     CO2_times = period_to_seconds_float.(CO2_dates .- date0)
-    # convert from ppm to fraction
+    # convert from ppm to fraction, data is in fourth column of the text file
     CO2_vals = CO2_text[:, 4] .* 10^(-6)
     CO2_timevaryinginput = TimeVaryingInput(CO2_times, CO2_vals;)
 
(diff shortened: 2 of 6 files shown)


artifacts/artifact_funcs.jl
```diff
@@ -1,15 +1,8 @@
...
-    co2_dataset = AW.ArtifactWrapper(
-        @__DIR__,
-        "co2",
-        AW.ArtifactFile[AW.ArtifactFile(
-            url = "https://caltech.box.com/shared/static/xg028wnsn57wam6euwrh98fe43ibei8g.nc",
-            filename = "mauna_loa_co2.nc",
-        ),],
-    )
-    return AW.get_data_folder(co2_dataset)
+function co2_dataset_path(; context = nothing)
```
juliasloan25, line 4: maybe we can just call `@clima_artifact` directly instead of `co2_dataset_path`, like we do for SST and SIC [here](https://github.com/CliMA/ClimaCoupler.jl/blob/6d022101262c2c93c572979b7562dcc929bfc2ab/experiments/ClimaEarth/run_amip.jl#L184)

experiments/ClimaEarth/run_amip.jl
```diff
@@ -336,7 +338,13 @@ if mode_name == "amip"
     )
 
     ## CO2 concentration from temporally varying file
-    CO2_timevaryinginput = TimeVaryingInput(co2_data, "co2", boundary_space, reference_date = date0)
+    CO2_text = DelimitedFiles.readdlm(co2_data, Float64; comments = true)
+    # The text file only has month and year, so we set the day to 15th of the month
+    CO2_dates = Dates.DateTime.(CO2_text[:, 1], CO2_text[:, 2]) + Dates.Day(14)
```
juliasloan25, line 343: For future users/developers, it would be helpful to somehow show which entry of `CO2_text` is the month and which is the year. You could do something like:
```julia
year = CO2_text[:, 1]
month = CO2_text[:, 2]
```
or just include it in a comment

juliasloan25: It would also be good to include this information about the format of the data in the documentation for the artifact in ClimaArtifacts, if it isn't already there. Including that the CO2 data itself is in the 4th column

juliasloan25 (commented):
This is great, thank you Teja! Just a couple minor comments

artifacts/artifact_funcs.jl
juliasloan25: you can also delete this function once you make this change

experiments/ClimaEarth/run_amip.jl
```diff
@@ -194,7 +196,7 @@ catch error
         "MODEL.ICE.HAD187001-198110.OI198111-202206_lowres.nc",
     )
 end
-co2_data = joinpath(co2_dataset_path(), "mauna_loa_co2.nc")
+co2_data = joinpath(@clima_artifact("co2_dataset", comms_ctx), "co2_mm_mlo.txt")
```
juliasloan25, line 199: I'm not sure why this change fails. One difference I notice is that now we're passing `comms_ctx` into `@clima_artifact`, where it was `nothing` before. I would expect this to work, so maybe it's a bug in ClimaArtifacts?

Since we don't need to download this artifact lazily, we don't need to provide the context so we could try calling `@clima_artifact` without the context. We should also figure out if this is a bug in ClimaArtifacts or if we're using `@clima_artifact` in a way that isn't meant to be supported

imreddyTeja: It seems like there's a second Artifacts.toml file in experiments/ClimaEarth, and adding @clima_artifacts made it search for the co2 dataset in that file. Because the co2_data is not used anywhere else at the moment, should I remove it from the main Artifact.toml and move it to the Artifacts.toml file in experiments/ClimaEarth? Or should I revert to using co2_dataset_path()? It also seems like the Artifacts.toml file in experiments/ClimaEarth has the landmark artifact, which is also the main  Artifacts.toml, and it is accessed using mask_dataset_path() instead of @clima_artifacts

juliasloan25: Oh good catch! I don't think we should have two Artifacts.toml files when all the artifacts are used for ClimaEarth experiments. I would recommend adding CO2 to `experiments/ClimaEarth/Artifacts.toml`, then we can get rid of `artifacts/Artifacts.toml` altogether. Our goal is to slim down the `artifacts/` folder as we switch to using ClimaArtifacts, and eventually get rid of the folder

imreddyTeja: CO2 data is now in `experiments/ClimaEarth/Artifacts.toml`. There are only three remaining entries in `artifacts/Artifacts.toml`: land_mask, sic, and sst. I can't find any usage of sic and set. It seems like their usage is replaced by historical_sst_sic. When I move land_mask to ClimaArtifacts I will also get rid of the unused sic and set, which should just remove all files from `artifacts/` except the script that downloads the Atmos and Land artifacts.

Sbozzolo: I think `sic` and `sst` should be removed, they are not needed anymore. I think we also might be at the point where all  land artifact are automatically downloaded (the one missing was the albedo).

Sbozzolo (approved):

artifacts/artifact_funcs.jl
```diff
@@ -1,16 +1,5 @@
...
-        @__DIR__,
-        "co2",
-        AW.ArtifactFile[AW.ArtifactFile(
-            url = "https://caltech.box.com/shared/static/xg028wnsn57wam6euwrh98fe43ibei8g.nc",
-            filename = "mauna_loa_co2.nc",
-        ),],
-    )
-    return AW.get_data_folder(co2_dataset)
-end
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
```
juliasloan25, line 2: we don't need this import anymore

experiments/ClimaEarth/run_amip.jl
juliasloan25: This sounds good, thank you Teja!

juliasloan25 (approved):
Looks good! I just had one minor comment, then it's good to squash and merge

PR #1018, opened 2024-10-15
make one nightly amip deterministic

szy21:
## Purpose 

This can make debugging a bit easier. Also changes one longrun job from 4 GPU to 1 GPU as the SYPD doesn't differ much. I'll open an issue to keep track of this.

cc @trontrytel 

## Content

----

### DIFF
--- a/config/longrun_configs/amip_target_topo_diagedmf_gpu.yml
+++ b/config/longrun_configs/amip_target_topo_diagedmf_gpu.yml
@@ -16,7 +16,7 @@ netcdf_output_at_levels: true
 output_default_diagnostics: true
 start_date: "20100101"
 surface_setup: "PrescribedSurface"
-t_end: "366days"
+t_end: "300days"
 topo_smoothing: true
 topography: "Earth"
 coupler_toml_file: "toml/amip_target_topo_diagedmf.toml"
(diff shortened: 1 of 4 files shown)


config/longrun_configs/amip_target_topo_diagedmf_gpu.yml
```diff
@@ -16,7 +16,7 @@ netcdf_output_at_levels: true
 output_default_diagnostics: true
 start_date: "20100101"
 surface_setup: "PrescribedSurface"
-t_end: "366days"
+t_end: "300days"
```
juliasloan25, line 19: did you mean to include this change too?

juliasloan25 (approved):
Should we change other longruns to use 1 GPU too?

config/longrun_configs/amip_target_topo_diagedmf_gpu.yml
szy21: Yes, the longrun has 1 day walltime limit and we can get about 300 days with the current SYPD.

szy21:
> Should we change other longruns to use 1 GPU too?

I think this is the only one in the longrun that uses more than 1 GPU?

PR #1027, opened 2024-10-17
use 55km top and viscous sponge in nightly amip

szy21:
## Purpose 

One failure mode in the nightly amip is the top layer becomes too cold with too large horizontal velocity. Adding a viscous sponge seems to help. I extended the top back to 55km and increased dt back to 240 seconds. This config works for 2 nightly amip runs: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/111#_. I'll be a bit optimistic and get it in and see how it works.

## Content

----


trontrytel (approved):

szy21:
I'll change the target amip config as well to make it consistent.

PR #1026, opened 2024-10-16
run CI on julia 1.11

juliasloan25:
## Purpose 
Updates everything to use Julia v1.11

Closes #1001 

## Content
- [x] update all CI workflows to Julia v1.11
- [x] update compat entries to use 1.11-compatible versions
- [x] update dependencies in existing Manifests
- [x] generate new Manifests for Julia v1.11
- [x] update package README and NEWS to reflect these changes

----
- [x] I have read and checked the items on the review checklist.

PR #1028, opened 2024-10-17
Make `modules` uniform across buildkite pipelines

akshaysridhar:
Update to use `climacommon` (auto latest)
```
	modified:   .buildkite/amip/pipeline.yml
	modified:   .buildkite/benchmarks/pipeline.yml
	modified:   .buildkite/hierarchies/pipeline.yml
	modified:   .buildkite/longruns/pipeline.yml
	modified:   .buildkite/nightly/pipeline.yml
	modified:   .buildkite/pipeline.yml
```

### DIFF
--- a/.buildkite/amip/pipeline.yml
+++ b/.buildkite/amip/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   queue: clima
   slurm_time: 96:00:00
-  modules: common
+  modules: climacommon
 
 env:
   JULIA_MAX_NUM_PRECOMPILE_FILES: 100
--- a/.buildkite/benchmarks/pipeline.yml
+++ b/.buildkite/benchmarks/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   queue: clima
   slurm_time: 24:00:00
-  modules: common
+  modules: climacommon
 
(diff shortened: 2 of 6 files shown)


akshaysridhar:
Is there a reason we currently pin modules in certain pipelines to May 24 versions ?

juliasloan25:
Thanks for doing this! I had included this in the Julia 1.11 update #1026 but we can have it separate too (and I missed our regular CI pipeline)

I think it would be better to specify `climacommon/2024_10_09` rather than just `climacommon`, so that we can control when we update to the latest version. For example, `2024_10_09` was released with Julia 1.11 when ClimaCore wasn't yet compatible with it, leading to some bugs in ClimaCoupler. If we use a specific version, we can make any required changes when we update

akshaysridhar:
Great, thank you @juliasloan25! I had started this to PR as a solution to some nightly pipeline instantiation issues I was seeing - I agree that being more specific with the version gives us more control ; (I'm happy to close this and fold it into the #1026 PR)

akshaysridhar:
Closing as this is addressed in #1026 with @juliasloan25 fixing the regular CI pipeline as well.

PR #1029, opened 2024-10-17
don't add atmos default diags in coupler

juliasloan25:
## Purpose 
Currently, the nightly AMIP runs are broken because we're outputting duplicate diagnostics. This is because we're saving diagnostics via the coupler that are already output by atmos.

Also reverts JuliaFormatter to v1 after 2.0.0 release was yanked, so that formatter check will pass

## To-do
- [x] only specify additional atmos diagnostics defined in the coupler, not diagnostics we get from atmos
- [ ] check outputs in CI
- [x] revert formatter

### DIFF
--- a/.dev/Project.toml
+++ b/.dev/Project.toml
@@ -2,4 +2,4 @@
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 
 [compat]
-JuliaFormatter = "2"
+JuliaFormatter = "1"
--- a/experiments/ClimaCore/heat-diffusion/run.jl
+++ b/experiments/ClimaCore/heat-diffusion/run.jl
@@ -256,7 +256,7 @@ function coupler_solve!(stepping, ics, parameters)
 
 
     ## coupler stepping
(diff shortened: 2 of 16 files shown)


Sbozzolo:
Maybe we should just save the coupler diagnostics in a different folder?

juliasloan25:
> Maybe we should just save the coupler diagnostics in a different folder?

Is there a way of specifying the output folder from the config? If not, I think we would need to duplicate the atmos diagnostics in the coupler so we can use a ClimaDiagnostics writer that we define with the output path we want. Or am I missing something?

juliasloan25:
> > Maybe we should just save the coupler diagnostics in a different folder?
> 
> Is there a way of specifying the output folder from the config? If not, I think we would need to duplicate the atmos diagnostics in the coupler so we can use a ClimaDiagnostics writer that we define with the output path we want. Or am I missing something?

We discussed offline and decided to merge the fix in this PR for now, and investigate further in ClimaDiagnostics.jl why the duplicate diagnostics were not caught before

Sbozzolo (approved):

szy21 (approved):

PR #1033, opened 2024-10-19
add amip test to exp/ClimaEarth/

### LINKED ISSUE #984: add small coupler AMIP test
To prevent upstream packages from breaking the coupler, we want to add downstream tests in those packages that exercises our AMIP experiment code. It isn't very helpful to run ClimaCoupler's `run_tests.jl` because while this does test coupler functionality like flux calculation or coupler interfaces with simple simulation objects, it is independent of many important upstream packages and doesn't test their integration.

Ideally, we would be able to add a test running AMIP in our `test` environment and include this in our `run_tests.jl`, which upstream packages can easily use. However, the distinction between `test` and `experiments/ClimaEarth` environments makes this difficult; many packages used in `experiments/ClimaEarth`, and specifically those we're interested in testing, are not in the `test` environment. Because of this, it's easier to add the AMIP experiment test in the `experiments/ClimaEarth` environment.

We should make a new test that runs the actual AMIP driver in a simple (coarse) setup for just a few timesteps. This can be used by upstream packages to test if their changes will break the interfaces used by the coupler, though it still won't test changes to simulation stability.

juliasloan25:
## Purpose 
closes #984 

see issue for details

## Content
- [x] add a file `experiments/ClimaEarth/test/runtests.jl` that can be used by upstream packages for downstream testing
- [x] add a small AMIP test and call it from `runtests.jl`
- [x] add a config file that is coarse and short for this test


experiments/ClimaEarth/test/amip_test.jl
```diff
@@ -0,0 +1,11 @@
...
+# This script runs a coarse AMIP simulation for a short period of time.
+# This is meant to be used by upstream packages to test if changes to those
+# packages affect the interfaces used by ClimaCoupler.jl.
+# Note that this test doesn't check stability or correctness of the model.
+
+driver_path = joinpath("..", "run_amip.jl")
+config_file = "amip_test.yml"
+
+# Run the AMIP test
+run(`julia --project $driver_path --config_file=$config_file --job_id="amip_test"`)
```
Sbozzolo, line 11: This will behave differently depending on the folder where it is run from (because of `driver_path`).

I think we should be able to `include("run_amip.jl")` instead

juliasloan25: Yeah, I wasn't sure about the paths. I didn't want to just call `include` because that won't let us specify the config file. Also `run_amip` is in a different directory so I didn't think we're able to include it directly like that

Sbozzolo: You can specify the config file by setting the `ARGS` variable, and you can also include across directories. I think `include("../run_amip.jl`)` should work (at the end of the day, this is how we include stuff in `src`).

Sbozzolo (approved):

PR #1040, opened 2024-10-22
fix simulation length

juliasloan25:
## Purpose 
The downstream coupler test is meant to run for only two timesteps, but I had accidentally copied the tiny AMIP config without changing `t_end`. It is fixed here


juliasloan25:
duplicate of https://github.com/CliMA/ClimaCoupler.jl/pull/1039

PR #1041, opened 2024-10-22
use 65km top in amip

szy21:
## Purpose 

Moves model top to 65km in amip configs. For the coarse resolution one I have to increase the number of levels to have enough levels in the sponge to stabilize it, and as a result I reduce t_end. It is stable for a year: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/124. It may still be finicky and I may need to fine-tune the grids in the future.

## Content

----


szy21:
The model top pressure goes below 1Pa, so let me think about it before merging. Here is July average pressure near south pole.
image

szy21:
As discussed in the AMIP meeting I will put this on hold.

PR #1045, opened 2024-10-28
Make target AMIP reproducible

Sbozzolo:
We need to control the seed to ensure that the evolution is reproducible.


szy21 (approved):
The coupler team prefers the config to be alphabetical I think.

Sbozzolo:
> The coupler team prefers the config to be alphabetical I think.

I thought so, then I read the config, and my brain read `dz` as just `z` and concluded that this file was not alphabetized.

I updated it.

PR #1052, opened 2024-11-01
update deps

juliasloan25:
## Purpose 
biweekly dependency updates

updates to:
- ClimaAtmos v0.27.8
- ClimaLand v0.15.4

### DIFF
--- a/.dev/.gitignore
+++ b/.dev/.gitignore
@@ -1,1 +1,0 @@
-Manifest.toml
--- a/experiments/ClimaCore/Manifest-v1.11.toml
+++ b/experiments/ClimaCore/Manifest-v1.11.toml
@@ -44,9 +44,9 @@ version = "0.1.38"
 
 [[deps.Adapt]]
 deps = ["LinearAlgebra", "Requires"]
-git-tree-sha1 = "6a55b747d1812e699320963ffde36f1ebdda4099"
+git-tree-sha1 = "50c3c56a52972d78e8be9fd135bfb91c9574c140"
 uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
-version = "4.0.4"
+version = "4.1.1"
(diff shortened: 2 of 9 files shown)


update_deps.sh
```diff
@@ -0,0 +1,21 @@
...
+# julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.update()'
+julia --project=perf/ -e 'using Pkg; Pkg.update()'
+
+# Update checked-in Manifests for Julia v1.11
+juliaup add 1.11.1
+juliaup default 1.11.1
+# julia --project=experiments/ClimaCore/ -e 'using Pkg; Pkg.update()'
+# julia --project=experiments/ClimaCore/sea_breeze/ -e 'using Pkg; Pkg.update()'
+# julia --project=experiments/ClimaEarth/ -e 'using Pkg; Pkg.update()'
+julia --project=perf/ -e 'using Pkg; Pkg.update()'
```
Sbozzolo, line 21: I use juliaup differently, so I personally wouldn't use this script. Also, this script would precompile all the packages and take a while. You can set `JULIA_PKG_PRECOMPILE_AUTO=0` to avoid those precompilation

Previously, I was using Charlie's PkgDevTools, but this doesn't work with Julia 1.11 (https://github.com/CliMA/PkgDevTools.jl/issues/11).

juliasloan25: Thanks, I'll set `JULIA_PKG_PRECOMPILE_AUTO=0`.

Do you have any suggestions for how to change the juliaup usage? Even if some of the developers have different ways of using it, I think it's helpful to have some script that automates the updates.

Sbozzolo (approved):
Up to you on what you want to do with the `update_deps.sh`

juliasloan25:
> Up to you on what you want to do with the `update_deps.sh`

I'll remove it from the repo and keep it for my personal use

PR #1059, opened 2024-11-05
Add specific component dt's to cli_options and parse

### LINKED ISSUE #1017: allow independent component model timesteps
Currently, each component model takes in a timestep that determines how often it will be stepped, but all component models take in the same timestep value. To be able to take advantage of each model using an independent timestep, we need to add arguments for each model's timestep. We also need to check and enforce that these and the coupling timestep are integer multiples of each other.

### To Do
- [x] instead of passing `Δt_component` to each component model, add arguments in `cli_options.jl` for `dt_atmos`, `dt_ocean`, `dt_land`, and `dt_seaice`, parse these, and pass them to the component model `init` functions
  - note that the atmos `get_simulation` function expects a value `dt` in the config file we pass in, so we may need to rename `dt_atmos` for that
- [x] verify that the coupling timestep is divisible each of the component model timesteps, and give an informative error message if this condition isn't met
- [x] add an experiment with component models using different timesteps

part of O2.5.2

imreddyTeja:
If both dt and component dts are passed in, the component dts are chosen, and a warning is shown.
Ff the coupler dt is not divisible by a component dt, an error is thrown with a message identifying the component dt which the coupler dt is not a multiple of.

The usage of the SciMLBase `step!` function did not need to be modified because all the component models call `step!` like the following

```
Interfacer.step!(sim::ClimaAtmosSimulation, t) = Interfacer.step!(sim.integrator, t - sim.integrator.t, true)
```

The SciMLBase [docs](https://docs.sciml.ai/DiffEqDocs/stable/basics/integrator/) say 
"passing a dt will make the integrator continue to step until at least integrator.t+dt, and passing true as the third argument will add a tstop to force it to step up to integrator.t+dt, exactly".

I also ran it interactively with the config for the component dts experiment, and all of the components with integrators end with the proper `t`. 

## Purpose 
Closes #1017 

## To-do
- [x] Add an experiment 

## Content
- add component dts as optional arguments to `cli_options.jl`
- make `run_amip.jl` support component dts or a general dt

----


experiments/ClimaEarth/run_amip.jl
```diff
@@ -166,8 +163,31 @@ land_sim_name = "bucket"
 t_end = Float64(time_to_seconds(config_dict["t_end"]))
 t_start = 0.0
 tspan = (t_start, t_end)
-Δt_component = Float64(time_to_seconds(config_dict["dt"]))
 Δt_cpl = Float64(config_dict["dt_cpl"])
+component_dt_names = [:dt_atmos, :dt_land, :dt_ocean, :dt_seaice]
+# if any of the component dt's are not specified, set them all to the overall dt
```
juliasloan25, line 168: Could you reword this comment to emphasize that the `if` branch uses individual component model dts, and the `else` branch ignores them to use the general dt?

experiments/ClimaEarth/run_amip.jl
```diff
@@ -166,8 +163,31 @@ land_sim_name = "bucket"
...
+if all(key -> !isnothing(config_dict[String(key)]), component_dt_names)
+    haskey(config_dict, "dt") && @warn "Ignoring dt in favor of individual component dt's"
+    for key in component_dt_names
+        component_dt = Float64(time_to_seconds(config_dict[String(key)]))
+        @assert Δt_cpl % component_dt == 0.0 "Coupler dt must be divisible by all component dt's\n dt_cpl = $Δt_cpl\n $key = $component_dt"
+        eval(:($key = $component_dt))
```
juliasloan25, line 174: This is clever! Very nice

experiments/ClimaEarth/run_amip.jl
```diff
@@ -166,8 +163,31 @@ land_sim_name = "bucket"
...
+    end
+    # the Atmos `get_simulation` function expects the atmos config to contain a value dt,
+    # which describes it simulation timestep, so we set it here
+    config_dict["dt"] = config_dict["dt_atmos"]
```
juliasloan25, line 178: It would be better to isolate this workaround to the atmosphere config dict. Could you move this to [get_atmos_config_dict](https://github.com/CliMA/ClimaCoupler.jl/blob/91878a67ec31321f198b5804ad3fe25720c44480/experiments/ClimaEarth/components/atmosphere/climaatmos.jl#L304)? You can use similar logic in there to check if the coupler config dict has `dt` or `dt_atmos`, and choose which to use.

I realized that if `dt_atmos` is specified but another component model dt isn't specified, we could have both `dt` and `dt_atmos` entries in the coupler config, and then it would be repetitive logic to choose which to use in the atmos file. To avoid that, could you also remove any entry in the config dict that gets ignored? I.e. anything that produces a warning in line 183

juliasloan25 (commented):
This is great, thank you Teja! Just a couple of comments

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -304,6 +304,12 @@ coupled/atmos-only comparisons.
 function get_atmos_config_dict(coupler_dict::Dict, job_id::String)
     atmos_config_file = coupler_dict["atmos_config_file"]
     atmos_config_repo = coupler_dict["atmos_config_repo"]
+    # the Atmos `get_simulation` function expects the atmos config to contains its timestep size
+    # in the `dt` field. If there is a `dt_atmos` field in coupler_dict, we temporarly change
```
juliasloan25, line 308: typo: "temporarly" -> "temporarily" (same in line 362)

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -351,7 +357,12 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String)
     # merge configs
     # (if there are common keys, the last dictionary in the `merge` arguments takes precedence)
     atmos_config = merge(atmos_config, Dict("output_dir" => atmos_output_dir))
```
juliasloan25, line 359:
```suggestion
    dt_atmos = haskey(coupler_dict, "dt_atmos") ? coupler_dict["dt_atmos"] : coupler_dict["dt"]
    atmos_config = merge(atmos_config, Dict("output_dir" => atmos_output_dir), "dt" => dt_atmos)
```
Instead of modifying the entry `coupler_dict["dt"]` above and below here, we can find whichever entry exists in `coupler_dict`, and add the `"dt"` entry to `atmos_config` as we create it.

juliasloan25 (approved):
Thanks Teja, this is great! It would be good to move this dt processing into its own function to isolate different parts of the setup, but let's revisit that next as we work on cleaning up the driver.

imreddyTeja:
> Is there a description somewhere on how stepping with different dts work?

No. There is a description of how the dt and dt_component fields are dealt with when both exist [here](https://clima.github.io/ClimaCoupler.jl/dev/generated/amip/run_amip/). 

Do you mean there should be a description somewhere of how the each component steps performs multiple steps until the total stepped time is equivalent to the coupler_dt?

juliasloan25:
> My current concern is with restarts. Two of the things I am interested in are the order of operations and the constraints on the allowed dts (e.g., do they all have to be integer dividers of dt_cpl?).

Yes, we check that each dt is a divider of dt_cpl [here](https://github.com/CliMA/ClimaCoupler.jl/pull/1059/files#diff-865c9e725db93a3a98aca38df51a1d6cdc08bb367832143022264fafa11e97a8R177). What do you mean by order of operations?

We can add more documentation about how to run using just `dt` or specifying `dt_component` for each component. Right now this is all just in the driver but as we refactor it will be moved elsewhere

imreddyTeja:
> My current concern is with restarts. Two of the things I am interested in are the order of operations and the constraints on the allowed dts (e.g., do they all have to be integer dividers of dt_cpl?).

- yes, they do have to be integers dividers of dt_cpl. If not, an error stating that is thrown. This constraint is imposed by us, not SciMLBase. The SciMLBase step function has functionality that supports forcing any timestep size to step exactly to a target time, even if the time step size does not perfectly divide the target dt.   
- For order of operations, are you referring to the order in which the time steps occur? If there is no parallelization, it seems like one components performs all of its steps, then the next components performs all of its steps, and so on.. Then the coupling happens and it repeats.

Sbozzolo:
Thank you!

> If there is no parallelization, it seems like one components performs all of its steps, then the next components performs all of its steps, and so on.. Then the coupling happens and it repeats.

Yes, this is what I was looking for.

PR #1062, opened 2024-11-05
Remove config args which are no longer used in ClimaAtmos

akshaysridhar:


szy21:
One more here: https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/test/amip_test.yml

szy21 (approved):

Issue #1065, opened 2024-11-06
Coupler interface planning (WIP)


juliasloan25:
Brainstorming/ideas to work towards a clean coupling interface

### Changes to make
- Add `examples/` folder with stable, clean simulation setups
- Config file should be optional
  - Maintain support for it, but also add examples without it
- Specify space in ClimaCoupler drivers
  - boundary space or full space?
- Move prescribed data (CO2, SST, SIC) into relevant component models (atmos, ocean, sea ice) rather than having them as independent entities in `CoupledSimulation.mode`

### Questions
- Should coupler specify just the boundary (surface) space, or all spaces (above surface, surface, subsurface)? Either way, need consistency with topography
- Do we want to continue requiring 4 component models (atmos, ocean, land, ice), or allow 1 atmos + 1-3 surface models?
  - I think it's better to allow `nothing` for a surface model, rather than requiring it's defined and zeroing out the contributions/area fraction

PR #1068, opened 2024-11-07
sketch out driver interfaces (WIP)

juliasloan25:
## Purpose 
We want to work on cleaning up how we set up a simulation in ClimaCoupler. This PR demonstrates some potential interfaces we can work towards with that goal in mind. This is a work in progress and will be continually iterated on.

more thoughts in #1065

## Content

----


juliasloan25:
We've moved towards this design, and can come back to it when we want to make more improvements

PR #1071, opened 2024-11-08
Refactor independent time step argument reading

imreddyTeja:
The code that read the component dts used eval
and symbols to create the component dt vars.

This is changed to keep them in a dict. dt_atmos is held in this dict, but it is not actually used. I kept it there for consistency, but it could be removed.

## Purpose 
Remove usage of eval to create vars based on config/command line arguments in run_amip.

## Content
The component dts are now held in a dict

----


Sbozzolo (approved):
Thank you!

PR #1072, opened 2024-11-08
add downgrade test

### LINKED ISSUE #1021: add downgrade test to CI
e.g. see ClimaLand downgrade test [here](https://github.com/CliMA/ClimaLand.jl/blob/main/.github/workflows/downgrade.yml)

juliasloan25:
## Purpose 
This PR adds a github action to run a downgrade test. Passing this test will not be required to merge PRs. This PR also adds compatibility entries for `LinearAlgebra`, which we didn't have before.

closes #1021


Sbozzolo:
Added in #1140

PR #1085, opened 2024-11-15
don't soft-fail AMIP

juliasloan25:
## Purpose 
We want to be able to clearly see when AMIP fails


szy21 (approved):
Let's face the failure :)

PR #1086, opened 2024-11-16
Isolate config and argument parsing

### LINKED ISSUE #678: Isolate argument parsing and configuration setup
It would clean up our driver if we have a function that takes in the config file and extracts all the type information and parameters from it in one place.

We should also have a clear separation between the coupler config dict and the atmos config dict, and remove the merge of them we currently have.

juliasloan25:
## Purpose 
closes #678

This PR isolates the argument and configuration file parsing to functions. Since we have one configuration file each for the coupled simulation and for the atmosphere, we end up with the following functions, as well as smaller helper functions. These are located in the new `experiments/ClimaEarth/arg_parsing.jl file`.
- `get_coupler_config`, `get_coupler_args`, `get_atmos_args`

These functions are called from the driver `run_amip.jl`, where the returned arguments are unpacked. After that point, the config files are not accessed further down in the driver, so we have isolated the config file access to the initial step.

Many files are changed in this PR because argument changes are propagated to many config files. The main files to look at for review are: `run_amip.jl`, `arg_parsing.jl`, `cli_options.jl`, and `climaatmos.jl`.

Note that this PR introduces some very minor behavioral changes in buildkite runs because the atmos config is no longer added to the coupler config. This will allow us to have better separation of and control over the two in the future.

## To-do
- [x] Add function to read in config file and job id
- [x] Add function to parse input arguments (from config file)
  - Separate functions for coupler and atmos?
- [x] Update atmos config function to not modify coupler (make these two independent)
- [x] Unify `argparse` and `get_args` functions - want same parameters available from command line and from config files
- [x] Allow t_start as an argument
- [x] Unify `dt_cpl` to use same format as all other times
- [x] Change `restart_dir` default to `nothing`
- [x] Rename "diagnostics" -> "extra_atmos_diagnostics" for clarity
- [x] Check CI outputs

## Notes
- Separate PR: Remove atmos-related args from coupler configs, and move them into atmos configs
  - extra_atmos_diagnostics, EDMF-related arguments
  - This PR will become very large if we do this here, and I'd prefer to change that logic in an isolated PR


config/ci_configs/amip_coarse_ft32.yml
```diff
@@ -1,7 +1,5 @@
 FLOAT_TYPE: "Float32"
 apply_limiter: false
-dt: "400secs"
```
juliasloan25, line 3: Removed because `400secs` is the default for `dt` and `dt_cpl`

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+If all component timesteps `dt_\$component` are specified in the config file, use those.
+Otherwise, use the generic component timestep `dt` specified in the config file.
+
+# Arguments
+- `config_dict`: A dictionary mapping configuration keys to the specified settings
+
+# Returns
+- `component_dt_dict`: A dictionary mapping component model names to their respective timesteps
+"""
+function parse_component_dts(config_dict)
```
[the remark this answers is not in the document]
imreddyTeja: Maybe we don't need to modify the `config_dict` here. Removing unused dts isn't really needed because the config dict is only used again to create the `atmos_config_dict`.  

If we don't remove unused dts, then the only case we would change the `config_dict` is when `dt_atmos` is specified but not all the other components are. Instead of reading dt_atmos/dt from the `config_dict` in `get_atmos_config_dict`, we could pass in `component_dt_dict`, or just `dt_atmos` to `get_atmos_config_dict`, and then overwrite `dt_atmos` in `atmos_config_dict` after merging with `dt_config`.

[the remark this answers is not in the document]
juliasloan25: I thought about this a bit, and here's what I came up with

> If we don't remove unused dts, then the only case we would change the config_dict is when dt_atmos is specified but not all the other components are.

I agree, but we should avoid thinking about `dt_atmos` as a special case. For example, we might want to set up land with a config dict in the future, in which case all the same things would apply to `dt_land`.

> Instead of reading dt_atmos/dt from the config_dict in get_atmos_config_dict, we could pass in component_dt_dict, or just dt_atmos to get_atmos_config_dict, and then overwrite dt_atmos in atmos_config_dict after merging with dt_config.

We could do this, but we should avoid treating certain arguments differently from the others. Right now the only things we pass into `get_atmos_config_dict` are `coupler_dict` and `job_id` (which is special because we read it in from the command line rather than from the dict)

To me, it seems like the best options are:
1. Add `component_dt_dict` to the `config_dict` after we finish choosing which dts to use. Then we don't have to remove the unused keys from the dict, and it will be clear to the component models that they should use whichever dt appears in the `component_dt_dict`. However, I don't like that this solution changes the structure of keys in the config dictionary at different points (e.g. when parsing dts for the coupler vs for the atmos/land), and I don't like that this keeps keys in the config dicts that are unused
2. Clean up the dts when we parse the config file (as is done now). As I think about it more, I think it's okay to have a "cleanup" phase when parsing the config. We could extend this to check consistency more generally, e.g. that EDMF parameters are not in the config if EDMF is not being used.

Right now I prefer 2, but let me know what you think. We can also discuss in person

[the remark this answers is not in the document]
imreddyTeja: 2 seems good to me, especially if there are more keys that could potentially be removed. I think adding new keys into `config_dict` (what would happen in option 1 if component_dts are not specified) could be confusing later on, but removing keys is probably not as confusing.

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,136 @@ import ArgParse
...
-        help = " Ocean simulation time step"
+        ### ClimaCoupler flags
+        # Simulation-identifying information
+        "--config_file"
+        help = "[REQUIRED] A yaml file used to set the configuration of the coupled model"
         arg_type = String
-        "--dt_seaice"
-        help = " Sea Ice simulation time step"
+        "--job_id"
+        help = "[REQUIRED] A unique identifier for this run"
```
Sbozzolo, line 11: Why is this required? We can always make one up if the user doesn't provide it.

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,136 @@ import ArgParse
...
+        help = "Time interval for saving output [\"10days\" (default); allowed formats: \"Nsecs\", \"Nmins\", \"Nhours\", \"Ndays\", \"Inf\"]"
+        arg_type = String
+        default = "10days"
+        # Checkpointing information
         "--hourly_checkpoint"
-        help = "Boolean flag indicating whether to checkpoint at intervals of 1 hour or multiple hours"
+        help = "Boolean flag indicating whether to checkpoint at multiple-hourly intervals [false (default), true]"
         arg_type = Bool
         default = false
         "--hourly_checkpoint_dt"
```
Sbozzolo, line 79: Could you rename this to just `checkpoint_dt` and take the same time format as the other time intervals?

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,136 @@ import ArgParse
...
+        help = "Boolean flag indicating whether to soft fail on conservation errors [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
+        # Output information
+        "--coupler_output_dir"
+        help = "Directory to save output files. Note that TempestRemap fails if interactive and paths are too long. [\"experiments/ClimaEarth/output\" (default)]"
+        arg_type = String
+        default = "experiments/ClimaEarth/output"
+        "--ci_plots"
+        help = "Boolean flag indicating whether to make CI plots [`false` (default), `true`]"
```
Sbozzolo, line 125: In the future, we should remove all the references to CI. Users don't care about our CI. Instead, we should treat our CI as a user itself.

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -296,10 +296,19 @@ FluxCalculator.get_surface_params(sim::ClimaAtmosSimulation) = CAP.surface_fluxe
...
+
+The `atmos_config_repo` flag allows us to use a configuration specified within
+either the ClimaCoupler or ClimaAtmos repos, which is useful for direct
 coupled/atmos-only comparisons.
+
+In this function, parameters are overwritten in a specific order, from lowest to highest priority:
+    1. Default atmos config
+    2. Provided atmos config file (if any)
+    3. Provided atmos TOML file (if any)
+    4. Provided coupler TOML file (if any)
```
Sbozzolo, line 310: Why are we mixing the TOML files in here? Aren't they only used for parameters?

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,252 @@
...
+    ## coupler simulation default configuration
+    parsed_args = parse_commandline(argparse_settings())
+
+    ## modify parsed args for fast testing from REPL #hide
+    if isinteractive()
+        parsed_args["config_file"] =
+            isnothing(parsed_args["config_file"]) ? joinpath(pkg_dir, "config/ci_configs/interactive_debug.yml") :
+            parsed_args["config_file"]
+        parsed_args["job_id"] = "interactive_debug"
+    end
```
Sbozzolo, line 22: I think we should remove this. We can provide a way `debug_config` function if there's such a need

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,252 @@
...
+# Arguments
+- `config_dict`: A dictionary mapping configuration keys to the specified settings
+
+# Returns
+- All arguments needed for the coupled simulation
+"""
+function get_coupler_args!(config_dict::Dict)
+    # Simulation-identifying information; Print `config_dict` if requested
+    config_dict["print_config_dict"] && @info(config_dict)
+    mode_name = config_dict["mode_name"]
```
Sbozzolo, line 50: Here's a useful pattern for keyword dispatch.

```julia

"Type describing what to do with coupling"
abstract type AbstractCouplingMode end

"XYZ"
struct AMIP <: AbstractCouplingMode end

"XYZ"
struct Slabplanet <: AbstractCouplingMode end

# In the function 
allowed_modes = Dict(lowercase(string(nameof(t))) => t for t in subtypes(AbstractCouplingMode))
mode_name = lowercase(config_dict["mode_name"])

if haskey(allowed_modes, mode_name)
  coupling_mode = allowed_modes[mode_name]() 
else
  error("$mode_name is not an allowed coupling mode. Allowed modes are: $(keys(allowed_modes))")
end

# Then, you dispatch instead of doing `if-else`
```

The advantages of this pattern:
- Supports lower/upper cases
- Provides a useful error message
- Automatically extends to new subtypes, including subtypes that a user might define
- Allows us and users to dispatch
- Automatically catches invalid modes

Note, however, that these are probably different experiments, not really `CouplingModes`, so the concept is probably ill-posed to begin with.

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,252 @@
...
+    random_seed = config_dict["unique_seed"] ? time_ns() : 1234
+    FT = config_dict["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
+    comms_ctx = Utilities.get_comms_context(config_dict)
+
+    # Time information
+    t_end = Float64(Utilities.time_to_seconds(config_dict["t_end"]))
+    t_start = Float64(Utilities.time_to_seconds(config_dict["t_start"]))
+    date0 = date = Dates.DateTime(config_dict["start_date"], Dates.dateformat"yyyymmdd")
+    Δt_cpl = Float64(Utilities.time_to_seconds(config_dict["dt_cpl"]))
+    component_dt_dict = parse_component_dts!(config_dict)
```
Sbozzolo, line 62: I'd advise against having mutating functions that return something. It's harder to know what we should expect from them. Either they return something, or they mutate the argument. If you want the return value, you can grab it in a second function that only grabs the return value

experiments/ClimaEarth/user_io/arg_parsing.jl
juliasloan25: Agreed. We can just add default values for `config_file` and `job_id`, that will get used if we run interactively without specifying alternatives

experiments/ClimaEarth/cli_options.jl
juliasloan25: Yes and I would like to generalize this, but it requires some changes in `TimeManager` since the `HourlyCallback` and `MonthlyCallback` are hardcoded to take in number of hours/months respectively. It won't be difficult, but I'd rather do it in a follow-up PR to keep this one from getting too large

experiments/ClimaEarth/cli_options.jl
juliasloan25: I renamed everything to do with CI plots to diagnostics plots. Now the only remaining references to CI are int he `ci_configs/` folder, which we can address in the future.

experiments/ClimaEarth/user_io/arg_parsing.jl
juliasloan25: I like this idea. Since we're planning to separate the driver based on simulation type soon, I don't want to invest too much time in setting this up. But as we start working on that, we could use this structure

experiments/ClimaEarth/cli_options.jl
juliasloan25: I opened an issue for it here https://github.com/CliMA/ClimaCoupler.jl/issues/1099

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,137 @@ import ArgParse
...
         arg_type = String
-        default = "CombinedStateFluxesMOST"
+        default = "Float64"
+        "--device"
+        help = "Device type to use [\"auto\" (default), \"CPUSingleThreaded\", \"CPUMultiThreaded\", \"CUDADevice\"]"
+        arg_type = String
+        default = "auto"
+        # Time information
+        "--t_end"
+        help = "[REQUIRED] End time of the simulation (no default) [allowed formats: \"Nsecs\", \"Nmins\", \"Nhours\", \"Ndays\", \"Inf\"]"
```
Sbozzolo, line 41: Maybe the default could just be t_end = t_start

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+    config_dict = merge(parsed_args, YAML.load_file(parsed_args["config_file"]))
+    config_dict["job_id"] = job_id
+    return config_dict
+end
+
+"""
+    get_coupler_args!(config_dict)
+
+Extract the necessary arguments from the coupled configuration dictionary.
+This function may modify the input dictionary to remove unnecessary keys.
```
Sbozzolo, line 31: Let's split this into two: a function that removes the unnecessary keys, and one that returns the the coupler configs. 

Having mutating functions that return stuff adds complexity that can be easily avoided.

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
+"""
+    get_coupler_config()
+
+Reads in the configuration file and job ID from the command line.
```
Sbozzolo, line 4:
```suggestion
Read in the configuration file and job ID from the command line.
```

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+A dictionary is constructed from the input configuration file, and this
+is returned along with the job ID.
```
Sbozzolo, line 6: This function now only returns `config_dict`

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+"""
+function get_coupler_config()
+    # Read in command line arguments
+    parsed_args = parse_commandline(argparse_settings())
+
+    # Extract the configuration file and job ID
+    config_file = parsed_args["config_file"]
+    job_id = parsed_args["job_id"]
+    # Get the job ID from the config file string if not provided
+    job_id = isnothing(job_id) ? string(split(split(config_file, '/')[end], '.')[1]) : job_id
```
Sbozzolo, line 19: Just note that that this function would fail if my config_file were `myconfig` (without the extension).

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+- 10 days for simulations longer than 30 days
+- 1 day for simulations longer than 1 day
+- 1 hour for simulations shorter than 1 day
+
+# Arguments
+- `t_start`: The start time of the simulation
+- `t_end`: The end time of the simulation
+
+# Returns
+- `period`: A String of how often to average and output diagnostics
```
Sbozzolo, line 174: Is `period` needed?

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,248 @@
...
+    # check if all component dt's are specified
+    if all(key -> !isnothing(config_dict[key]), component_dt_names)
+        # when all component dt's are specified, ignore the dt field
+        if haskey(config_dict, "dt")
+            @warn "Removing dt in favor of individual component dt's"
+            delete!(config_dict, "dt")
+        end
+        for key in component_dt_names
+            component_dt = Float64(Utilities.time_to_seconds(config_dict[key]))
+            @assert Δt_cpl % component_dt == 0.0 "Coupler dt must be divisible by all component dt's\n dt_cpl = $Δt_cpl\n $key = $component_dt"
```
Sbozzolo, line 233: It's typically not a good idea to do exact comparison involving floating points. For example, 1e-300 != 0.0 so this would fail

Sbozzolo (commented):
This is a good step in the direction of isolating config/argument parsing. The next step would be to make sure that we don't use the parsed_args anywhere after the initial parsing. This would entail making everything types/symbols and ensuring that we are not matching strings  after the initial parsing

juliasloan25:
> This is a good step in the direction of isolating config/argument parsing. The next step would be to make sure that we don't use the parsed_args anywhere after the initial parsing. This would entail making everything types/symbols and ensuring that we are not matching strings after the initial parsing

If we did this, wouldn't we still use the args we've read in after the initial parsing? We would just use them for dispatch instead of matching strings

experiments/ClimaEarth/cli_options.jl
juliasloan25: I'd rather make it 2 * default dt = 800secs, since we often run simulations for 2 steps to check interfaces

experiments/ClimaEarth/user_io/arg_parsing.jl
Sbozzolo: Okay, thank you!

experiments/ClimaEarth/cli_options.jl
Sbozzolo: Okay!

juliasloan25:
> > > This is a good step in the direction of isolating config/argument parsing. The next step would be to make sure that we don't use the parsed_args anywhere after the initial parsing. This would entail making everything types/symbols and ensuring that we are not matching strings after the initial parsing
> > 
> > 
> > If we did this, wouldn't we still use the args we've read in after the initial parsing? We would just use them for dispatch instead of matching strings
> 
> Yes, but this would allow users to run without any argument parsing at all, just setting up the correct types (and only the relevant types, as we might want to do in a CI context). It would also allow users to extend the behavior of functions with their custom types, whereas it is not possible to extend string-matching without changing the source code.

I see! That would be nice for things like the turbulent flux partition, albedo to use, etc, but I think we'll still need to parse user inputs for things like dt, t_end, output directory, etc

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,138 @@ import ArgParse
...
-        default = "experiments/ClimaEarth/output"
+        # Restart information
         "--restart_dir"
         help = "Directory containing restart files"
         arg_type = String
-        default = "unspecified"
+        default = nothing
         "--restart_t"
-        help = "Restart time"
+        help = "Integer time to use at `t_start` for restarted simulation [0 (default)]"
```
Sbozzolo, line 91:
```suggestion
        help = "Time in seconds rounded to the nearest index to use at `t_start` for restarted simulation [0 (default)]"
```

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,270 @@
...
+        for key in component_dt_names
+            if !isnothing(config_dict[key])
+                @warn "Removing $key from config in favor of dt because not all component dt's are specified"
+            end
+            delete!(config_dict, key)
+            component_dt_dict[key] = Float64(Utilities.time_to_seconds(config_dict["dt"]))
+        end
+    end
+    config_dict["component_dt_dict"] = component_dt_dict
+end
```
Sbozzolo, line 233:
```suggestion
    config_dict["component_dt_dict"] = component_dt_dict
    return nothing
end
```

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -0,0 +1,270 @@
...
+
+        # Additional atmosphere diagnostics
+        !haskey(config_dict, "extra_atmos_diagnostics") &&
+            (config_dict["extra_atmos_diagnostics"] = Vector{Dict{Any, Any}}())
+        push!(
+            config_dict["extra_atmos_diagnostics"],
+            Dict("short_name" => ["toa_fluxes_net"], "reduction_time" => "average", "period" => period),
+        )
+    end
+    config_dict["calendar_dt"] = calendar_dt
```
Sbozzolo, line 269:
```suggestion
    config_dict["calendar_dt"] = calendar_dt
    return nothing
```

config/longrun_configs/slabplanet_aqua_target_nocouple.yml
```diff
@@ -1,7 +1,7 @@
 atmos_config_file: "config/longrun_configs/longrun_aquaplanet_allsky_0M.yml"
 dt: "120secs"
 dt_cloud_fraction: "1hours"
-dt_cpl: 10368000
+dt_cpl: "10368000secs"
```
Sbozzolo, line 4:
```suggestion
dt_cpl: "120days"
```

experiments/ClimaEarth/cli_options.jl
```diff
@@ -2,107 +2,138 @@ import ArgParse
...
+        default = false
+        "--conservation_softfail"
+        help = "Boolean flag indicating whether to soft fail on conservation errors [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
+        # Output information
+        "--coupler_output_dir"
+        help = "Directory to save output files. Note that TempestRemap fails if interactive and paths are too long. [\"experiments/ClimaEarth/output\" (default)]"
+        arg_type = String
+        default = "experiments/ClimaEarth/output"
```
Sbozzolo, line 125: This would be a little weird if you ran from `experiments/ClimaEarth/` (it would create another layer of `experiments/ClimaEarth/`), but I think it's fine for the time being

Sbozzolo (approved):

PR #953, opened 2024-09-13
remove Diagnostics module

juliasloan25:
## Purpose 
Remove Diagnostics module in favor of ClimaDiagnostics and ClimaAnalysis

closes #954

will supersede https://github.com/CliMA/ClimaCoupler.jl/issues/875 and https://github.com/CliMA/ClimaCoupler.jl/issues/517

## Content
This PR touches a lot of files, but most of it is just propagating a few main changes:
- delete the ClimaCoupler Diagnostics module, tests, docs
- use ClimaAtmos.Diagnostics to output atmos diagnostics, and ClimaDiagnostics to output coupler diagnostics
  - coupler diagnostics are defined in `user_io/amip_diagnostics.jl`
  - the diagnostics handler is added to our `CoupledSimulation` object, which results in changes in many files
- add default AMIP diagnostics in driver
- `user_io/ci_plots.jl` is generalized to meet the needs of our updated diagnostic plotting
- artifacts are now output to a subdirectory of the run name, i.e. `run_name/artifacts/`, rather than an adjacent directory `run_name_artifacts/`
  - this isn't necessary for the other changes of the PR, but I found that it made more sense as I was looking at run outputs

## To-do
- [x] generalize `make_ci_plots` to work for AMIP plots as well; update amip_paperplots function to use this
- [x] recreate AMIP paperplots using existing ClimaAtmos diagnostics defined in https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/diagnostics/core_diagnostics.jl
  - coupler symbol -> atmos symbol:
  - ":T" -> "ta"
  - ":u" -> "ua"
  - ":q_tot" -> "hus"
  - ":q_liq_ice" -> "clw"
  - ":precipitation_rate" -> "pr"
  - ":T_sfc" -> "ts"
- [x] remove output_short_name, output_long_name
- [x] for the few quantities that aren't already computed in ClimaAtmos diagnostics, add new variables in `climaatmos_extra_diags.jl`
  - ":toa_fluxes": sum fluxes that can be accessed from atmos cache
  - ":turbulent_energy_fluxes": access calculated value from CoupledSimulation (requires extending ClimaDiagnostics functions)
- [x] remove `amip_visualizer.jl`
- [x] remove diagnostics from CoupledSimulation object
- [x] remove `use_coupler_diagnostics` flag
- [x] remove Diagnostics module and tests

### Output checks
[...]

PR #1096, opened 2024-11-21
Change artifact path for heat-diffusion on buildkite

### LINKED ISSUE #1095: Artifacts from the heat_diffusion experiment not saved on buildkite
The path for artifact in the pipeline.yml and the path the plots are saved to in heat_diffusion/run.jl do not match.

imreddyTeja:
## Purpose 
Save the output plots of the heat-diffusion experiment on buildkite.
closes #1095 

## Content
- changed artifact path to match output path
- Now build kite heat-diffusion shows 3 saved artifacts vs 0 before

----

### DIFF
--- a/experiments/ClimaCore/heat-diffusion/run.jl
+++ b/experiments/ClimaCore/heat-diffusion/run.jl
@@ -304,7 +304,7 @@ ENV["GKSwstype"] = "nul"
 import Plots
 Plots.GRBackend()
 
-ARTIFACTS_DIR = joinpath("experiments/ClimaCore/output/heat-diffusion_artifacts")
+ARTIFACTS_DIR = joinpath("experiments/ClimaCore/output/heat-diffusion/artifacts")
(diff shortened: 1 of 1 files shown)


juliasloan25:
Good catch! I missed the heat diffusion example when I changed the artifact paths in https://github.com/CliMA/ClimaCoupler.jl/pull/953/commits/e4bbc3008ab4a6fd0ea91bf184f589070d930849. I changed the paths so the artifacts are in a subdirectory of the directory created for the job, rather than another directory at the same level (e.g. `amip_default_artifacts/` became `amip_default/artifacts/`. To keep the heat diffusion example consistent with everything else, we should update the dir used in the example ([here](https://github.com/CliMA/ClimaCoupler.jl/blob/e2f30e852f395c5178d1c8e876b5a2d0bce30341/experiments/ClimaCore/heat-diffusion/run.jl#L307)), rather than in buildkite.

juliasloan25 (approved):

PR #1097, opened 2024-11-22
Replace Plots usage with Makie

### LINKED ISSUE #1091: Remove Plots.jl dependency
Our codebase currently uses both Makie.jl and Plots.jl. It would be good to converge to using only Makie.jl and remove the extra Plots.jl dependency, to decrease our overall number of dependencies and decrease precompilation time.

This will only decrease our precompilation time after upstream packages remove Plots.jl too, e.g. in https://github.com/CliMA/ClimaLand.jl/issues/927

imreddyTeja:
All uses of Plots and ClimaCorePlots are removed in this commit. For almost all of the generated plots, the Makie version is visually similar to the Plots.jl versions.

Due to a limitation of Makie, the sea_breeze output animations are slightly different. The Plots.jl animations had a colorbar and levels that rescaled each frame. This commit makes the animations use a consistent colorbar and levels.

Plots and ClimaCorePlots are removed from Project.toml of both experiments/ClimaEarth and experiments/ClimaCore.

Some of the contour plots look slightly different due to differences in Makie and Plots.jl graphics, and because the default behavior of ClimaCoreMakie is to use viridis.

## Purpose 
Closes #1091

## To-do
- update manifests?
- double check build kite artifacts vs an old buildkite-ci run

## Content

----


test/Project.toml
```diff
@@ -8,6 +8,7 @@ ClimaAnalysis = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
 ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaCoreMakie = "908f55d8-4145-4867-9c14-5dad1a479e4d"
 ClimaCorePlots = "cf7c7e5a-b407-4c48-9047-11a94a308626"
```
juliasloan25, line 12: Can we remove ClimaCorePlots and Plots from the test environment now too?

juliasloan25 (approved):
The plots look great! Thank you Teja

experiments/ClimaCore/heat-diffusion/run.jl
```diff
@@ -301,8 +301,8 @@ sol_atm, sol_lnd = integ_atm.sol, integ_lnd.sol;
 # `parent(sol_atm.u[<time-index>].atm_F_sfc)[<z-index>,<variable-index>]`
 
 ENV["GKSwstype"] = "nul"
-import Plots
-Plots.GRBackend()
+import Makie
+import CairoMakie
 
 ARTIFACTS_DIR = joinpath("experiments/ClimaCore/output/heat-diffusion_artifacts")
```
juliasloan25, line 307: There are no artifacts on buildkite for the heat diffusion run because this should be `"/heat-diffusion/artifacts"`, but after rebasing on main it should be fine

test/Project.toml
juliasloan25: I see that we have some commented-out plotting lines in regridder_tests.jl. It's fine to remove those lines since they're not being used and we want to switch to ClimaUtilities regridding soon anyway. That might be the only place we use Plots/ClimaCorePlots in test/

PR #1103, opened 2024-12-02
Remove unused direct dependencies

### LINKED ISSUE #1075: Identify and document removable dependencies
We want to go through the whole ClimaCoupler package and identify dependencies that we can easily remove. This will decrease compile time, and make sure that we're actually using all the dependencies we pull in.

### Environments to go through
- top-level Project
- experiments/ClimaEarth
- test
- experiments/ClimaCore
- docs

Part of https://github.com/CliMA/ClimaCoupler.jl/issues/1050

(see comments for dependencies we can remove)

imreddyTeja:
## Purpose 

Closes #1075 

## Content

----


experiments/ClimaEarth/run_amip.jl
```diff
@@ -55,7 +55,6 @@ import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
 import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput, evaluate!
 import ClimaUtilities.Utils: period_to_seconds_float
 import ClimaUtilities.ClimaArtifacts: @clima_artifact
-import Interpolations
```
Sbozzolo, line 58: Interpolations.jl is used because it triggers an extension loading in ClimaUtilities. We can probably get away with removing it because it is triggered by other dependencies as well, but maybe we should not

juliasloan25: We should keep it and add a comment noting why we need the import

test/Project.toml
```diff
@@ -1,27 +1,20 @@
...
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
-ClimaAnalysis = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
 ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
 ClimaCoreMakie = "908f55d8-4145-4867-9c14-5dad1a479e4d"
 ClimaCoupler = "4ade58fe-a8da-486c-bd89-46df092ec0c7"
-ClimaLand = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
 ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
 ClimaTimeSteppers = "595c0a79-7f3d-439a-bc5a-b232dc3bde79"
```
juliasloan25, line 12: From your comment in #1075, it looks like ClimaTimeSteppers, IntervalSets, LinearAlgebra, SciMLBase, and Statistics can also be removed. These are indirect dependencies, so if you're planning to do them in a separate PR that's fine, but I see that this PR removes the indirect dependencies in the `experiments/ClimaEarth/` environment, so maybe you could do them all at once. Whichever is easiest for you!

juliasloan25 (approved):
Thanks Teja!

test/Project.toml
imreddyTeja: My comment in #1075 was incorrect about a few things. 

For `experiments/ClimaEarth`, Interpolations.jl cannot be removed as a direct dependency because it is used once in `plot_helper.jl`. 

For `test`, ClimaTimeSteppers, IntervalSets, LinearAlgebra, SciMLBase, and Statistics cannot be removed as direct dependencies, because they are imported in scripts in `experiments/ClimaEarth`, which are then `include`-ed in tests in `test`.

I've updated the comment reflect this

Sbozzolo: IntervalSets can be often removed with small changes

experiments/ClimaEarth/run_amip.jl
imreddyTeja: Comment added

test/Project.toml
imreddyTeja: IntervalSets now removed.

PR #1109, opened 2024-12-05
Remove regridding funcs that are in ClimaUtilities

### LINKED ISSUE #425: Use ClimaUtilities.jl Regridder module
Switch from using ClimaCoupler.jl Regridder module to using shared ClimaUtilities.jl Regridder.

tracked by https://github.com/CliMA/ClimaUtilities.jl/issues/39
closed by https://github.com/CliMA/ClimaCoupler.jl/pull/1109

## To Do
- [x] Delete ClimaCoupler.jl Regridder module
- [x] ~~Add ClimaCoupler.jl SurfaceCombiner module~~ #422 
- [x] Update changed function usages
- [x] Update files using ClimaCoupler.jl Regridder to load ClimaUtilities.jl instead
- [x] Test

### LINKED ISSUE #1107: Remove functions from Regridder.jl that exist in ClimaUtilities.Regridders
Many of the functions in Regridder.jl also are defined in ClimaUtilities.Regridders. These should be removed, and their usage replaced.

imreddyTeja:
Closes #1107 and Closes #425
Regridder module removed. The hdwrite_regridfile_rll_to_cgll only works for 2d spaces in ClimaUtilities, and the function in Regridder.jl also worked for 3d spaced, so this functionality is lost.

~~The functions read_available_dates and yyyymmmddd_to_datetime are copied from ClimaUtilities. This is necessary because land_fraction needs the first date in the dataset. TempestRegridder calls hdwrite_regridfile_rll_to_cgll, which extracts the dates and returns them, but TempestRegridder then discards those extracted dates.~~

The regridder functions in ClimaUtilities do not work properly when the time dim is made up of numbers, and it errors when the time dim is made up of large floats. To function properly, the time time must be datetimes. The functions in Regridder.jl worked with the time dim as numbers.

Many of the regridding functions in Regridder.jl allowed the outfile_root to be set. When using the TempestRegridder, it is autmatically set to the varname. This results in file with names structured like: "varname_varname_time.hdf5"

## To-do

- ~~move Regridder module docs to news.md~~
- make changes in ClimaUtilities (time reading, outfile_root, varname to varnames)
see [here](https://github.com/CliMA/ClimaUtilities.jl/pull/140)
changes to the time reading probably require a small discussion, and making TempestRegridder support multiple varnames requires enough changes to be a separate issue (implementing would change interface, or would need to continue supporting old interface) 
- ~~Docs differentiate internal vs external functions~~
- ~~test for binary mask~~

## Content
- Regridder module removed (also checked that there is equivalent functionality in ClimaUtilities. If there was not, it is noted)
- added to NEWS.md

----


src/FieldExchanger.jl
```diff
@@ -6,10 +6,95 @@ atmospheric and surface component models.
...
+
+Simple stand-in function for remapping.
+For AMIP we don't need regridding of surface model CC.Fields.
+When we do, we re-introduce the ClimaCoreTempestRemap remapping functions.
+
+# Arguments
+- `target`: [CC.Fields.Field] destination of remapping.
+- `source`: [CC.Fields.Field] source of remapping.
+"""
+function dummmy_remap!(target, source)
```
juliasloan25, line 68: This is just a small style choice, but it's good practice to put internal functions at the end of a file, so if a user looks into the file to find a function they're using they can find it more easily.

test/TestHelper.jl
```diff
@@ -118,13 +119,41 @@ function gen_ncdata_time(FT, path, varname, val)
     # Define variables
     lon = NCDatasets.defVar(nc, "lon", FT, ("lon",))
     lat = NCDatasets.defVar(nc, "lat", FT, ("lat",))
-    time = NCDatasets.defVar(nc, "time", Float64, ("time",))
+    time = NCDatasets.defVar(nc, "time", Dates.DateTime.([2020.0, 2021.0]), ("time",))
```
juliasloan25, line 122: why was this change needed?

juliasloan25: it looks like `gen_ncdata_time` and `gen_ncdata_date` were only used in the regridder tests so now they're unused and can be removed

test/field_exchanger_tests.jl
```diff
@@ -56,6 +56,29 @@ Interfacer.get_field(sim::Union{TestSurfaceSimulation2, TestSurfaceSimulation2},
...
+struct TestSurfaceSimulationA <: Interfacer.SurfaceModelSimulation end
+struct TestSurfaceSimulationB <: Interfacer.SurfaceModelSimulation end
+struct TestSurfaceSimulationC <: Interfacer.SurfaceModelSimulation end
+struct TestSurfaceSimulationD <: Interfacer.SurfaceModelSimulation end
+
+# Initialize weights (fractions) and initial values (fields)
+Interfacer.get_field(::TestSurfaceSimulationA, ::Val{:random}) = 1.0
+Interfacer.get_field(::TestSurfaceSimulationB, ::Val{:random}) = 1.0
+Interfacer.get_field(::TestSurfaceSimulationC, ::Val{:random}) = 1.0
+Interfacer.get_field(::TestSurfaceSimulationD, ::Val{:random}) = 1.0
```
juliasloan25, line 68: it would be good for some of these values to be not 1 (like you have for the `area_fraction` functions below)

juliasloan25 (approved):
This looks great, thank you Teja! I just left a few small comments

juliasloan25:
It will be good to make the ClimaUtilities TempestRegridder changes we talked about too, but I'm okay with those going in after this is merged :)

test/TestHelper.jl
imreddyTeja: That change was made because the `read_available_dates` in ClimaUtilities incorrectly interprets the time dimension when it is a Number. The unused `gen_ncdata_time` and `gen_ncdata_date` are now removed.

PR #1094, opened 2024-11-21
Add postprocessing for 4D variables (lat, lon, time, pressure level)

ph-kev:
## Purpose 
closes #1093 - This PR adds postprocessing for 4D variables (lat, lon, time, pressure level). See plots below.

## To-do
- [ ] Make new release of ClimaAnalysis and update here in ClimaCoupler

## Content
- [X] Add bias plots after taking time average and slicing at specific pressure levels (850, 500, 250 hPA)
- [X] Add bias plots over latitude and pressure level after taking average over time and longitude

- [X] I have read and checked the items on the review checklist.
bias_vars_in_pfull
lat_pfull_heatmaps

PR #1114, opened 2024-12-09
Load OrderedDict

Sbozzolo:
PR #1103 and #1094 are inconsistent. This uses OrderedDict directly from ClimaAnalysis


ph-kev (approved):

Sbozzolo:
> Should we just make OrderedCollections an explicit and direct dependency?

I don't think it's really strongly needed at this point. If it becomes more important, we can turn it into a direct dependency

PR #1117, opened 2024-12-12
Switch to PartitionedFluxes

Sbozzolo:
`CombinedStateFluxes`, while simple, is not super physically less sound (combining states of different component models into a single average state is necessarily the best idea). In addition to this, `CombinedStateFluxes` relies on the infrastructure in ClimaAtmos to compute turbulent fluxes, which is highly complex and difficult to understand.

This commit fixes `PartionedStateFluxes` and switches all the jobs to use this system. `PartionedStateFluxes` was broken for certain jobs because it was not setting certain values, such as ustar, which are needed (e.g., in EDMF).

In this commit, we also change how surface temperature is propagated to atmos. Here, we take directly the area-weighted value computed from the surface models and directly set it in atmos.

This PR also fixes a few other things along the way:
- Using `CC.to_cpu` instead of implementing our own
- Fixing Eisenman compatibility with partition fluxes
- Allowing debug plots to plot quantities with NaNs or Infs
- Adding a default `show` method for simulations (otherwise the REPL is overwhelmed with text)

Passing AMIP buildkite: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/318

## Summary by CodeRabbit

- **New Features**
  - Adopted a new default method for computing turbulent fluxes by partitioning them per component. Configuration updates now reflect this change, potentially altering simulation outcomes.
  
- **Documentation**
  - Enhanced documentation with reformatted tables and expanded details on key diagnostic outputs such as friction velocity, Obukhov length, and buoyancy flux.
  
- **Refactor**
  - Streamlined simulation setup, debugging, and testing workflows to align with the updated flux partitioning approach.


NEWS.md
```diff
@@ -6,7 +6,7 @@ ClimaCoupler.jl Release Notes
 
 ### ClimaCoupler features
```
juliasloan25, line 7: can you also add an entry for this change to PartitionedSurfaceFluxes please?

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -336,6 +344,13 @@ function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:turbulent_fl
     parent(sim.integrator.p.precomputed.sfc_conditions.ρ_flux_q_tot) .=
         parent(F_turb_moisture) .* parent(surface_normal) # (evap)
 
+    # If available, also Monin-Obukhov quantities
+    if :L_MO in propertynames(fields)
```
juliasloan25, line 348: I thought Monin Obukhov quantities will always be available when running with PartitionedStateFluxes. Is this to support CombinedStateFluxes, or does PartitionedFluxes also work with `BulkScheme` now?

experiments/ClimaEarth/user_io/debug_plots.jl
```diff
@@ -113,8 +113,15 @@ function debug(cs_fields::CC.Fields.Field, dir, cs_fields_ref = nothing)
...
             # Copy field onto cpu space if necessary
-            cpu_field = to_cpu(field)
-            field_min, field_max = extrema(cpu_field)
+            cpu_field = CC.to_cpu(field)
+
+            # ClimaCoreMakie doesn't support NaNs/Infs, so we substitute them with 100max
+            FT = CC.Spaces.undertype(axes(cpu_field))
+            isinvalid = (x) -> isnan(x) || isinf(x)
+            extrema_with_valid_input = extrema(map(x -> isinvalid(x) ? FT(0) : x, parent(cpu_field)))
+            field_min, field_max = extrema(extrema_with_valid_input)
```
juliasloan25, line 122:
```suggestion
            field_valid_min, field_valid_max = extrema(map(x -> isinvalid(x) ? FT(0) : x, parent(cpu_field)))
```

experiments/ClimaEarth/user_io/debug_plots.jl
```diff
@@ -131,8 +138,16 @@ function debug(cs_fields::CC.Fields.Field, dir, cs_fields_ref = nothing)
                 field_name = field_names[field_index]
                 field = getproperty(cs_fields, field_name)
                 # Copy field onto cpu space if necessary
-                cpu_field = to_cpu(field)
-                field_min, field_max = extrema(cpu_field)
+
```
juliasloan25, line 141:
```suggestion
```

test/flux_calculator_tests.jl
```diff
@@ -155,107 +154,84 @@ for FT in (Float32, Float64)
...
+        windspeed = @. hypot(atmos_sim.integrator.p.u, atmos_sim.integrator.p.v)
+
+        thermo_params = get_thermo_params(atmos_sim)
+        thermo_state_int = Interfacer.get_field(atmos_sim, Val(:thermo_state_int))
+
+        surface_thermo_states = similar(thermo_state_int)
+        surface_thermo_states .= FluxCalculator.surface_thermo_state(ocean_sim, thermo_params, thermo_state_int)
+
+        # NOTE: This test is very weak! We should add more stringent tests
+        @test Array(parent(fields.F_turb_moisture))[1] ≈ FT(0)
```
juliasloan25, line 234: can we add similar checks for the momentum and energy fluxes?

juliasloan25 (commented):
Thank you Gabriele! I left a few comments. Some other notes:

1. Could you add more details to the FluxCalculator doc section about partitioned fluxes? Some things that would be helpful are:
- which quantities from the component models (surface, atmos) are used by the coupler to calculate fluxes with this option? i.e. what does the flux calculation interface look like
- information about the extra monin obukhov quantities that are exchanged if available. These aren't described anywhere else in the code currently, so we should add information about what they represent physically, units, etc. Also note that these are area-weighted when exchanged and explain why this is done, why it isn't ideal

2. As we've discussed, we should also remove the CombinedStateFluxes option entirely. I'm okay with doing most of this in a separate PR to avoid this one becoming very large, but I would like to change `climaatmos_coarse_short.yml` to use partitioned fluxes in this PR. We should open an issue to keep track of that next piece of work. Some (not all) of the places it's still used/involved:
- `experiments/ClimaEarth/test/component_model_tests/climaatmos_coarse_short.yml`
- hierarchy drivers
- FluxCalculator module
- `setup_run.jl`
- other functions we'll be able to remove:  `get_field(atmos_sim, :turbulent_energy_flux)`, `get_new_cache(atmos_sim)`, etc

juliasloan25:
Also note that it looks like we have grid imprinting in the momentum fluxes, seen in the `debug_coupler.png` plots of the default AMIP run [in buildkite](https://buildkite.com/clima/climacoupler-ci/builds/6117#019615db-d572-4b5c-b4d8-b031f78b5b19). We should double check that all fields are continuous - maybe we need to `dss` some of the new fields. It could be related to the `uv_int` change? I'm not sure which of the other changes in this PR would cause that

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
Sbozzolo: I guess that's true, I can just assume they are there.

I was probably tried to get more compatibility with other schemes

docs/src/fluxcalculator.md
```diff
@@ -1,36 +1,53 @@
...
+The key function that computes surface fluxes is
+[`FluxCalculator.partitioned_turbulent_fluxes!`](@ref). This function computes
+turbulent fluxes and ancillary quantities, such as the Obukhov length, using
+[SurfaceFluxes.jl](https://github.com/CliMA/SurfaceFluxes.jl). Generally, this
+function is called at the end of each coupling step.
+
+All the quantities computed in `partitioned_turbulent_fluxes!` are calculated
+separately for each surface type using the
+[`FluxCalculator.compute_surface_fluxes!`](@ref) function. This function can be
+extended by component models if they need specific type of flux calculation. The
```
juliasloan25, line 16:
```suggestion
extended by component models if they need specific type of flux calculation, and
a default is provided for models that can use the standard flux calculation. The
```

docs/src/fluxcalculator.md
```diff
@@ -1,36 +1,53 @@
...
+final result of `partitioned_turbulent_fluxes!` is an area-weighted sum of all
+the contributions of the various surfaces.
+
+[`FluxCalculator.compute_surface_fluxes!`](@ref), in turn, calls
```
juliasloan25, line 20:
```suggestion
The default method of [`FluxCalculator.compute_surface_fluxes!`](@ref), in turn, calls
```

docs/src/fluxcalculator.md
```diff
@@ -1,36 +1,53 @@
...
+[`FluxCalculator.get_surface_fluxes!`](@ref). This function uses a thermal state
+obtained by using the model surface temperature, extrapolates atmospheric
+density adiabatically to the surface, and with the surface humidity (if
+available, if not, assuming a liquid phase). `compute_surface_fluxes!` also
+updates the component internal fluxes fields with
+[`FluxCalculator.update_turbulent_fluxes!`](@ref).
```
juliasloan25, line 26:
```suggestion
available, if not, assuming a liquid phase). `compute_surface_fluxes!`
updates the component model's internal fluxes fields with
[`FluxCalculator.update_turbulent_fluxes!`](@ref), and adds the area-weighted
contribution from this component model to the CoupledSimulation fluxes fields.
Any extension of this function for a particular surface model is also expected to
update both the component models' internal fluxes and the CoupledSimulation
object's fluxes fields.
```

docs/src/interfacer.md
```diff
@@ -138,18 +139,26 @@ This function is expected to be extended for the
...
 - `calculate_surface_air_density(atmos_sim::Interfacer.AtmosModelSimulation, T_sfc::ClimaCore.Fields.Field)`:
 A function to return the air density of the atmosphere simulation
 extrapolated to the surface, with units of [kg m⁻³].
 
+ClimaAtmos should also add the following coupler fields for Monin-Obukhov similarity theory:
+| Coupler name    | Description       | Units  |
+|-----------------+-------------------+--------|
+| `ustar`         | friction velocity | m s⁻¹  |
+| `L_MO`          | Obukhov length    | m      |
+| `buoyancy_flux` | flux of buoyancy  | m⁻²s⁻³ |
```
juliasloan25, line 159: Is this specific to ClimaAtmos, or will any atmosphere model need to provide these? If it's any atmosphere model, we should generalize to `AtmosModelSimulation`

experiments/ClimaEarth/test/component_model_tests/climaatmos_coarse_short.yml
```diff
@@ -20,7 +20,7 @@ prescribe_ozone: true
 rad: allskywithclear
 surface_setup: DefaultMoninObukhov
 t_end: 240secs
-turb_flux_partition: "CombinedStateFluxesMOST"
+turb_flux_partition: "PartitionStateFluxes"
```
juliasloan25, line 23:
```suggestion
turb_flux_partition: "PartitionedStateFluxes"
```

NEWS.md
```diff
@@ -6,7 +6,14 @@ ClimaCoupler.jl Release Notes
...
 
-### Split `setup_and_run` in multiple functions. PR[#1251](https://github.com/CliMA/ClimaCoupler.jl/pull/1251)
+#### Switch to `PartitionStateFluxes` by default. PR[#1117](https://github.com/CliMA/ClimaCoupler.jl/pull/1117)
```
juliasloan25, line 9:
```suggestion
#### Switch to `PartitionedStateFluxes` by default. PR[#1117](https://github.com/CliMA/ClimaCoupler.jl/pull/1117)
```

juliasloan25 (commented):
Mostly just a typo that "PartitionStateFluxes" should be "PartitionedStateFluxes", but other than that this looks good!

test/flux_calculator_tests.jl
juliasloan25: Is there a reason we can't add checks for the other fluxes?

juliasloan25 (approved):

src/Interfacer.jl
```diff
@@ -175,6 +175,11 @@ abstract type SeaIceModelSimulation <: SurfaceModelSimulation end
 abstract type LandModelSimulation <: SurfaceModelSimulation end
 abstract type OceanModelSimulation <: SurfaceModelSimulation end
 
+# Simulation objects tend to be very big, so it is best to make sure they are not printed in the REPL
+function Base.show(io::IO, sim::ComponentModelSimulation)
```
charleskawczynski, line 179:
```suggestion
function Base.show(io::IO, Base.@nospecialize(sim::ComponentModelSimulation))
```
since the implementation doesn't depend on anything in `sim`.

test/flux_calculator_tests.jl
Sbozzolo: Do we know that they should be zero?

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -320,7 +332,8 @@ function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:turbulent_fl
     surface_normal = @. CA.C3(CA.unit_basis_vector_data(CA.C3, surface_local_geometry))
 
     # get template objects for the contravariant components of the momentum fluxes (required by Atmos boundary conditions)
-    vec_ct12_ct1 = @. CA.CT12(CA.CT2(CA.unit_basis_vector_data(CA.CT1, surface_local_geometry)), surface_local_geometry)
+    # NOTE: This is allocating! Fix me if you can!
```
szy21, line 335: We can probably use the tensor_from_components function in ClimaAtmos, not sure it can fix the allocation. (You don't need to change it in this PR, just a comment.)

src/FluxCalculator.jl
```diff
@@ -304,19 +337,20 @@ function get_surface_fluxes!(inputs, surface_params::SF.Parameters.SurfaceFluxes
...
-    return (;
-        F_turb_ρτxz = F_turb_ρτxz,
-        F_turb_ρτyz = F_turb_ρτyz,
-        F_shf = F_shf,
-        F_lhf = F_lhf,
-        F_turb_moisture = F_turb_moisture,
-    )
+    L_MO = outputs.L_MO
+    # We need to be careful with L_MO because `Inf` is valid output
+    L_MO = ifelse.(isinf.(outputs.L_MO), zero(outputs.L_MO), outputs.L_MO)
```
szy21, line 348: Why do we need this?

Sbozzolo: We probably don't need this, since we are already filtering for NaNs in a different part of the code

Sbozzolo: I'll try removing it

docs/src/interfacer.md
Sbozzolo: I think they are required if you use the default FluxCalculator function, but you can probably provide methods to not use them.

src/FluxCalculator.jl
szy21: Is this setting Inf to zero for L_MO? It would be good if we can keep Inf in L_MO.

Sbozzolo: Yes, I'll do that

Sbozzolo:
I re-run AMIP and it passes: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/318

PR #1120, opened 2024-12-12
Tr/isolate run amip postprocessing

### LINKED ISSUE #1119: Isolate postprocessing/diagnostics

imreddyTeja:
## Purpose 

closes #1119 

## Content
This PR adds a struct for each different mode_name.
A struct is returned by get_cupler_args as mode_type (suggest a better name for the var).
This is used everywhere mode_name was. The structs do have hierarchy,
but the Abstract types should be given better names.

It also adds a new file,
experiments/ClimaEarth/user_io/postprocessing.jl, which contains
the `postprocess_sim` function. This function copies the if statements
previously at the end of run_amip.jl, but it uses dispatch for the simulation
mode. The postprocessing if statements are now moved into the functions. 

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,29 @@ pressure levels are being created.
 
 ### Code cleanup
 
+#### Structs representing simulation modes & postprocessing changes - PR [#1120](https://github.com/CliMA/ClimaCoupler.jl/pull/1120)
+
+The available simulation modes are now represented by the following structs:
+
+- `ClimaCoupler.Interfacer.AMIPMode`
+- `ClimaCoupler.Interfacer.SlabplanetMode`
+- `ClimaCoupler.Interfacer.SlabplanetAquaMode`
+- `ClimaCoupler.Interfacer.SlabplanetTerraMode`
+- `ClimaCoupler.Interfacer.SlabplanetEisenmanMode`
+
+All of the above structs are subtypes of the abstract
+`ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode`, and all of them except
+`ClimaCoupler.Interfacer.AMIPMode` are subtypes of `ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode`.
+
+These structs are used in `experiments/ClimaEarth/run_amip.jl` instead of representing the
+simulation mode as a string.
+
+The postprocessing in `experiments/ClimaEarth/run_amip.jl` is now moved into functions in
+the new `experiments/ClimaEarth/user_io/postprocessing.jl`. When the simulation is complete,
+`postprocess_sim` is called using the struct representing the simulation mode for dispatch.
+`postprocess_sim` has one method for all slabplanet simulation modes and another for the AMIP
+simulation mode. All postprocessing common to all simulation modes is done in the `common_postprocessing`
+function, which is called by both `postprocess_sim` methods.
 
 #### Output path updates - PRs [#1058][1], [#1106][2], [#1123][3]
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,9 +39,9 @@ pressure levels are being created.
 
 ### Code cleanup
 
-#### Structs representing simulation modes & postprocessing changes - PR [#1120](https://github.com/CliMA/ClimaCoupler.jl/pull/1120)
+#### Abstract types representing simulation modes & postprocessing changes - PR [#1120](https://github.com/CliMA/ClimaCoupler.jl/pull/1120)
 
-The available simulation modes are now represented by the following structs:
+The available simulation modes are now represented by the following abstract types:
 
 - `ClimaCoupler.Interfacer.AMIPMode`
 - `ClimaCoupler.Interfacer.SlabplanetMode`
@@ -49,16 +49,16 @@ The available simulation modes are now represented by the following structs:
 - `ClimaCoupler.Interfacer.SlabplanetTerraMode`
 - `ClimaCoupler.Interfacer.SlabplanetEisenmanMode`
 
-All of the above structs are subtypes of the abstract
+All of the above types are subtypes of the abstract
 `ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode`, and all of them except
 `ClimaCoupler.Interfacer.AMIPMode` are subtypes of `ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode`.
 
-These structs are used in `experiments/ClimaEarth/run_amip.jl` instead of representing the
+These types are used in `experiments/ClimaEarth/run_amip.jl` instead of representing the
 simulation mode as a string.
 
 The postprocessing in `experiments/ClimaEarth/run_amip.jl` is now moved into functions in
 the new `experiments/ClimaEarth/user_io/postprocessing.jl`. When the simulation is complete,
-`postprocess_sim` is called using the struct representing the simulation mode for dispatch.
+`postprocess_sim` is called using the type representing the simulation mode for dispatch.
 `postprocess_sim` has one method for all slabplanet simulation modes and another for the AMIP
 simulation mode. All postprocessing common to all simulation modes is done in the `common_postprocessing`
 function, which is called by both `postprocess_sim` methods.
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,11 +39,12 @@ pressure levels are being created.
 
 ### Code cleanup
 
-#### Output path updates - PRs [#1058](https://github.com/CliMA/ClimaCoupler.jl/pull/1058),
-    [#1106](https://github.com/CliMA/ClimaCoupler.jl/pull/1106),
-    [#1123](https://github.com/CliMA/ClimaCoupler.jl/pull/1123)
 
+#### Output path updates - PRs [#1058][1], [#1106][2], [#1123][3]
 
+[1]: https://github.com/CliMA/ClimaCoupler.jl/pull/1058
+[2]: https://github.com/CliMA/ClimaCoupler.jl/pull/1106
+[3]: https://github.com/CliMA/ClimaCoupler.jl/pull/1123
 Previously, ClimaEarth simulation outputs were saved in a path
 `experiments/ClimaEarth/output/$mode_name/$job_id/artifacts/`. Now, `ClimaEarth`
 creates output folders with an increment (increasing the counter every time the
@@ -69,7 +70,7 @@ coupler_output_dir_amip/
 ├── output_0002/
 │   └── ... component model outputs in their folders ...
 └── output_active -> output_0002/
-``
+```
 Note that any external scripts that assume an output path will need to be updated.
 
 #### Remove ClimaCoupler.Diagnostics module - PR [#953](https://github.com/CliMA/ClimaCoupler.jl/pull/953)
--- a/docs/src/interfacer.md
+++ b/docs/src/interfacer.md
@@ -217,5 +217,5 @@ end
 ```@docs
     ClimaCoupler.Interfacer.stub_init
     ClimaCoupler.Interfacer.AbstractSimulation
-    ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode
+    ClimaCoupler.Interfacer.AbstractSimulationMode
 ```
--- a/docs/src/interfacer.md
+++ b/docs/src/interfacer.md
@@ -199,11 +199,23 @@ end
     ClimaCoupler.Interfacer.AtmosModelSimulation
     ClimaCoupler.Interfacer.SurfaceModelSimulation
     ClimaCoupler.Interfacer.ComponentModelSimulation
-    ClimaCoupler.Interfacer.AbstractSimulation
     ClimaCoupler.Interfacer.SurfaceStub
-    ClimaCoupler.Interfacer.stub_init
     ClimaCoupler.Interfacer.float_type
     ClimaCoupler.Interfacer.name
     ClimaCoupler.Interfacer.get_field
     ClimaCoupler.Interfacer.update_field!
+    ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode
+    ClimaCoupler.Interfacer.AMIPMode
+    ClimaCoupler.Interfacer.SlabplanetMode
+    ClimaCoupler.Interfacer.SlabplanetAquaMode
+    ClimaCoupler.Interfacer.SlabplanetTerraMode
+    ClimaCoupler.Interfacer.SlabplanetEisenmanMode
+```
+
+## Interfacer Internal Functions and Types
+
+```@docs
+    ClimaCoupler.Interfacer.stub_init
+    ClimaCoupler.Interfacer.AbstractSimulation
+    ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode
 ```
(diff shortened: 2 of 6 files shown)


experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -1,5 +1,21 @@
 import YAML
 
+abstract type AbstractModeType end
+abstract type AbstractSlabPlanetModeType <: AbstractModeType end
+abstract type AbstractSlabPlanetModeSubType1 <: AbstractSlabPlanetModeType end
```
imreddyTeja, line 5: This needs a better name. It is the group of modes that is used for model setup [here](https://github.com/CliMA/ClimaCoupler.jl/blob/5318c3bc6723f0176036eef48758515b09ec9182/experiments/ClimaEarth/run_amip.jl#L337)

imreddyTeja: Or it could just be removed

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -1,5 +1,21 @@
 import YAML
 
+abstract type AbstractModeType end
```
juliasloan25, line 3: We could change `ModeType` -> `SimMode` here and throughout (e.g. `mode_type` would become `sim_mode` too). What do you think?

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+        compute_leaderboard(leaderboard_base_path, atmos_output_dir)
+        compute_pfull_leaderboard(leaderboard_base_path, atmos_output_dir)
+    end
+    common_postprocessing(cs, postprocessing_vars)
+end
+
+"""
+    common_postprocessing(cs, postprocessing_vars)
+
+Perfrom postprocessing common to all simulation types.
```
juliasloan25, line 103:
```suggestion
Perform postprocessing common to all simulation types.
```

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -1,5 +1,21 @@
...
+abstract type AbstractSlabPlanetModeType <: AbstractModeType end
+abstract type AbstractSlabPlanetModeSubType1 <: AbstractSlabPlanetModeType end
+struct AMIP_mode <: AbstractModeType end
```
juliasloan25, line 6: Julia convention is to use camel case for struct names, so this would be `AMIPMode` (same for the structs below, e.g. `SlabplanetMode`

experiments/ClimaEarth/user_io/arg_parsing.jl
juliasloan25: In the model setup, we could dispatch off of `AbstractSlabPlanetModeType`, and have a special case just for `SlabPlanetEisenman`

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+        dir_paths,
+        plot_diagnostics,
+        use_coupler_diagnostics,
+        output_default_diagnostics,
+        t_end,
+        conservation_softfail,
+        atmos_output_dir,
+        amip_diags_handler,
+        comms_ctx,
+    ) = postprocessing_vars
```
juliasloan25, line 22: `dir_paths`, `amip_diags_handler`, and `comms_ctx` are in `cs` and can be accessed from there in this function

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+postprocessing tasks that are common to all simulation types.
+"""
+function postprocess_sim(mode_type::AMIP_mode, cs, postprocessing_vars)
+    (;
+        dir_paths,
+        plot_diagnostics,
+        use_coupler_diagnostics,
+        output_default_diagnostics,
+        t_end,
+        conservation_softfail,
```
juliasloan25, line 57:
```suggestion
```

We don't need to extract args that won't be used in this function

juliasloan25: Same for the slabplanet method - e.g. we don't need the `amip_diags_handler` for that case

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+    ## plot all model states and coupler fields (useful for debugging)
+    !CA.is_distributed(comms_ctx) && debug(cs, dir_paths.artifacts)
+
+    # if isinteractive() #hide
+    #     ## clean up for interactive runs, retain all output otherwise #hide
+    #     rm(dir_paths.output; recursive = true, force = true) #hide
+    # end #hide
+
+    ## close all AMIP diagnostics file writers
+    !isnothing(amip_diags_handler) && map(diag -> close(diag.output_writer), amip_diags_handler.scheduled_diagnostics)
```
juliasloan25, line 122: It would be good for this to be isolated to the AMIP postprocessing method. It needs to be called last (which is maybe why it's here), but we can accomplish both of those things if we call `common_postprocessing` at the start of each mode-specific postprocess method, and move this to the end of the AMIP method.

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+"""
+    postprocess_sim(mode_type::AbstractSlabPlanetModeType, cs, postprocessing_vars)
+
+If they exist, perform conservation checks, for any slabplanet simulation, and then call
+`common_postprocessing` to perform common postprocessing tasks that are common to all simulation types.
+"""
+function postprocess_sim(mode_type::AbstractSlabPlanetModeType, cs, postprocessing_vars)
+    (;
+        dir_paths,
+        plot_diagnostics,
```
juliasloan25, line 14: The `postprocess_sim` methods don't need `plot_diagnostics`, only `common_postprocessing` does. It would be good to not retrieve it in methods it isn't needed, and we should check that all the other variables are needed too

experiments/ClimaEarth/user_io/arg_parsing.jl
juliasloan25: `SimulationMode`

juliasloan25: move these to Interfacer module

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,123 @@
...
+        make_diagnostics_plots(atmos_output_dir, dir_paths.artifacts)
+    end
+
+    ## plot all model states and coupler fields (useful for debugging)
+    !CA.is_distributed(comms_ctx) && debug(cs, dir_paths.artifacts)
+
+    # if isinteractive() #hide
+    #     ## clean up for interactive runs, retain all output otherwise #hide
+    #     rm(dir_paths.output; recursive = true, force = true) #hide
+    # end #hide
```
juliasloan25, line 119: we can just delete this

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,100 @@
+## helpers for user-specified IO
+include("debug_plots.jl")
+include("diagnostics_plots.jl")
+
+"""
+    postprocess_sim(sim_mode::AbstractSlabplanetSimulationMode, cs, postprocessing_vars)
+
+If they exist, perform conservation checks, for any slabplanet simulation, and then call
+`common_postprocessing` to perform common postprocessing tasks that are common to all simulation types.
```
juliasloan25, line 9: update docstring since `common_postprocessing` is called at the beginning of the function now

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -0,0 +1,100 @@
...
+
+    # Check this because we only want monthly data for making plots
+    if t_end > 84600 * 31 * 3 && output_default_diagnostics
+        include("leaderboard/leaderboard.jl")
+        leaderboard_base_path = cs.dirs.artifacts
+        compute_leaderboard(leaderboard_base_path, atmos_output_dir)
+        compute_pfull_leaderboard(leaderboard_base_path, atmos_output_dir)
+    end
+
+    !isnothing(cs.amip_diags_handler) &&
```
juliasloan25, line 81:
```suggestion

    # close all AMIP diagnostics file writers
    !isnothing(cs.amip_diags_handler) &&
```

src/Interfacer.jl
```diff
@@ -242,4 +248,24 @@ reinit!(sim::ComponentModelSimulation) = error("undefined reinit! for " * name(s
 # Include file containing the surface stub simulation type.
 include("surface_stub.jl")
 
+"""
+    AbstractModeType
```
juliasloan25, line 252:
```suggestion
    AbstractSimulationMode
```

src/Interfacer.jl
```diff
@@ -242,4 +248,24 @@ reinit!(sim::ComponentModelSimulation) = error("undefined reinit! for " * name(s
...
+
+An abstract type representing a simulation mode for slabplanet models.
+"""
+abstract type AbstractSlabplanetSimulationMode <: AbstractSimulationMode end
+
+struct AMIPMode <: AbstractSimulationMode end
+struct SlabplanetMode <: AbstractSlabplanetSimulationMode end
+struct SlabplanetAquaMode <: AbstractSlabplanetSimulationMode end
+struct SlabplanetTerraMode <: AbstractSlabplanetSimulationMode end
+struct SlabplanetEisenmanMode <: AbstractSlabplanetSimulationMode end
```
juliasloan25, line 269: It would be good to include short comments about each of these simulation setups. You can get this information from experiments/ClimaEarth/README.md

juliasloan25 (commented):
Thanks Teja! Just a note about documentation - we should update the NEWS.md explaining how the postprocessing has changed (e.g. now there are specific methods for each AMIP/slabplanet simulation type), and the new simulation type structs

experiments/ClimaEarth/run_amip.jl
```diff
@@ -327,18 +332,18 @@ if mode_name == "amip"
     CO2_field = Interfacer.update_field!(atmos_sim, Val(:co2), CO2_init)
 
     mode_specifics = (;
-        name = mode_name,
+        type = sim_mode,
```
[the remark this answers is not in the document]
imreddyTeja: I think the field is used inside of solver_coupler lines 691 and 752 
`cs.mode.type isa AMIPMode`

NEWS.md
```diff
@@ -29,11 +29,35 @@ pressure levels are being created.
...
+`postprocess_sim` is called using the type representing the simulation mode for dispatch.
+`postprocess_sim` has one method for all slabplanet simulation modes and another for the AMIP
+simulation mode. All postprocessing common to all simulation modes is done in the `common_postprocessing`
+function, which is called by both `postprocess_sim` methods.
+
+#### Output path updates - PRs [#1058][1], [#1106][2], [#1123][3]
+
+[1]: https://github.com/CliMA/ClimaCoupler.jl/pull/1058
+[2]: https://github.com/CliMA/ClimaCoupler.jl/pull/1106
+[3]: https://github.com/CliMA/ClimaCoupler.jl/pull/1123
```
juliasloan25, line 60: It's nice to keep these in the [...] (...) format because it creates a hyperlink (e.g. for PR 1058 [here](https://github.com/CliMA/ClimaCoupler.jl/blob/main/NEWS.md#output-path-updates---prs-1058)). However as it is in current main, the 2nd and 3rd PRs don't get linked because they're on newlines. Maybe we could do something like this:
```suggestion
#### Output path updates - PRs [#1058](https://github.com/CliMA/ClimaCoupler.jl/pull/1058), [#1106](https://github.com/CliMA/ClimaCoupler.jl/pull/1106), [#1123](https://github.com/CliMA/ClimaCoupler.jl/pull/1123)
```

docs/src/interfacer.md
```diff
@@ -199,11 +199,23 @@ end
...
+    ClimaCoupler.Interfacer.SlabplanetTerraMode
+    ClimaCoupler.Interfacer.SlabplanetEisenmanMode
+```
+
+## Interfacer Internal Functions and Types
+
+```@docs
+    ClimaCoupler.Interfacer.stub_init
+    ClimaCoupler.Interfacer.AbstractSimulation
+    ClimaCoupler.Interfacer.AbstractSlabplanetSimulationMode
```
juliasloan25, line 220:
```suggestion
```

`AbstractSlabplanetSimulationMode` is exported, so it shouldn't be under internal functions (it's already included in the API list above)

experiments/ClimaEarth/run_amip.jl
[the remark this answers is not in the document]
juliasloan25: Oh yeah, good point

src/Interfacer.jl
```diff
@@ -242,4 +248,65 @@ reinit!(sim::ComponentModelSimulation) = error("undefined reinit! for " * name(s
...
+ClimaLand.jl bucket land model, a prescribed ocean model, and a simple thermal sea ice model.
+"""
+abstract type AMIPMode <: AbstractSimulationMode end
+
+"""
+    SlabplanetMode
+
+An abstract type represeting the slabplanet simulation mode with a ClimaAtmos.jl atmosphere model,
+a ClimaLand.jl bucket land model, a thermal slab ocean model, and no sea ice model. Instead
+of using a sea ice model, the ocean evaluated in areas that would be covered in ice.
```
juliasloan25, line 281:
```suggestion
of using a sea ice model, the ocean is evaluated in areas that would be covered in ice.
```

juliasloan25 (approved):
This looks great, thank you!

NEWS.md
imreddyTeja: The notation I used also creates hyperlinks. See the preview [here](https://github.com/CliMA/ClimaCoupler.jl/blob/tr/isolate-run_amip-postprocessing/NEWS.md#output-path-updates---prs-1058-1106-1123)

I prefer this notation because it avoids making a very long line that wraps around when the source is viewed on most screens. If you prefer the normal hyperlink notation, I'll make the change.

docs/src/interfacer.md
imreddyTeja: Ah! Good catch. That was supposed to be `ClimaCoupler.Interfacer.AbstractSimulationMode`

imreddyTeja:
> This looks great, thank you!

Thank you! I wanted to double check that changing the simulation mode types from empty structs to abstract types is ok. I've seen both, so I'm not sure what the convention is. I changed them to abstract types to follow the pattern used by the simulation components

juliasloan25:
> > This looks great, thank you!
> 
> Thank you! I wanted to double check that changing the simulation mode types from empty structs to abstract types is ok. I've seen both, so I'm not sure what the convention is. I changed them to abstract types to follow the pattern used by the simulation components

I think either is fine in this case! I'm not sure what the convention is either, but from the [Julia manual](https://docs.julialang.org/en/v1/manual/types/#man-abstract-types):
> An important point to note is that there is no loss in performance if the programmer relies on a function whose arguments are abstract types, because it is recompiled for each tuple of concrete argument types with which it is invoked.

So for our usage of just dispatching off of these types, defining them as abstract types is okay (and empty structs would also be okay)

PR #1121, opened 2024-12-12
Simplify callbacks

Sbozzolo:
This commit removes most of the callback infrastructure in `ClimaCoupler` in favor of a simpler one that leverages what we already use in other packages. The current "schedule" system is the one from `ClimaDiagnostics`, but the plan is to move it to `ClimaUtilities` as part of the work towards the integer time.

Using the schedule system from `ClimaDiagnostics` eliminated the need for keeping track of the first day of the month.

I also updated the checkpointing callbacks so that they can run with frequencies that are smaller than one hour.


NEWS.md
```diff
@@ -6,27 +6,52 @@ ClimaCoupler.jl Release Notes
...
-### Leaderboard for variables over longitude, latitude, time, and pressure - PR [#1094](https://github.com/CliMA/ClimaCoupler.jl/pull/1094)
+#### Leaderboard for variables over longitude, latitude, time, and pressure - PR [#1094](https://github.com/CliMA/ClimaCoupler.jl/pull/1094)
 
 As a part of the post processing pipeline, bias plots for variables at the
 pressure levels of 850.0, 500.0, 250.0 hPa and bias plots over latitude and
 pressure levels are being created.
 
+### Breaking changes
+
+#### `hourly_checkpoint_dt` is now just `checkpoint_dt`
```
juliasloan25, line 32:
```suggestion
#### `hourly_checkpoint_dt` is now just `checkpoint_dt` - PR[#1121](https://github.com/CliMA/ClimaCoupler.jl/pull/1121)
```

config/ci_configs/amip_coarse_ft64_hourly_checkpoints_restart.yml
```diff
@@ -3,8 +3,7 @@ dt_save_restart: "10days"
 dt_save_to_sol: "1days"
 energy_check: false
 h_elem: 6
-hourly_checkpoint: true
-hourly_checkpoint_dt: 1
+checkpoint_dt: 1
```
juliasloan25, line 6: does this need to be:
```suggestion
checkpoint_dt: "1hours"
```
?

config/amip_configs/amip.yml
```diff
@@ -9,8 +9,7 @@ dt_save_to_sol: "30days"
 dz_bottom: 30.0
 energy_check: false
 h_elem: 16
-hourly_checkpoint: true
-hourly_checkpoint_dt: 720
+checkpoint_dt: "720hours"
```
juliasloan25, line 12: I find it's easier to parse the YML files when we keep them alphabetized - Could you move the `checkpoint_dt` args up in the YML files that have it now?

experiments/ClimaEarth/cli_options.jl
```diff
@@ -73,15 +73,10 @@ function argparse_settings()
...
-        "--hourly_checkpoint"
-        help = "Boolean flag indicating whether to checkpoint at multiple-hourly intervals [false (default), true]"
-        arg_type = Bool
-        default = false
-        "--hourly_checkpoint_dt"
-        help = "Time interval for hourly checkpointing in hours [480 (default)]"
-        arg_type = Int
-        default = 480
+        "--checkpoint_dt"
+        help = "Time interval for hourly checkpointing [480hours (default)]"
```
juliasloan25, line 77:
```suggestion
        help = "Time interval for hourly checkpointing ["480hours" (default)]"
```

src/TimeManager.jl
```diff
@@ -59,123 +50,82 @@ datetime_to_strdate(datetime::Dates.DateTime) =
...
-"""
-struct EveryTimestep <: AbstractFrequency end
-
-"""
-    trigger_callback(cs, ::Monthly)
+The string can be in one of two formats:
 
-Returns `true` if the current date is equal to or exceeds the saved first of the month at time of 00:00:00.
+1. **Months:** `<NUM>months`, where `<NUM>` is an integer representing the number of months. For example, `"2months"` represents 2 months.
+2. **Other units:** `<NUM><unit>`, where `<NUM>` is a number (integer or floating point) and `<unit>` is one of `secs`, `mins`, `hours`, or `days`.
```
juliasloan25, line 60: Isn't months just a special case of the generalized "other units"? Why does it have to be handled differently?

src/TimeManager.jl
```diff
@@ -59,123 +50,82 @@ datetime_to_strdate(datetime::Dates.DateTime) =
...
+        return Dates.Millisecond(1000 * time_to_seconds(s))
+    end
 end
 
 """
-    MonthlyCallback{FT}
+    Callback
 
-This is a callback type that triggers at intervals of 1 month or multiple months.
+A small struct containing a schedule, and the function to be executed on the
```
juliasloan25, line 95:
```suggestion
A small struct containing a schedule, and the function to be executed if the
```

src/TimeManager.jl
```diff
@@ -59,123 +50,82 @@ datetime_to_strdate(datetime::Dates.DateTime) =
...
+schedule is true.
 
-# Fields
+A `schedule` is a callable object (ie, a function) that takes an integrator-type
+of object and returns true or false.
+TODO: If `cs` contained the correct time, we could just pass `cs` to the schedule
```
juliasloan25, line 100: I recently removed time from `CoupledSimulation` because it wasn't being used or correctly updated, but if it would be useful we can add it back.

juliasloan25 (approved):
Looks good, thank you! Have you verified that this works as expected? E.g. that the checkpoints in run_amip.jl are done as expected?

src/TimeManager.jl
Sbozzolo: All the other units have a fixed duration, but the number of seconds in month depends on the start date. But this doesn't really matter from a user point of view, so I will remove this distinction

config/ci_configs/amip_coarse_ft64_hourly_checkpoints_restart.yml
Sbozzolo: Yes, I missed it!

src/TimeManager.jl
Sbozzolo: It could be, but we can do this later

Sbozzolo:
> Looks good, thank you! Have you verified that this works as expected? E.g. that the checkpoints in run_amip.jl are done as expected?

I verified in [this amip](build) that the output is the same. The only difference is that the previous system was also checkpointing at t = 0.

PR #1129, opened 2024-12-16
add bucket initial condition

szy21:
## Purpose 

Adds an option to read bucket initial condition from a netcdf file

## Content

----


experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -138,6 +139,48 @@ function bucket_init(
     Y.bucket.Ws .= 0.0
     Y.bucket.σS .= 0.0
 
+    # Overwrite initial conditions with interpolated values from a netcdf file using
+    # the `SpaceVaryingInputs` tool. We expect the file to contain the following variables:
```
juliasloan25, line 143: It would be good to add an assertion checking that the file does contain all these variables, and provide an informative message if not

juliasloan25 (approved):
Thank you!

experiments/ClimaEarth/run_cloudy_slabplanet.jl
```diff
@@ -206,6 +206,7 @@ land_sim = bucket_init(
     tspan,
     "sphere",
     "map_static",
+    "",
```
Sbozzolo, line 209: Could you name a variable a pass it here, instead of just passing an empty string (so that it is clear what's this argument for)

Sbozzolo (approved):

PR #1130, opened 2024-12-18
Update climacommon modules

### LINKED ISSUE #1112: Fix circular dependencies in precompilation
On buildkite, our precompilation step skips many dependencies (50+ per environment) because of circular dependencies. This happens for the `exp/ClimaCore/`, `exp/ClimaEarth`, and `test` environments, but not the top-level environment. We should investigate why this happens, since skipping precompilation will negatively impact performance.

See the init step of [this build](https://buildkite.com/clima/climacoupler-ci/builds/5281#019397a6-95a2-4bb9-889e-1853d710825b) for an example.

Sbozzolo:
This should remove the circular dependency error

Closes #1112

### DIFF
--- a/.buildkite/amip/pipeline.yml
+++ b/.buildkite/amip/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   queue: clima
   slurm_time: 96:00:00
-  modules: climacommon/2024_10_09
+  modules: climacommon/2024_12_16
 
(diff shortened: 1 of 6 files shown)


juliasloan25 (approved):

juliasloan25:
cc @charleskawczynski

Issue #1122, opened 2024-12-12
Improve user experience


juliasloan25:
We want to streamline a few use cases for running the `run_amip.jl` driver:

- running from the REPL (config file choice)
- changing a few config args from one run to the next in REPL, without re-running everything

PR #1133, opened 2024-12-21
Interpret config file/args as relative to pkg_dir

imreddyTeja:
## Purpose 

This addressed the "make sure output dir doesn't end up being experiments/ClimaEarth/experiments/ClimaEarth/... when run from exp/ClimaEarth/ directory" point from issue #1122 

## To-do
- improve the cli help strings to make it clear the way paths are interpreted

## Content
Interpret the config file path as relative to the
ClimaCoupler.jl package. Do the same for the restart_dir and coupler_output_dir in the config.

This enables `run_amip` to also be run from
`experiments/ClimaEarth` with the default config path and without modifying the usual output path.

----

### DIFF
--- a/experiments/ClimaEarth/run_amip.jl
+++ b/experiments/ClimaEarth/run_amip.jl
@@ -139,7 +139,7 @@ postprocessing and the conservation checks) of the simulation will be saved,
 #and `dir_paths.checkpoints`, where restart files are saved.
 =#
 
-COUPLER_OUTPUT_DIR = joinpath(output_dir_root, job_id)
+COUPLER_OUTPUT_DIR = joinpath(pkg_dir, output_dir_root, job_id)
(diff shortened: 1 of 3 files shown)


experiments/ClimaEarth/run_amip.jl
```diff
@@ -139,7 +139,7 @@ postprocessing and the conservation checks) of the simulation will be saved,
 #and `dir_paths.checkpoints`, where restart files are saved.
 =#
 
-COUPLER_OUTPUT_DIR = joinpath(output_dir_root, job_id)
+COUPLER_OUTPUT_DIR = joinpath(pkg_dir, output_dir_root, job_id)
```
Sbozzolo, line 142: I don't think it should not be relative to `pkg_dir`. `pkg_dir` could be inside the `.julia` in the user's home directory. This is a hidden folder, with possibly storage quotas. It is not a great place where to save files.

PR #1134, opened 2024-12-23
Move include outside of `postprocess_sim` function

imreddyTeja:
## Purpose 

#1120 moved post-processing for `run_amip.jl` into a function. Inside the function, there is an include for the leaderboard code. This moves the include outside of the function. 

## Content

----

### DIFF
--- a/experiments/ClimaEarth/user_io/postprocessing.jl
+++ b/experiments/ClimaEarth/user_io/postprocessing.jl
@@ -1,6 +1,7 @@
 ## helpers for user-specified IO
 include("debug_plots.jl")
 include("diagnostics_plots.jl")
+include("../leaderboard/leaderboard.jl")
 
 """
     postprocess_sim(::Type{AbstractSlabplanetSimulationMode}, cs, postprocessing_vars)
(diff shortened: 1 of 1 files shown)


experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -72,7 +73,6 @@ function postprocess_sim(::Type{AMIPMode}, cs, postprocessing_vars)
 
     # Check this because we only want monthly data for making plots
     if t_end > 84600 * 31 * 3 && output_default_diagnostics
-        include("leaderboard/leaderboard.jl")
```
charleskawczynski, line 75: Can we also move `include("user_io/diagnostics_plots.jl")` into global scope? Julia does not permit loading code inside functions like that.

charleskawczynski: And below (`include("user_io/diagnostics_plots.jl")`)

imreddyTeja: Ah sorry, I didn't see all the includes inside the functions. Fixed now.

charleskawczynski: It looks like the includes were not added, they were only removed

szy21: It seems `diagnostics_plots.jl` is already included at the top of this file.

szy21:
I'll get this in to see if it fixes the nightly amip.

szy21 (approved):

PR #1135, opened 2025-01-14
use proper constructor naming convention

### LINKED ISSUE #1126: make constructors use object name
For example, `bucket_init` is really a constructor that returns a `BucketSimulation` object (similar for `atmos_init`, `ocean_init`, `ice_init`, etc). We should rename the function to reflect that

juliasloan25:
## Purpose 
Our constructors have been using names like `<sim>_init`. This PR renames them to use e.g. `SimName` instead, following constructor naming convention.

closes #1126

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,12 @@ callback = Callback(schedule, func)
 maybe_trigger_callback(callback, cs, t)
 ```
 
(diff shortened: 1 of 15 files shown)


Sbozzolo (approved):
Thank you!

PR #1141, opened 2025-01-17
move SIC TimeVaryingInput into PrescribedIceSim

### LINKED ISSUE #1136: Update surface models with prescribed data from file in `step!`
Currently, we read in `SST` and `SIC` from files in the AMIP setup. These are created as `TimeVaryingInput`s which get evaluated at each coupling timestep in the AMIP case only. This approach requires us to access the simulation mode within the coupling loop, and adds unnecessary complexity to the coupling loop.

Instead, we should have prescribed surface simulations that read in data directly from files. Specifically, we'll need to cover two cases:
- `PrescribedOceanSimulation` which reads in SST from a provided file and reads a new value at each coupling timestep
  - Instead of starting from scratch, we can extend `SurfaceStub` and use the get/update field functionality already implemented for that.
- `PrescribedIceSimulation` which reads in SIC and performs some simple calculation based on it.
  - This almost has what we want already, but we should be updating the sea ice concentration during the step, rather than externally via a call to `evaluate!` from the driver

Part of https://github.com/CliMA/ClimaCoupler.jl/issues/1011

```[tasklist]
### To Do
- [x] Move SIC `evaluate!` calls from driver into `PrescribedIceSimulation` initialization and step #1141
- [ ] Create new `PrescribedOceanSimulation` extending `SurfaceStub`, and move SST `evaluate!` calls from the driver into `PrescribedOceanSimulation` initialization and step #1144
```

juliasloan25:
## Purpose 
Move SIC TimeVaryingInput object and related updates out of the driver and into the `PrescribedIceSimulation` object. Also does some misc. cleanup of `PrescribedIceSimulation` that I found along the way.

first part of #1136 , to be followed by moving SST TimeVaryingInput out of driver

## To do
- [x] move SIC updates out of driver, into PrescribedIceSimulation
- [x] update tests to use new setup
- [x] check AMIP CI results


Sbozzolo (approved):

juliasloan25:
Comparing the [most recent build on main](https://buildkite.com/clima/climacoupler-ci/builds/5412#_) to the [build from this PR](https://buildkite.com/clima/climacoupler-ci/builds/5473), there are slight changes in the output AMIP plots. I believe this is due to the change in order of operations introduced by this PR. Previously, the coupling loop would have this order: read in the current SIC, update surface fractions of surface models, send information from coupler fields to component models, step each model, and continue with the rest of the coupling loop steps. Now, we first update surface fractions of surface models, send information from coupler to components, and then step the models (the SIC read and another surface fraction update for sea ice are now included in the step).

#### AMIP Float32: sea ice surface temperature (before this PR)
Screenshot 2025-01-23 at 9 55 47 AM

#### AMIP Float32: sea ice surface temperature (after this PR)
Screenshot 2025-01-23 at 9 55 24 AM

#### AMIP Float32: atmos 2D fields (before this PR)
Screenshot 2025-01-23 at 9 59 05 AM

#### AMIP Float32: atmos 2D fields (after this PR)
Screenshot 2025-01-23 at 9 59 16 AM


In runs that only run for 2 timesteps (many of our CI runs), the sea ice output looks very different. Again, I expect this is due to order of operations changing. Note that the other sea ice fields (area fraction and surface humidity) look identical between these two runs.

#### Target AMIP albedo from function: sea ice surface temperature (before this PR)
Screenshot 2025-01-23 at 10 04 21 AM

#### Target AMIP albedo from function: sea ice surface temperature (after this PR)
Screenshot 2025-01-23 at 10 03 35 AM

I think these changes are acceptable, but we should be aware that this PR does change behavior slightly.

PR #1143, opened 2025-01-17
Remove CO2 from AMIP driver

### LINKED ISSUE #1137: Remove CO2 update from driver
Currently, we read in `CO2` from a file in the AMIP setup. This is created as a `TimeVaryingInput` which gets evaluated at each coupling timestep in the AMIP case only. This approach requires us to access the simulation mode within the coupling loop, and adds unnecessary complexity to the coupling loop.

Instead, we should move the prescribed `CO2` into ClimaAtmos, alongside similar prescribed quantities like ozone and aerosols.

Part of https://github.com/CliMA/ClimaCoupler.jl/issues/1011

```[tasklist]
#### To Do
- [ ] Add PrescribedCO2 to ClimaAtmos cache
- [ ] Move CO2 artifact out of ClimaCoupler artifacts, into ClimaAtmos
- [ ] Remove `update_field!` methods for atmos simulations in ClimaCoupler
- [ ] Remove CO2 updates from ClimaCoupler driver
```

Sbozzolo:
https://github.com/CliMA/ClimaAtmos.jl/pull/3522 implements reading CO2 directly in Atmos

Closes #1137

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,14 @@ object initialization and `step!`. Note that this results in changed order of
 operations and slightly different behavior for AMIP simulations - see the PR
 description for more details.
 
+#### CO2 is now updated by ClimaAtmos directly PR[#1143](https://github.com/CliMA/ClimaCoupler.jl/pull/1143)
+
+Previously, CO2 was read in directly from the driver during init and in the
+coupling loop. Now, it is read internally by ClimaAtmos. Note that this results
+in changed order of operations and slightly different behavior for AMIP
+simulations - see the PR description for more details.
+
+
 #### Simplified callbacks PR [#1121](https://github.com/CliMA/ClimaCoupler.jl/pull/1121)
 
 Callbacks were also reworked, and the previous system was removed. Here is an example of the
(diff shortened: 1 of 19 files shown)


config/benchmark_configs/amip_diagedmf.yml
```diff
@@ -2,6 +2,7 @@ FLOAT_TYPE: "Float32"
 aerosol_radiation: true
 atmos_config_file: "config/benchmark_configs/climaatmos_diagedmf.yml"
```
juliasloan25, line 3: This config (and `amip_diagedmf_io.yml`) take the atmosphere setup from `climaatmos_diagedmf.yml` and `climaatmos_diagedmf_io.yml` in this same directory, respectively. CO2 doesn't need to be specified here because it's already specified in the atmos-specific configs

juliasloan25 (approved):

config/benchmark_configs/amip_diagedmf.yml
Sbozzolo: Thank you, I removed it from the two files.

PR #1144, opened 2025-01-18
implement PrescribedOceanSim to read in SST

juliasloan25:
## Purpose 
Implement a simple `PrescribedOceanSimulation` model that reads in SST from prescribed data at each timestep. This will allow us to remove the SST reading from the driver.

Second and final part of #1136

## Content
- [x] add new `PrescribedOceanSimulation` type that reads in prescribed data during init and step
  - this simulation has very similar behavior to the existing `SurfaceStub` type, with the addition of reading prescribed data
- [x] add `AbstractSurfaceStub` type that `SurfaceStub` and `PrescribedOceanSimulation` both extend to share methods between these two very similar models
- [x] add tests
- [x] rename `land_area_fraction` to `land_fraction` for naming consistency with area fractions of other surface models
- [x] check AMIP CI results


experiments/ClimaEarth/run_amip.jl
```diff
@@ -308,9 +273,14 @@ if sim_mode <: AMIPMode
...
         date0,
         mono_surface,
-        land_area_fraction,
+        land_fraction,
     )
 
+    ## ocean model using prescribed data
+    ice_fraction = Interfacer.get_field(ice_sim, Val(:area_fraction))
+    ocean_fraction = FT(1) .- ice_fraction .- land_fraction
+    ocean_sim = PrescribedOceanSimulation(FT, boundary_space, date0, t_start, ocean_fraction, thermo_params, comms_ctx)
```
juliasloan25, line 282: The ocean model init is moved after the sea ice model init so we are consistent in our priority of area fractions (land fraction is applied first, then sea ice, then ocean). This now matches the ordering in the FieldExchanger module

experiments/ClimaEarth/components/ocean/prescr_seaice.jl
```diff
@@ -229,7 +231,8 @@ function ice_rhs!(dY, Y, p, t)
 
     # Overwrite ice fraction with the static land area fraction anywhere we have nonzero land area
     #  max needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
-    @. p.area_fraction = max(min(p.area_fraction, FT(1) - p.land_area_fraction), FT(0))
+    @. p.area_fraction = apply_binary_mask(p.area_fraction, p.mono_surface)
```
juliasloan25, line 234: I forgot this line in the previous sea ice PR

juliasloan25:
This PR causes minor behavioral changes in AMIP runs. E.g. see the following plots before and after these changes

#### Atmos 2D fields - before this PR
Screenshot 2025-01-24 at 11 15 07 AM

#### Atmos 2D fields - after this PR
Screenshot 2025-01-24 at 11 15 20 AM

#### Coupler fluxes - before this PR
Screenshot 2025-01-24 at 11 22 17 AM

#### Coupler fluxes - after this PR
Screenshot 2025-01-24 at 11 22 48 AM

experiments/ClimaEarth/components/ocean/prescr_ocean.jl
```diff
@@ -0,0 +1,116 @@
+import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput, evaluate!
+import ClimaUtilities.ClimaArtifacts: @clima_artifact
+import Interpolations # triggers InterpolationsExt in ClimaUtilities
+import Thermodynamics as TD
+import ClimaCoupler: Checkpointer, FieldExchanger, Interfacer
+
+"""
+    PrescribedOceanSimulation{P, Y, D, I}
```
Sbozzolo, line 8:
```
PrescribedOceanSimulation{I}
```
What does `I` stand for?

src/surface_stub.jl
```diff
@@ -1,87 +1,106 @@
...
+contains the necessary fields for the simulation, at minimum the following:
+    - `T_sfc` (surface temperature [K])
+    - `ρ_sfc` (surface air density [kg / m3])
+    - `z0m` (roughness length for momentum [m])
+    - `z0b` (roughness length for tracers [m])
+    - `beta` (evaporation scaling factor)
+    - `α_direct` (direct albedo)
+    - `α_diffuse` (diffuse albedo)
+    - `area_fraction` (fraction of the grid cell covered by the ocean)
+    - `phase` (phase of the water used to calculate surface humidity)
```
Sbozzolo, line 15: This doesn't look accurate. What would this be for land?

experiments/ClimaEarth/components/ocean/prescr_ocean.jl
```diff
@@ -0,0 +1,116 @@
...
+        date0,
+        area_fraction,
+        thermo_params,
+        comms_ctx;
+        z0m = FT(5.8e-5),
+        z0b = FT(5.8e-5),
+        beta = FT(1),
+        α_direct_val = FT(0.06),
+        α_diffuse_val = FT(0.06),
+        phase = TD.Liquid(),
```
Sbozzolo, line 46: Do you need phase to be an option?

Wouldn't this always be Liquid?

test/component_model_tests/prescr_ocean_tests.jl
```diff
@@ -0,0 +1,74 @@
...
+
+@testset "PrescribedOceanSimulation name" begin
+    sim = PrescribedOceanSimulation((;))
+    @test Interfacer.name(sim) == "PrescribedOceanSimulation"
+end
+
+@testset "PrescribedOceanSimulation constructor" begin
+    space = TestHelper.create_space(FT)
+    date0 = Dates.DateTime(2000, 1, 1)
+    t_start = Float64(0)
```
Sbozzolo, line 22:
```suggestion
t_start = 0.0
```

experiments/ClimaEarth/components/ocean/prescr_ocean.jl
juliasloan25: I just copied this over from the SurfaceStub definition. I can change it to `C` if you prefer

src/surface_stub.jl
juliasloan25: It's used when we calculate surface humidity [here](https://github.com/CliMA/ClimaCoupler.jl/blob/e049e032dc9bbd122a0a0e7b442db1d8b2d76d90/src/surface_stub.jl#L25-L26). Land has a different [method](https://github.com/CliMA/ClimaCoupler.jl/blob/e049e032dc9bbd122a0a0e7b442db1d8b2d76d90/experiments/ClimaEarth/components/land/climaland_bucket.jl#L238-L239) to compute surface humidity so it doesn't use the phase

juliasloan25: SurfaceStub has this option so we can distinguish between ocean and ice surfaces using SurfaceStub

Sbozzolo (approved):

szy21 (approved):
Thanks! Are `α_direct` and `α_diffuse` still updated correctly with this change?

juliasloan25:
> Thanks! Are `α_direct` and `α_diffuse` still updated correctly with this change?

Yes, with the new `AbstractSurfaceStub` type, fields of `PrescribedOceanSim` (including `α_direct` and `α_diffuse`) will be updated and accessed exactly how they were previously when we were using a `SurfaceStub` type for the AMIP ocean model, via the `update_field!` and `get_field` functions in src/surface_stub.jl

PR #1148, opened 2025-01-23
use t_start correctly in driver

### LINKED ISSUE #1147: t_start not correctly used in driver
There are a couple of places where we implicitly assume `t_start=0` in the driver, by using `dt_cpl` where we should be using `t_start + dt_cpl`.

- initial `step!` calls [here](https://github.com/CliMA/ClimaCoupler.jl/blob/c3d396b19e2a7ffa974e0939c7857220e340b453/experiments/ClimaEarth/run_amip.jl#L625-L627)
- initial 2 steps for precompilation [here](https://github.com/CliMA/ClimaCoupler.jl/blob/c3d396b19e2a7ffa974e0939c7857220e340b453/experiments/ClimaEarth/run_amip.jl#L751-L757)

juliasloan25:
## Purpose 
Use `t_start + Δt_cpl` instead of `Δt_cpl` to avoid implicit assumption that `t_start = 0`

closes #1147

### DIFF
--- a/experiments/ClimaEarth/run_amip.jl
+++ b/experiments/ClimaEarth/run_amip.jl
@@ -622,9 +622,9 @@ FieldExchanger.update_model_sims!(cs.model_sims, cs.fields, cs.turbulent_fluxes)
 
 # 3.surface vapor specific humidity (`q_sfc`): step surface models with the new surface density to calculate their respective `q_sfc` internally
 ## TODO: the q_sfc calculation follows the design of the bucket q_sfc, but it would be neater to abstract this from step! (#331)
-Interfacer.step!(land_sim, Δt_cpl)
-Interfacer.step!(ocean_sim, Δt_cpl)
-Interfacer.step!(ice_sim, Δt_cpl)
+Interfacer.step!(land_sim, tspan[1] + Δt_cpl)
+Interfacer.step!(ocean_sim, tspan[1] + Δt_cpl)
+Interfacer.step!(ice_sim, tspan[1] + Δt_cpl)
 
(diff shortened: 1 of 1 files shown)


experiments/ClimaEarth/run_amip.jl
```diff
@@ -749,11 +749,11 @@ beginning and end of the simulation timespan to the correct values.
...
 
 ## run the coupled simulation for two timesteps to precompile
-cs.tspan[2] = Δt_cpl * 2
+cs.tspan[2] = t_start + Δt_cpl * 2
 solve_coupler!(cs)
 
 ## update the timespan to the correct values
-cs.tspan[1] = Δt_cpl * 2
+cs.tspan[1] = t_start + Δt_cpl * 2
 cs.tspan[2] = tspan[2]
```
imreddyTeja, line 757: I'm not completely sure I understand the difference between `tspan[1]` and `t_start`, and between `tspan[2]` and `t_end`.  If it is always the case that `tspan = = [t_start, t_end]`, then I think it might make more sense to only use `t_start` and `t_end`, or to only use `tspan` when modifying `cs.tspan`, instead of a combination of `t_start` and `t_span[2]`

imreddyTeja (approved):

experiments/ClimaEarth/run_amip.jl
juliasloan25: Good point! We always have `tspan == [t_start, t_end]`. It looks like throughout the driver we always use `tspan` after initialization, so I'll change this to use `tspan` only

PR #1151, opened 2025-01-24
allow list of TOML files

### LINKED ISSUE #1149: allow list of TOML files rather than just one
For calibration, we want to be able to use mulitiple TOML parameter files and merge them using `ClimaParams.merge_toml_files`. Right now we only take in a single TOML, so we should update this input.

juliasloan25:
## Purpose 
For calibration purposes, we want to be able to take in multiple TOML files. This PR updates ClimaCoupler to use a more similar structure to ClimaAtmos in allowing multiple TOMLs. Also renames "coupler_toml_file" to "coupler_toml" for brevity.

closes #1149

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -342,16 +342,14 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
         error("Invalid `atmos_config_repo`; please use \"ClimaCoupler\" or \"ClimaAtmos\"")
     end
 
-    # use coupler toml if atmos is not defined
-    atmos_toml_file = atmos_config["toml"]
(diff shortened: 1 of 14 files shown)


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -342,14 +343,12 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
         error("Invalid `atmos_config_repo`; please use \"ClimaCoupler\" or \"ClimaAtmos\"")
     end
 
-    # use coupler toml if atmos is not defined
-    atmos_toml_file = atmos_config["toml"]
-    coupler_toml_file = coupler_dict["coupler_toml_file"]
+    # use atmos toml if coupler toml is not defined
```
juliasloan25, line 346: This is the logic we had in the code before, but the comment didn't match what the code was doing

nefrathenrici (approved):
Thank you!

PR #1152, opened 2025-01-27
Don't force logger for ClimaComms >= 0.6.6

Sbozzolo:
`ClimaComms` 0.6.6 introduces new features related to logging. Now, silencing non-root loggers is default and requires no action on the user's part.

The version also introduces new loggers that can be used to save the output to file.

https://clima.github.io/ClimaComms.jl/dev/logging/

(To be merged after we bump dependencies)


juliasloan25 (approved):
#1155 has been merged, so this can be merged now

PR #1154, opened 2025-01-29
Remove radiative_energy_toa

Sbozzolo:
`radiative_energy_toa` is a field fully owned by atmos. There is no reason for ClimaCoupler to maintain a copy. Instead, we can directly access the value from atmos if we need it (it's only needed for conservation checks).

This PR also fixes the `ConservationChecker` test. The test was not accounting for the movement of water and energy across the component, and was "working" only because there was a bug in `runtime_check` that made one of the asserts always trivially satisfied.

### DIFF
--- a/experiments/ClimaEarth/test/debug_plots_tests.jl
+++ b/experiments/ClimaEarth/test/debug_plots_tests.jl
@@ -3,7 +3,7 @@ import Test: @test, @testset, @test_logs
 import ClimaCore as CC
 import ClimaCoupler: Interfacer
 import ClimaComms
-@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
+ClimaComms.@import_required_backends
 
 # Prevent GKS headless operation mode warning
 ENV["GKSwstype"] = "nul"
--- a/experiments/ClimaEarth/test/debug_plots_tests.jl
+++ b/experiments/ClimaEarth/test/debug_plots_tests.jl
@@ -48,7 +48,6 @@ plot_field_names(sim::Interfacer.SurfaceStub) = (:stub_field,)
         :beta,
         :z0b_sfc,
         :z0m_sfc,
-        :radiative_energy_flux_toa,
     ]
     atmos_names = (:atmos_field,)
     surface_names = (:surface_field,)
--- a/src/Utilities.jl
+++ b/src/Utilities.jl
@@ -11,7 +11,7 @@ import ClimaCore as CC
 import Logging
 import ClimaUtilities.OutputPathGenerator: generate_output_path
 
-export swap_space!, get_device, get_comms_context, show_memory_usage, setup_output_dirs, time_to_seconds
+export swap_space!, get_device, get_comms_context, show_memory_usage, setup_output_dirs, time_to_seconds, integral
 
 """
     swap_space!(space_out::CC.Spaces.AbstractSpace, field_in::CC.Fields.Field)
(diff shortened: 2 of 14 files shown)


experiments/ClimaEarth/test/debug_plots_tests.jl
```diff
@@ -35,7 +35,7 @@ plot_field_names(sim::Interfacer.SurfaceStub) = (:stub_field,)
 @testset "import_atmos_fields!" begin
 
     boundary_space = TestHelper.create_space(FT)
-    coupler_names = [
+    coupler_names = (
```
juliasloan25, line 38: Just curious, why did you change this from a vector to a tuple?

juliasloan25 (approved):
thank you!

experiments/ClimaEarth/test/debug_plots_tests.jl
Sbozzolo: Oops, it was a mistake!

src/Utilities.jl
```diff
@@ -188,4 +188,26 @@ function time_to_seconds(s::String)
 end
 
 
+"""
+    integral(field)
+
+Return the integral for `field` along its spatial dimensions.
```
juliasloan25, line 194: Is the input always expected to be a 2D field? What will happen if a 3D field is passed in? Maybe also note explicitly that the return value will be a scalar

test/conservation_checker_tests.jl
```diff
@@ -41,7 +42,10 @@ for FT in (Float32, Float64)
         space = CC.CommonSpaces.CubedSphereSpace(FT; radius = FT(6371e3), n_quad_points = 4, h_elem = 4)
 
         # set up model simulations
-        atmos = TestAtmos((; space = space))
+        initial_energy = ones(space) .* CC.Spaces.undertype(space)(1e6)
+        initial_water = ones(space) .* CC.Spaces.undertype(space)(1e5)
```
juliasloan25, line 46:
```suggestion
        initial_energy = ones(space) .* 1e6
        initial_water = ones(space) .* 1e5
```
I think the float type will be inherited from the space

test/conservation_checker_tests.jl
```diff
@@ -82,17 +84,29 @@ for FT in (Float32, Float64)
         )
 
         # set non-zero radiation and precipitation
-        F_r = cf.radiative_energy_flux_toa
+        F_r = 200 .* CC.Fields.ones(space)
         P = cf.P_liq
         Δt = float(cs.Δt_cpl)
 
+        volume = integral(ones(space))
+        # 0.5 is the area fraction
```
juliasloan25, line 92: which area fraction? If it's ice (which has area fraction 0.5), let's use `get_field(ice, Val(:area_fraction))` instead

test/conservation_checker_tests.jl
```diff
@@ -82,17 +84,29 @@ for FT in (Float32, Float64)
...
+        water_from_precipitation = integral(P) .* FT(Δt) .* FT(0.5)
+        energy_from_radiation = integral(F_r) .* FT(Δt)
+        energy_per_unit_cell = CC.Fields.ones(space) .* energy_from_radiation ./ volume
+        water_per_unit_cell = CC.Fields.ones(space) .* water_from_precipitation ./ volume
+
         # analytical solution
-        tot_energy_an = sum(FT.(F_r .* 3Δt .+ 1e6 .* 1.25))
-        tot_water_an = sum(FT.(.-P .* 3Δt .* 0.5 .+ CC.Fields.ones(space)))
+        # 1.25 comes from the area fraction of the ocean
```
juliasloan25, line 99: why do we include the ocean area fraction but not ice or land?

src/Utilities.jl
```diff
@@ -188,4 +188,30 @@ function time_to_seconds(s::String)
...
+"""
+function integral(field::CC.Fields.Field)
```
juliasloan25, line 196: We could add a test for this function that `sum(2d field) == integral(level field)` since that's the problem we're trying to solve with it

test/conservation_checker_tests.jl
Sbozzolo: 1e6 is Float64 and `ones()` can be Float32, which leads to a Float64 Field, and this cannot be used in other operations below because of the mixed type

src/Utilities.jl
Sbozzolo: I added a check to ensure that it is only 2d

test/conservation_checker_tests.jl
Sbozzolo: Because there's water only on the ocean as the `get_field` are defined here (and there's no ice in this test).

test/conservation_checker_tests.jl
Sbozzolo: Added the get_field

PR #1156, opened 2025-01-29
use main branch of CTS for nightly AMIP

juliasloan25:
## Purpose 
This should help us catch breaking changes coming from ClimaTimeSteppers sooner

note: this will break until packages downstream of ClimaTimeSteppers are updated to use v0.8.0 (~ClimaAtmos https://github.com/CliMA/ClimaAtmos.jl/pull/3542~, ClimaLand https://github.com/CliMA/ClimaLand.jl/pull/996, ClimaCoupler #1160 )


szy21 (approved):

charleskawczynski (approved):

charleskawczynski:
Should we also have a job that runs with the latest versions within the allowable compat? Or is that too much?

juliasloan25:
> Should we also have a job that runs with the latest versions within the allowable compat? Or is that too much?

That could be helpful, but I think running it every day like we do with this pipeline would be too much. I think we update dependencies frequently enough (every 1-2 weeks) that we don't need to add this right now

PR #1163, opened 2025-01-31
move experiment-related tests into experiments/

### LINKED ISSUE #532: replace sea ice test with experiment
We currently have a test for Eisenman sea ice ([eisenman_seaice_tests.jl](https://github.com/CliMA/ClimaCoupler.jl/blob/main/test/component_model_tests/eisenman_seaice_tests.jl)) which constructs and steps a simulation. This requires functionality from SciMLBase, so that package has to be added to the `test/` environment. ~I suggest that we delete this test, and make sure that the [experiment](https://github.com/CliMA/ClimaCoupler.jl/blob/main/.buildkite/pipeline.yml#L171-L176) testing Eisenman sea ice makes sufficient checks.~ We should move all test for for code in `experiments/` to `experiments/`, so that `test/` is only testing code in `src/`, and `experiments/` is self-contained. This way we can remove timestepping-related dependencies (SciMLBase.jl) from `test`, while maintaining that this sea ice model is tested.

### Files to move
- [x] atmos
- [x] eisenman
- [x] prescr_seaice
- [x] prescr_ocean
- [x] slab_ocean
- [x] debug_plots
- [x] coupled_sims

### other
- [x] remove deps from `test/` environment
- [x] remove includes for these tests from `test/runtests.jl`, add them to `experiments/ClimaEarth/test/run_tests.jl`
- [x] move TestHelper.jl into `experiments/ClimaEarth/`
- [ ] add GHA for each set of tests (`experiments/ClimaCore/test/`, `experiments/ClimaEarth/test/`, `test/`

juliasloan25:
## Purpose 
Tests for code in the `experiments/` folder, including for component models and
debug plotting have been moved into the `experiments/ClimaEarth/` and
`experiments/ClimaCore` folders. This allows us to remove packages from the `test/`
environment and keep the `experiments/` environments more self-contained.
Github Actions have been added for the new `experiments/ClimaCore/` tests.

closes #532


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   test:
-    name: ci ${{ matrix.version }} - ${{ matrix.os }}
+    name: ci experiments/ClimaEarth/ ${{ matrix.version }} - ${{ matrix.os }}
```
juliasloan25, line 18: undo name change

Sbozzolo (approved):

PR #1168, opened 2025-02-04
update test compat

juliasloan25:
## Purpose 
The ClimaAtmos-only benchmark runs have been [failing](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/220#_) because the test environment compat wasn't updated correctly when we updated to ClimaTimeSteppers v0.8. This PR fixes the test compat so the test environment can use the latest Atmos and CTS versions


Sbozzolo (approved):

szy21 (approved):

juliasloan25:
I'm merging manually because the MPI AMIP run is holding up CI and this PR doesn't affect that run

PR #1169, opened 2025-02-04
Reduce time for local_checks.sh

Sbozzolo:
This sometimes runs for a very long time

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -320,8 +320,8 @@ steps:
         command: "test/mpi_tests/test_sbatch_script.sh"
         agents:
           slurm_ntasks: 1
+          slurm_time: 30
         soft_fail: true
-        timeout_in_minutes: 40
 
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -322,7 +322,7 @@ steps:
         agents:
           slurm_ntasks: 1
```
nefrathenrici, line 323:
```suggestion
          slurm_ntasks: 1
          slurm_time: 40
```

Sbozzolo: Thank you

nefrathenrici (approved):
Thank you!

juliasloan25:
Why did you change from `timeout_in_minutes` to `slurm_time`? I would have thought they would do the same thing

Sbozzolo:
I realzied there's a problem with this job. When buildkite times out, the associated slurm job is not removed from the queue.

PR #1178, opened 2025-02-06
add wrapper function to driver

### LINKED ISSUE #1177: wrap driver in function for calibration
For calibration, we want to wrap a large chunk of the driver in a single function. Specifically, this function should take in a path to a config file, and contain all the code to initialize component models and run the simulation. The postprocessing code can be inside too, and will just be disabled for calibration runs.

juliasloan25:
## Purpose 
closes #1177

Move most of the driver code after package imports and includes into one function. Also move the `solve_coupler!` function outside of it.

The `use_coupler_diagnostics` flag is now used to determine if any diagnostics are output (default diagnostics, debug plots, leaderboard, and conservation), whereas before the leaderboard and conservation were independent. This makes it easier to turn off all diagnostics with one flag. The leaderboard and conservation flags can still be used to control output of those diagnostics if `use_coupler_diagnostics == true`.

### Content
- [x] wrap everything after the config file read in a function `setup_and_run`
- [x] move the `solve_coupler!` function outside to avoid closure
- [x] `use_coupler_diagnostics` turns off all diagnostics
- [x] verify no behavioral changes


experiments/ClimaEarth/run_amip.jl
```diff
@@ -97,640 +97,654 @@ include("user_io/arg_parsing.jl")
...
+    The postprocessing includes:
+    - Energy and water conservation checks (if running SlabPlanet with checks enabled)
+    - Animations (if not running in MPI)
+    - AMIP plots of the final state of the model
+    - Error against observations
+    - Optional additional atmosphere diagnostics plots
+    - Plots of useful coupler and component model fields for debugging
+    =#
+
+    if ClimaComms.iamroot(comms_ctx)
```
nefrathenrici, line 743: It would be nice to be able to disable postprocessing to save time when calibrating

experiments/ClimaEarth/run_amip.jl
```diff
@@ -97,640 +97,654 @@ include("user_io/arg_parsing.jl")
...
-- Optional additional atmosphere diagnostics plots
-- Plots of useful coupler and component model fields for debugging
-=#
-
-if ClimaComms.iamroot(comms_ctx)
-    postprocessing_vars = (; use_coupler_diagnostics, output_default_diagnostics, t_end, conservation_softfail)
-    postprocess_sim(cs, postprocessing_vars)
-end
+# Set up and run the coupled simulation
+setup_and_run(config_dict)
```
nefrathenrici, line 750: Could we split the function into another file? If I include the file in the calibration workflow for access to the function it will always run the forward model

nefrathenrici (commented):
Thanks! Just a few sticking points

experiments/ClimaEarth/run_amip.jl
```diff
@@ -97,640 +97,654 @@ include("user_io/arg_parsing.jl")
...
-    ## land model
-    land_sim = BucketSimulation(
+"""
+    setup_and_run(config_dict = "../../config/ci_configs/amip_default.yml")
+
+This function sets up and runs the coupled model simulation specified by the
+input config file. It initializes the component models, all coupler objects,
+diagnostics, and conservation checks, and then runs the simulation.
+"""
+function setup_and_run(config_dict = "../../config/ci_configs/amip_default.yml")
```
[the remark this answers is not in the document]
nefrathenrici: I think this is needed to run from any directory

nefrathenrici (approved):
Thank you, this looks great!

PR #1179, opened 2025-02-06
Support identical restarts with JLD2 files

Sbozzolo:
This commit implements a way to restart simulations by saving both state and caches of component models, as well as the coupler fields.

Given that caches are complex object, I implemented this using JLD2 files.

The challenges with JLD2 files are that:
- they are not MPI compatible,
- they are not GPU compatible.

For this reason, I have to move everything to the CPU, and have each process write to its own output. This adds a restriction: only the same number of MPI process (and the same machine) can be used for restarts.

In addition to this, this approach requires component models to implement their functions to restore their caches.

Something that can be improved in the future is that, ClimaAtmos is currently producing two checkpoints, one independently, and one from ClimaCoupler. This should not be needed, but it is currently needed because there's no other way to start ClimaAtmos at a different time.

More over, when restarting, ClimaAtmos `t_end` will be from the `t_start`, which is different from climacoupler `t_end`.

The other problem here is that the MPI test occasionally hangs (as it does in ClimaAtmos).p

Closes #1063


docs/src/checkpointer.md
```diff
@@ -1,12 +1,136 @@
...
+specify how their cache has to be restored.
+
+If you are adding a component model with a cache, you have to extend the
+```
+Checkpointer.get_model_cache
+Checkpointer.restore_cache!
+```
+methods. 
+
+`ClimaCoupler` moves objects to the CPU is done with `Adapt(Array, x)`. `Adapt`
```
juliasloan25, line 117:
```suggestion
`ClimaCoupler` moves objects to the CPU with `Adapt(Array, x)`. `Adapt`
```

docs/src/checkpointer.md
```diff
@@ -1,12 +1,136 @@
...
+traverses the object recursively, and proper `Adapt` methods have to be defined
+for every object involved in the chain. The easiest way to do this is using the
+`Adapt.@adapt_structure` macro, which defines a recursive Adapt for the given
+object.
+
+Types to watch for:
+- `MPI` related objects (e.g., `MPICommsContext`)
+- `TimeVaryingInputs` (because they contain `NCDatasets`)
```
juliasloan25, line 125: Could you add a note explaining why NCDatasets are problematic here?

experiments/ClimaEarth/cli_options.jl
```diff
@@ -85,8 +85,8 @@ function argparse_settings()
         default = nothing
         "--restart_t"
```
juliasloan25, line 86: I can't comment on that line, but could you also delete "hourly" from the description for `checkpoint_dt` in line 78 please?

experiments/ClimaEarth/cli_options.jl
```diff
@@ -85,8 +85,8 @@ function argparse_settings()
...
         help = "Time in seconds rounded to the nearest index to use at `t_start` for restarted simulation [0 (default)]"
```
juliasloan25, line 87: The default has changed from 0 to `nothing`

src/Checkpointer.jl
```diff
@@ -46,30 +56,38 @@ function checkpoint_model_state(
 end
 
 """
-    restart_model_state!(sim::Interfacer.ComponentModelSimulation, comms_ctx::ClimaComms.AbstractCommsContext, t::Int; input_dir = "input")
+    checkpoint_model_cache(sim::Interfacer.ComponentModelSimulation, comms_ctx::ClimaComms.AbstractCommsContext, t::Int; output_dir = "output")
 
-Sets the model state of a simulation from a HDF5 file from a given time, t (in seconds).
+Checkpoints the model cache to a N JLD2 files at a given time, t (in seconds),
```
juliasloan25, line 61:
```suggestion
Checkpoints the model cache to N JLD2 files at a given time, t (in seconds),
```

src/Checkpointer.jl
```diff
@@ -35,8 +46,7 @@ function checkpoint_model_state(
     day = floor(Int, t / (60 * 60 * 24))
     sec = floor(Int, t % (60 * 60 * 24))
     @info "Saving checkpoint " * Interfacer.name(sim) * " model state to HDF5 on day $day second $sec"
-    mkpath(joinpath(output_dir, "checkpoint"))
-    output_file = joinpath(output_dir, "checkpoint", "checkpoint_" * Interfacer.name(sim) * "_$t.hdf5")
+    output_file = joinpath(output_dir, "checkpoint_" * Interfacer.name(sim) * "_$t.hdf5")
     checkpoint_writer = CC.InputOutput.HDF5Writer(output_file, comms_ctx)
```
juliasloan25, line 50: Why do you use `HDF5Writer` here but `jldsave` in `checkpoint_model_cache` below?

juliasloan25: I'm also unsure of why we don't need to `adapt` here. Is it because the structure of Y is simpler than p?

src/Checkpointer.jl
```diff
@@ -78,18 +96,188 @@ end
...
+                Int((t - t0) / 1e3),
+                output_dir = cs.dirs.checkpoints,
+            )
+        end
+    end
+
+    # Checkpoint the Coupler fields
+    output_dir = cs.dirs.checkpoints
+    comms_ctx = cs.comms_ctx
+    time = Int((t - t0) / 1e3)
```
juliasloan25, line 121: Could you move these above the for loop and use them uniformly for the calls to `checkpoint_model_state`, `checkpoint_model_cache`, and the coupler fields save?

src/Checkpointer.jl
```diff
@@ -78,18 +96,188 @@ end
...
+
+Find checkpoints in `checkpoint_dir` for the coupler fields and write them in `cs`.
+"""
+function auto_restart_coupler_fields!(cs, checkpoint_dir)
+    isdir(checkpoint_dir) || return false
+    pid = ClimaComms.mypid(cs.comms_ctx)
+    restart_file_rx = Regex("checkpoint_coupler_fields_$(pid)_\\d+.jld2")
+    restarts = filter(f -> !isnothing(match(restart_file_rx, f)), readdir(checkpoint_dir))
+    isempty(restarts) && return false
+    input_file = joinpath(checkpoint_dir, last(sort_by_creation_time(restarts)))
```
juliasloan25, line 186: Could you explain the behavior more in the docstring please? E.g. if there are multiple files in the checkpoint directory matching the expected format, the most recent file will be used to restart

(same comment applies to `auto_restart_model_cache!`, `auto_restart_model_state!`)

src/Checkpointer.jl
```diff
@@ -78,18 +96,188 @@ end
...
+    @info "Restarting $(Interfacer.name(sim)) cache from $input_file"
+    return true
+end
+
+function restart_model_cache!(sim, input_file)
+    ispath(input_file) || error("File $(input_file) not found")
+    restore_cache!(sim, JLD2.jldopen(input_file)["cache"])
```
juliasloan25, line 214: It would be helpful to note here again that `restore_cache!` will be implemented by component models with caches, so readers know where to look for the implementation

experiments/ClimaEarth/setup_run.jl
```diff
@@ -209,8 +210,7 @@ function setup_and_run(config_dict::AbstractDict)
     #and `dir_paths.checkpoints`, where restart files are saved.
     =#
 
-    COUPLER_OUTPUT_DIR = joinpath(output_dir_root, job_id)
-    dir_paths = Utilities.setup_output_dirs(output_dir = COUPLER_OUTPUT_DIR, comms_ctx = comms_ctx)
+    dir_paths = Utilities.setup_output_dirs(output_dir = output_dir_root, comms_ctx = comms_ctx)
```
juliasloan25, line 213: It would be good to avoid changing the output paths if possible

experiments/ClimaEarth/setup_run.jl
```diff
@@ -90,14 +90,15 @@ and exchanges combined fields and calculates fluxes using the selected turbulent
...
-
 function solve_coupler!(cs)
     (; model_sims, Δt_cpl, tspan, comms_ctx) = cs
     (; atmos_sim, land_sim, ocean_sim, ice_sim) = model_sims
 
     @info("Starting coupling loop")
     ## step in time
     for t in ((tspan[begin] + Δt_cpl):Δt_cpl:tspan[end])
+        # Main.@infiltrate
+
```
juliasloan25, line 101:
```suggestion
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -231,6 +231,22 @@ function setup_and_run(config_dict::AbstractDict)
     Random.seed!(random_seed)
     @info "Random seed set to $(random_seed)"
 
+    isnothing(restart_t) && (restart_t = Checkpointer.t_start_from_checkpoint(dir_paths.checkpoints))
+    isnothing(restart_dir) && (restart_dir = dir_paths.checkpoints)
+    if !isnothing(restart_t)
+        if t_start isa ITime
+            t_start, _ = promote(ITime(restart_t), t_start)
+        else
+            t_start = maybe_t_start_from_checkpoint
```
juliasloan25, line 240: I don't see `maybe_t_start_from_checkpoint` defined anywhere

experiments/ClimaEarth/setup_run.jl
```diff
@@ -667,17 +684,20 @@ function setup_and_run(config_dict::AbstractDict)
...
-    cs.tspan[2] = tspan[2]
+        ## update the timespan to the correct values
+        cs.tspan[1] = tspan[1] + Δt_cpl * 2
+        cs.tspan[2] = tspan[2]
+    end
 
     ## Run garbage collection before solving for more accurate memory comparison to ClimaAtmos
     GC.gc()
 
+
```
juliasloan25, line 700:
```suggestion
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -736,6 +756,8 @@ function setup_and_run(config_dict::AbstractDict)
     end
 
     # Close all diagnostics file writers
+
```
juliasloan25, line 759:
```suggestion
```

experiments/ClimaEarth/run_amip.jl
```diff
@@ -22,4 +22,4 @@ include("setup_run.jl")
 config_file = parse_commandline(argparse_settings())["config_file"]
 
 # Set up and run the coupled simulation
-setup_and_run(config_file)
+cs = setup_and_run(config_file)
```
juliasloan25, line 25: why return `cs`?

docs/src/checkpointer.md
```diff
@@ -1,12 +1,136 @@
...
+
+`ClimaCore.InputOutput` provides a loss-less way to save the content of certain
+`ClimaCore` objects to HDF5 files. Objects saved in this way are not tied to a
+particular computing device or configuration. When running with MPI,
+`ClimaCore.InputOutput` are also efficiently written in parallel.
+
+Unfortunately, `ClimaCore.InputOutput` only supports certain objects, such as
+`Field`s and `Space`s, but the cache in component models is more complex than
+this and contains complex objects with highly stateful quantities (e.g., C
+pointers).
```
juliasloan25, line 91: After re-reading, I understand that this is why we save the state to HDF5 and cache to JLD2 (I left a comment about this somewhere). It would be good to say that explicitly here.
```suggestion
Unfortunately, `ClimaCore.InputOutput` only supports certain objects, such as
`Field`s and `Space`s, but the cache in component models is more complex than
this and contains complex objects with highly stateful quantities (e.g., C
pointers). Because of this, model states are saved to HDF5 but caches must be
saved to JLD2 files.
```

juliasloan25 (commented):
I haven't had a chance to look at the atmos/bucket or test files yet, but here's my initial review.

One question that came to mind is why `Checkpointer.get_model_cache` and `Checkpointer.restart_model_cache!` don't need to be extended for the non-atmos/land models we're running (e.g. `PrescribedIceSimulation`, which does have a cache). I would expect that we need those methods, but the tests seem to pass without them.

docs/src/checkpointer.md
Sbozzolo: Added

experiments/ClimaEarth/run_amip.jl
Sbozzolo: Because I need a way to access the fields

experiments/ClimaEarth/setup_run.jl
Sbozzolo: I removed all the `auto_restart` infra because I can handle differently, in a simpler way

src/Checkpointer.jl
Sbozzolo: I removed this function because at the end of the day I could accomplish the same behavior with the `t_start_from_checkpoint` function

src/Checkpointer.jl
Sbozzolo: Added in docstring and comment

experiments/ClimaEarth/setup_run.jl
Sbozzolo: Do we need to have a different `COUPLER_OUTPUT_DIR` and `output_dir_root`?

experiments/ClimaEarth/setup_run.jl
```diff
@@ -90,7 +90,6 @@ and exchanges combined fields and calculates fluxes using the selected turbulent
 Note that we want to implement this in a dispatchable function to allow for
 other forms of timestepping (e.g. leapfrog).
 """
-
```
Sbozzolo, line 93: Removed this because docstrings have to next to their functions

Sbozzolo:
> I haven't had a chance to look at the atmos/bucket or test files yet, but here's my initial review.
> 
> One question that came to mind is why `Checkpointer.get_model_cache` and `Checkpointer.restart_model_cache!` don't need to be extended for the non-atmos/land models we're running (e.g. `PrescribedIceSimulation`, which does have a cache). I would expect that we need those methods, but the tests seem to pass without them.

Great question! The answer is that the PrescribedIce/Ocean are not evolving independently. When uncoupled, they don't really have a cache because everything is determined by the time. When coupled, they don't really have a cache either, because all the fields that are not default are updated using values from the coupled_fields. which is stored to disk.

docs/src/checkpointer.md
```diff
@@ -1,12 +1,136 @@
...
+useful to split a long simulation into smaller, more manageable chunks.
+
+Checkpoints are a mix of HDF5 and JLD2 files and are typically saved in a
+`checkpoints` folder in the simulation output. See
+[`Utilities.setup_output_dirs`](@ref) for more information.
+
+!!! known limitations
+
+    - The number of MPI processes has to remain the same across checkpoints
+    - Restart files are generally not portable across machines
```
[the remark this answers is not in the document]
Sbozzolo: Added

.buildkite/pipeline.yml
```diff
@@ -109,6 +100,25 @@ steps:
...
+        agents:
+          slurm_mem: 16GB
+
+      - label: "MPI restarts"
+        key: "mpi_restarts"
+        command: "srun julia --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/test/restart.jl"
+        env:
+          CLIMACOMMS_CONTEXT: "MPI"
+        timeout_in_minutes: 40
+        soft_fail: true         # The MPI test sometimes hangs
```
charleskawczynski, line 117: Is this hanging because of code introduced here? Is this something that should be fixed?

experiments/ClimaEarth/components/atmosphere/climaatmos_recursive.jl
```diff
@@ -0,0 +1,76 @@
...
+import ClimaCore.Fields: Field, FieldVector, field_values
+import ClimaCore.DataLayouts: AbstractData
+import ClimaCore.Geometry: AxisTensor
+import ClimaCore.Spaces: AbstractSpace
+import ClimaUtilities.TimeVaryingInputs: AbstractTimeVaryingInput
+import NCDatasets
+
+# Allow cache to be moved to CPU (this is a little bit of type piracy...)
+ClimaCore.Adapt.@adapt_structure CA.AtmosCache
+ClimaCore.Adapt.@adapt_structure CA.RRTMGPInterface.RRTMGPModel
```
charleskawczynski, line 14: We should not be doing type piracy like this. Please remove this.

experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -407,6 +408,63 @@ function make_land_domain(
...
+    return sim.integrator.p
+end
+
+function Checkpointer.restore_cache!(sim::BucketSimulation, new_cache)
+    old_cache = Checkpointer.get_model_cache(sim)
+    comms_ctx = ClimaComms.context(sim.model)
+    recursively_reset!(old_cache, new_cache, comms_ctx)
+end
+
+function recursively_reset!(
```
charleskawczynski, line 421: What are all these restore methods? And why do they have slightly different names? And why are they using `propertynames` instead of `fieldnames`?

experiments/ClimaEarth/components/atmosphere/climaatmos_recursive.jl
Sbozzolo: The reason I left them here is because this entire `atmosphere` module in ClimaEarth is intended to be something that ClimaAtmos should provide at some point (in an extension for ClimaCoupler, or in its source code). The same is true for the other modules in the `ClimaEarth/components` folder, and this is not the only type piracy that we have in there.

I am happy to move this directly to ClimaAtmos now.

.buildkite/pipeline.yml
Sbozzolo: This is hanging for the same reason the equialent test in ClimaAtmos hangs: sometimes, things hang when multiple Atmos simulations are started in the same REPL (with MPI).

I tried debugging/working around this in the past (e.g., creating one MPI communicator per simulation), and I tried again, but I could not figure out what's wrong. 

I would like to see this fixed.

experiments/ClimaEarth/components/land/climaland_bucket.jl
Sbozzolo: The entry point is `restore_model_cache!`, which:
```
    restart_model_cache!(sim, input_file)

Overwrite the content of `sim` with the cache from the `input_file`.

It relies on `restore_cache!(sim, old_cache)`, which has to be implemented by
the component models that have a cache.
```

Each component model has to implement its way to restore the cache from a JLD2 file, so the bucket component implements its method:
```
function Checkpointer.restore_cache!(sim::BucketSimulation, new_cache)
    old_cache = Checkpointer.get_model_cache(sim)
    comms_ctx = ClimaComms.context(sim.model)
    recursively_reset!(old_cache, new_cache, comms_ctx)
end
```

`recursively_reset!` then calls `recursively_reset_base!` for the base case of the recursion, or to exclude specific types.

> And why are they using propertynames instead of fieldnames?

Because in general there are NamedTuples in the tree we are traversing.

charleskawczynski: > Because in general there are NamedTuples in the tree we are traversing.

Why is that an issue?

Sbozzolo: > > Because in general there are NamedTuples in the tree we are traversing.
> 
> Why is that an issue?

Because `fieldnames` is not defined on NamedTuples

```julia-repl
julia> fieldnames((; a = 1))
ERROR: MethodError: no method matching fieldnames(::@NamedTuple{a::Int64})
The function `fieldnames` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  fieldnames(::Core.TypeofBottom)
   @ Base reflection.jl:254
  fieldnames(::Type{<:Tuple})
   @ Base reflection.jl:256
  fieldnames(::UnionAll)
   @ Base reflection.jl:253
```

charleskawczynski: That's not how `fieldnames` works:
```julia
julia> fieldnames(typeof((; a = 1)))
(:a,)
```

Sbozzolo: Aaah! I always used `propertynames` I thought that `fieldnames` were only for named structs.

When should one use one vs the other? I see that we use both in our codes

charleskawczynski: The idea is that `getproperty` (along with `propertynames`) is meant to be overloadable by users, the default behavior is to return `getfield` (based on `fieldnames`).

That said, for something like serialization, I think recursing through `getfield` / `fieldnames` should be much more generic, as these will recurse directly into the datalayouts arrays, and bypass the field variables, which should simplify that logic.

charleskawczynski: We use `propertynames` and `getproperty` a lot for user-facing ClimaCore code because DataLayouts overloads `propertynames` and `getproperty` to appear like a NamedTuple

Sbozzolo: I switched to using `fieldnames`, which allowed me to condense some of the restore methods and remove the `_base` ones. I took the opportunity to unify them across ClimaLand and ClimaAtmos.

juliasloan25:
> Something that can be improved in the future is that, ClimaAtmos is currently producing two checkpoints, one independently, and one from ClimaCoupler. This should not be needed, but it is currently needed because there's no other way to start ClimaAtmos at a different time.

Could you explain this a little more? Why does saving two checkpoints allow us to start ClimaAtmos at a different time? Also, what do you mean by "different time"?

In general, it would be good to open an issue listing next steps to improve the restarts, and link that issue from this PR description

experiments/ClimaEarth/components/ocean/prescr_ocean.jl
```diff
@@ -112,3 +112,17 @@ at each timestep.
...
+
+function Checkpointer.restore_cache!(sim::PrescribedOceanSimulation, new_cache)
+    old_cache = Checkpointer.get_model_cache(sim)
+    for p in propertynames(old_cache)
+        if getproperty(old_cache, p) isa Field
+            ArrayType = ClimaComms.array_type(getproperty(old_cache, p))
+            parent(getproperty(old_cache, p)) .= ArrayType(parent(getproperty(new_cache, p)))
+        end
+    end
+end
```
juliasloan25, line 128: This function is identical for `PrescribedOceanSimulation` and `PrescribedIceSimulation`, which both subtype `AbstractSurfaceStub`. It could be defined for `AbstractSurfaceStub` in src/surface_stub.jl instead

experiments/ClimaEarth/components/shared/restore.jl
```diff
@@ -0,0 +1,77 @@
...
+import ClimaCore: DataLayouts, Fields, Geometry
+import ClimaCore.Fields: Field, FieldVector, field_values
+import ClimaCore.DataLayouts: AbstractData
+import ClimaCore.Geometry: AxisTensor
+import ClimaCore.Spaces: AbstractSpace
+import ClimaUtilities.TimeVaryingInputs: AbstractTimeVaryingInput
+import StaticArrays
+import NCDatasets
+
+function restore!(v1::T1, v2::T2, comms_ctx; name = "", ignore) where {T1, T2}
```
juliasloan25, line 14: Can you add more documentation of the expected behavior of this function?

experiments/ClimaEarth/setup_run.jl
juliasloan25: No - this is fine with the change you made in `arg_parsing.jl`!

experiments/ClimaEarth/setup_run.jl
```diff
@@ -231,6 +232,22 @@ function setup_and_run(config_dict::AbstractDict)
     Random.seed!(random_seed)
     @info "Random seed set to $(random_seed)"
 
+    isnothing(restart_t) && (restart_t = Checkpointer.t_start_from_checkpoint(dir_paths.checkpoints))
+    isnothing(restart_dir) && (restart_dir = dir_paths.checkpoints)
+    if !isnothing(restart_t)
```
juliasloan25, line 237: Should this be
```suggestion
    if should_restart
```
where `should_restart = !isnothing(restart_t) && !isnothing(restart_dir)` is defined lower

docs/src/checkpointer.md
```diff
@@ -1,12 +1,136 @@
...
+  the number of MPI processes cannot change across restarts.
+3. Some quantities are best not saved and read (for example, anything with
+  pointers). For this, we write a recursive function that traverses the cache
+  and only restores quantities of a certain type (typically, `ClimaCore`
+  objects)
+
+Point 3. adds significant amount of code and requires component models to
+specify how their cache has to be restored.
+
+If you are adding a component model with a cache, you have to extend the
```
juliasloan25, line 110: If a user adds a component model and wants to checkpoint, they'll also need to extend `Checkpointer.get_model_prog_state`. I know this was an existing function before this PR, but could you mention this in the docs please?

experiments/ClimaEarth/components/atmosphere/climaatmos_recursive.jl
juliasloan25: added to ClimaAtmos in https://github.com/CliMA/ClimaAtmos.jl/pull/3673

juliasloan25 (approved):
Thank you!! I only had some minor comments

experiments/ClimaEarth/components/shared/restore.jl
Sbozzolo: Added

Sbozzolo:
> > Something that can be improved in the future is that, ClimaAtmos is currently producing two checkpoints, one independently, and one from ClimaCoupler. This should not be needed, but it is currently needed because there's no other way to start ClimaAtmos at a different time.
> 
> Could you explain this a little more? Why does saving two checkpoints allow us to start ClimaAtmos at a different time? Also, what do you mean by "different time"?
> 
> In general, it would be good to open an issue listing next steps to improve the restarts, and link that issue from this PR description

ClimaAtmos can either start a simulation from t = 0, or from the t read from a checkpoint. So, to align the time in component models, I am reading the checkpoint.

https://github.com/CliMA/ClimaCoupler.jl/issues/1218

> Is this still true? I noticed that you added get_model_cache and restore_cache! for them now. I would expect that even if the cache can be determined by the time and coupler_fields alone, when restarting you either need to read in the saved model cache, or evaluate/step the model once to set the correct values from t/coupler fields.

Yes indeed, it is not needed if we exchange fields, but the way I had originally implmented it was not doing the fields exchanges, so I saved the caches for those too.

Issue #1183, opened 2025-02-07
AMIP breaks with deep atmosphere


szy21:
The nightly AMIP breaks pretty fast with deep atmosphere: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/249. I set it to shallow atmosphere now. It would be good to understand why. The higher resolution AMIP may break as well but I will change it if that's the case.

We should check if some of the atmos longrun also break.

@akshaysridhar Could you take this on?

PR #1184, opened 2025-02-07
use shallow atmosphere in amip

szy21:
## Purpose 

The nightly AMIP breaks with deep atmosphere, so I set it to shallow atmosphere now. See #1183 
shallow atmosphere build: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/250

## Content

----


juliasloan25 (approved):
In the long term, will we want to use shallow or deep atmosphere?

akshaysridhar:
> In the long term, will we want to use shallow or deep atmosphere?

Deep atmospheres (this is now default in the atmos configurations)

akshaysridhar (approved):

PR #1187, opened 2025-02-11
Add AMIP calibration pipeline

nefrathenrici:
## Purpose 

Add the initial AMIP calibration pipeline with a trivial perfect model experiment.

## Content

----


experiments/calibration/run_calibration.jl
```diff
@@ -0,0 +1,124 @@
...
+ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
+ClimaComms.@import_required_backends
+
+single_member_dims = (1,)
+function CAL.observation_map(iteration)
+    G_ensemble = Array{Float64}(undef, single_member_dims..., ensemble_size)
+
+    for m in 1:ensemble_size
+        member_path = CAL.path_to_ensemble_member(output_dir, iteration, m)
+        simdir_path = joinpath(member_path, "model_config/output_active/clima_atmos")
```
juliasloan25, line 19: I don't think you need to go into the `model_config/` directory here, but I'm not sure what `member_path` returns so it may be correct
```suggestion
        simdir_path = joinpath(member_path, "output_active/clima_atmos")
```

juliasloan25 (approved):
Looks good! I left some comments about the paths

experiments/calibration/run_calibration.jl
nefrathenrici: I think the `model_config` comes from the name of the config file (`model_config.yml`). I could change this with the `job_id`, it would probably be good to keep a consistent job_id for calibration runs.

experiments/calibration/model_interface.jl
```diff
@@ -0,0 +1,17 @@
+import ClimaCoupler
+import ClimaCalibrate
+include(joinpath(pkgdir(ClimaCoupler), "experiments", "ClimaEarth", "setup_run.jl"))
+
+function ClimaCalibrate.forward_model(iter, member)
+    config_file = joinpath(ClimaCalibrate.project_dir(), "model_config.yml")
+    config_dict = get_coupler_config_dict(config_file)
+
+    output_dir_root = config_dict["coupler_output_dir"]
```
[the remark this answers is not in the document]
nefrathenrici: I think that I am taking this into account in the `observation_map` func. From previous sims I see that the moddel output is already in the format described in the NEWS.md section!

experiments/calibration/run_calibration.jl
juliasloan25: If `job_id = model_config` this path will be right, but I'm not sure `job_id` is being set (the default is nothing)

nefrathenrici: The job_id gets set to the name of the YAML file, I don't think it is ever `nothing` after the `get_coupler_config_dict` func:
`    job_id = isnothing(job_id) ? string(split(split(config_file, '/')[end], '.')[1]) : job_id`

https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/user_io/arg_parsing.jl#L26 

Since this is only manually tested I would like to run the calibration once a week, should I make a new pipeline or do you think we could put this in the longruns?

juliasloan25: Ahh yes you're right, I missed that line.

Let's add it to the longruns since we already have a lot of coupler pipelines

PR #1189, opened 2025-02-13
fig config diagnostic merging; print only on root

juliasloan25:
## Purpose 
This PR does 2 distinct things:
1. Fix merging of diagnostics specified in the coupler and atmosphere. Previously, `atmos_config_dict["diagnostics"]` was overwritten by the coupler `config_dict["extra_atmos_diags"]`. Now these additional diagnostics are appended, so we output all diagnostics.
2. Initialize the ClimaComms communication context immediately after creating the `config_dict`, so all printing happens on root in the MPI case

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -356,8 +356,12 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
     atmos_config["output_dir_style"] = "RemovePreexisting"
     atmos_config["output_dir"] = atmos_output_dir
 
-    # Access extra atmosphere diagnostics from coupler so we can rename for atmos code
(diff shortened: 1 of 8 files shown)


akshaysridhar (approved):

akshaysridhar:
Thanks @juliasloan25 . Can confirm following this change that diagnostics other than the default set are parsed correctly following the `map` change.

akshaysridhar:
@juliasloan25 Would splitting up the 2 commits help with this ? It would be helpful to have at least the diagnostic fix as part of the `main` branch.

juliasloan25:
> @juliasloan25 Would splitting up the 2 commits help with this ? It would be helpful to have at least the diagnostic fix as part of the `main` branch. (I've manually included the diagnostic update commit for now)

Sorry for the delay - there was a bug with the diagnostic fix that only appeared with some configs. It is fixed now in this PR

PR #1191, opened 2025-02-13
Add ITime

ph-kev:
## Purpose 
This PR adds `ITime` which is needed to do long run without floating point errors in time.

## To-do
- [x] Verify that the first few steps of a long run does not fail (see this [buildkite](https://buildkite.com/clima/climacoupler-longruns/builds/835#_))

One thing to note is that depending on the `dt` passed, it automatically defaults to the reasonable value for the period. This means that if `dt` is 120 seconds (as seen [here](https://buildkite.com/clima/climacoupler-longruns/builds/835#0195011d-c12a-4fdf-9518-e44f82e1334a)), then the period defaults to one minute which may or may not be desirable. This is not specific to ClimaCoupler, but also affects ClimaAtmos as well.

### DIFF
--- a/experiments/ClimaCore/sea_breeze/run.jl
+++ b/experiments/ClimaCore/sea_breeze/run.jl
@@ -117,7 +117,7 @@ system before executing the simulation.
 
 t_start, t_end = (0.0, 1e4)
 Δt_coupled = 0.1
-saveat = t_start:10.0:t_end
+saveat = collect(t_start:10.0:t_end)
 atm_nsteps, ocn_nsteps, lnd_nsteps = (5, 1, 1)
 
 ## Initialize Models
--- a/experiments/ClimaEarth/cli_options.jl
+++ b/experiments/ClimaEarth/cli_options.jl
@@ -159,6 +159,10 @@ function argparse_settings()
         help = "Boolean flag indicating whether to compute and output land model diagnostics [`true` (default), `false`]"
         arg_type = Bool
         default = true
+        "--use_itime"
+        help = "Boolean flag indicating whether to use ITime (integer time) or not (will use Float64) [`true` (default), `false`]"
+        arg_type = Bool
+        default = true
     end
     return s
 end
--- a/experiments/ClimaEarth/components/ocean/eisenman_seaice.jl
+++ b/experiments/ClimaEarth/components/ocean/eisenman_seaice.jl
@@ -44,7 +44,7 @@ Base.@kwdef struct EisenmanOceanParameters{FT <: AbstractFloat}
 end
 
 """
-    EisenmanIceSimulation(::Type{FT}, tspan; space = nothing, area_fraction = nothing, thermo_params = nothing, stepper = CTS.RK4(), dt = 0.02, saveat = 1.0e10)
+    EisenmanIceSimulation(::Type{FT}, tspan; space = nothing, area_fraction = nothing, thermo_params = nothing, stepper = CTS.RK4(), dt, saveat = 1.0e10)
 
 Initialize the Eisenman-Zhang sea ice model and simulation.
 """
@@ -55,7 +55,7 @@ function EisenmanIceSimulation(
     area_fraction = nothing,
     thermo_params = nothing,
     stepper = CTS.RK4(),
-    dt = 0.02,
+    dt,
     saveat = [1.0e10],
 ) where {FT}
 
@@ -79,8 +79,13 @@ function EisenmanIceSimulation(
         thermo_params = thermo_params,
         dss_buffer = CC.Spaces.create_dss_buffer(Y),
     )
-    problem = SciMLBase.ODEProblem(ode_function, Y, Float64.(tspan), cache)
-    integrator = SciMLBase.init(problem, ode_algo, dt = Float64(dt), saveat = Float64.(saveat), adaptive = false)
+    if typeof(dt) isa Number
+        dt = Float64(dt)
+        tspan = Float64.(tspan)
+        saveat = Float64.(saveat)
+    end
+    problem = SciMLBase.ODEProblem(ode_function, Y, tspan, cache)
+    integrator = SciMLBase.init(problem, ode_algo, dt = dt, saveat = saveat, adaptive = false)
 
     sim = EisenmanIceSimulation(params, space, integrator)
     return sim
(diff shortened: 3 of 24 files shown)


Sbozzolo:
I wonder if we should only support `ITime` and drop support for float. At the end, all the simulations we'll run here are realistic simulations with calendars.

@juliasloan25, what do you think?

juliasloan25:
> I wonder if we should only support `ITime` and drop support for float. At the end, all the simulations we'll run here are realistic simulations with calendars.
> 
> @juliasloan25, what do you think?

All runs in experiments/ClimaEarth/ will be realistic with calendars. One of the goals of ClimaCoupler is to allow coupling across a wide range of model complexities, including e.g. the simple heat diffusion example in experiments/ClimaCore,. I think there's value in continuing to support Floats for simple setups like this for educational/experimental purposes, if it isn't too much trouble.

experiments/ClimaEarth/run_dry_held_suarez.jl
```diff
@@ -136,6 +145,12 @@ thermo_params = get_thermo_params(atmos_sim)
 ## init a 2D boundary space at the surface
 boundary_space = ClimaCore.Spaces.horizontal_space(atmos_sim.domain.face_space)
 
+## Convert Δt_cpl and tspan to ITime
+if use_itime
+    Δt_cpl, t0, tf = promote(ITime(Δt_cpl), tspan[1], tspan[2])
+    tspan = (t0, tf)
+end
```
Sbozzolo, line 152: In the future, we can think of how to avoid having to do this/improve this

Sbozzolo (approved):

experiments/ClimaCore/heat-diffusion/run.jl
```diff
@@ -241,7 +241,13 @@ function coupler_solve!(stepping, ics, parameters)
     # `saveat` determines the frequency with which the solution is saved
     # `adaptive` determines whether the solver should use an adaptive timestep
     prob_atm = SciMLBase.ODEProblem(ode_function, Y_atm, (t_start, t_end), (parameters = parameters, T_sfc = atm_T_lnd))
-    integ_atm = SciMLBase.init(prob_atm, ode_algo, dt = Δt_min, saveat = t_start:(10Δt_min):t_end, adaptive = false)
+    integ_atm = SciMLBase.init(
+        prob_atm,
+        ode_algo,
+        dt = Δt_min,
+        saveat = [t_start:(10 * Δt_min):t_end..., t_end],
```
juliasloan25, line 248: Why does `saveat` need to change?

experiments/ClimaCore/sea_breeze/run.jl
```diff
@@ -117,7 +117,7 @@ system before executing the simulation.
 
 t_start, t_end = (0.0, 1e4)
 Δt_coupled = 0.1
-saveat = t_start:10.0:t_end
+saveat = 10.0
```
juliasloan25, line 120:
```suggestion
saveat = collect(t_start:10.0:t_end)
```
then just pass this to the component models and we don't need to change the atmos/land/ocean_rhs.jl files.

experiments/ClimaEarth/cli_options.jl
```diff
@@ -159,6 +159,10 @@ function argparse_settings()
         help = "Boolean flag indicating whether to compute and output land model diagnostics [`true` (default), `false`]"
         arg_type = Bool
         default = true
+        "--use_itime"
+        help = "Boolean flag indicating whether to use ITime (integer time) or not [`true` (default), `false`]"
```
juliasloan25, line 163:
```suggestion
        help = "Boolean flag indicating whether to use ITime (integer time) or not (will use Float64)  [`true` (default), `false`]"
```
It would be good to somehow make it clear that Float64 is the alternative, maybe like this?

experiments/ClimaEarth/components/ocean/eisenman_seaice.jl
```diff
@@ -79,8 +79,13 @@ function EisenmanIceSimulation(
         thermo_params = thermo_params,
         dss_buffer = CC.Spaces.create_dss_buffer(Y),
     )
-    problem = SciMLBase.ODEProblem(ode_function, Y, Float64.(tspan), cache)
-    integrator = SciMLBase.init(problem, ode_algo, dt = Float64(dt), saveat = Float64.(saveat), adaptive = false)
+    if typeof(dt) isa AbstractFloat
+        dt = Float64(dt)
+        tspan = Float64.(tspan)
+        saveat = Float64.(saveat)
+    end
```
juliasloan25, line 86: Should we instead always convert `span` and `saveat` to the type of `dt` so we always use consistent time types?
```suggestion
    tspan = typeof(dt).(tspan)
    saveat = typeof(dt).(saveat)
```

juliasloan25: We can also remove the default `dt` value from this function since we always pass it in

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -60,9 +60,20 @@ function get_coupler_args(config_dict::Dict)
...
     saveat = Float64(Utilities.time_to_seconds(config_dict["dt_save_to_sol"]))
-    saveat = [t_start:saveat:t_end..., t_end]
-    component_dt_dict = config_dict["component_dt_dict"]
-
+    saveat = promote(t_start:saveat:t_end..., t_end)
+    if use_itime
+        t_end = ITime(t_end, epoch = date0)
+        t_start = ITime(t_start, epoch = date0)
+        Δt_cpl = ITime(Δt_cpl, epoch = date0)
+        times = promote(t_end, t_start, Δt_cpl, ITime.(values(config_dict["component_dt_dict"]))...)
```
juliasloan25, line 68: does `values(config_dict["component_dt_dict"]))` need to be promoted here since it is later promoted in line 72?

ph-kev: The promotion is not done in place.

experiments/ClimaEarth/components/ocean/eisenman_seaice.jl
ph-kev: If `dt` is an `ITime`, then `tspan = typeof(dt).(tspan)` doesn't work.

experiments/ClimaEarth/user_io/arg_parsing.jl
juliasloan25: What I meant is, `times[4:6]` are never accessed, so we don't need to pass `values(config_dict["component_dt_dict"])` here

ph-kev: For line 68, we do need to pass in `ITime.(values(config_dict["component_dt_dict"]))` because it might change the types of `t_end`, `t_start`, and `Δt_cpl`.

juliasloan25: Okay, thanks for clarifying

juliasloan25 (approved):

PR #1192, opened 2025-02-13
Add AMIP with EDMF without MF

Sbozzolo:
AMIP with EDMF without mass flux (with tke based diffusion instead)

Requires ClimaAtmos 0.28.5


config/amip_configs/amip_edonly.yml
```diff
@@ -0,0 +1,29 @@
+FLOAT_TYPE: "Float32"
+albedo_model: "CouplerAlbedo"
+atmos_config_file: "config/longrun_configs/amip_target_edonly.yml"
+checkpoint_dt: "720hours"
+co2: "maunaloa"
+coupler_toml: ["toml/amip.toml"]
```
szy21, line 6: This file has many parameters you don't need, but it shouldn't be a problem I think.

szy21 (approved):

Sbozzolo:
Simulation is stable for 3 years

szy21:
That's surprising:)

PR #1199, opened 2025-02-19
add ClimaLandSimulation struct

### LINKED ISSUE #1182: add `ClimaLandModel` struct and interfacing functions
To run a coupled simulation with the full land model, we need a model struct to use for the full land model, analogous to `BucketModel` or `ClimaAtmosModel`. We should try implementing this and all the adapter functions required by the coupler, and see if any roadblocks come up.

### Fields to pass to land (`update_field!`)
- area_fraction
- liquid_precip
- snow_precip
- turbulent_energy_flux (3-tuple)
- turbulent_moisture_flux (3-tuple)
- SW_d
- LW_d
- cos(\theta s)
- diffuse_fraction
- c_co2

### Fields to get from land (`get_field`)
- area_fraction
- beta
  - for CombinedStateFluxes (which we may not support with the full land model at first anyway)
- z0b (3-tuple)
- z0m (3-tuple)
- surface_direct_albedo
- surface_diffuse_albedo
- surface_humidity
  - for existing PartitionedStateFluxes implementation (with our planned flux approach we won't need this)
- surface_temperature (4-tuple)
- emissivity
- displacement height

imreddyTeja:
## Purpose 
closes #1182

Adds a struct `ClimaLandSimulation` for the full, integrated land model. This PR also extends the necessary Interfacer functions for a simulation type (`get_field`, `update_field!`, `step!`, etc).

This PR requires https://github.com/CliMA/ClimaLand.jl/pull/1035 from ClimaLand, so it uses that branch in the `experiments/ClimaEarth/` environment. That PR will need to be merged and a release of ClimaLand made before we can merge this PR. - This is done as of 3/13/25

### Content
- [x] add `get_field` functions
- [x] add `update_field!` functions
- [x] add method for `add_coupler_fields!`, following https://github.com/CliMA/ClimaCoupler.jl/pull/1207
  - coupler exchange fields when running with the full land model will include the defaults 
    - `[:z0m_sfc, :z0b_sfc, :beta, :emissivity, :F_turb_energy, :F_turb_moisture, :F_turb_ρτxz, :F_turb_ρτyz, :temp1, :temp2]` 
  - and land model-specific quantities 
    - `[:SW_d, :LW_d, :cos_zenith_angle, :diffuse_fraction, :c_co2, :d_sfc, :P_air, :T_air, :q_air]`
- [x] add test to check that constructor is initialized correctly and can be stepped

### Next steps
This PR adds a lot of the software infrastructure to set up and run a `LandModel` in a coupled simulation, but doesn't actually run a coupled simulation with it. Next, we'll update partitioned turbulent fluxes to support the integrated land model, and then we can try using it in a coupled simulation.


experiments/ClimaEarth/components/land/climaland_bucket.jl
juliasloan25, line 1: I made some small changes to the bucket file as I went along, mostly to unify conventions between the two ClimaLand simulations. I left these as a separate commit since they're not strictly related to adding the `ClimaLandSimulation`.

experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -123,7 +104,6 @@ function BucketSimulation(
 
     # Initial conditions with no moisture
     Y, p, coords = CL.initialize(model)
-    p = get_new_cache(p, Y, energy_check)
```
imreddyTeja, line 126: Does this PR remove energy conservation checks? If this line is removed, `e_per_area` will never been the cache. If so, maybe the `get_field(bucket_sim::BucketSimulation, ::Val{:energy})` needs to be removed of changed as well.

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,531 @@
...
+    )
+
+Creates a ClimaLandSimulation object containing a land domain,
+a ClimaLand.LandModel, and an integrator.
+
+This type of model contains a canopy model, soil model, snow model, and
+soil CO2 model. Specific details about the complexity of the model
+can be found in the ClimaLand.jl documentation.
+"""
+function ClimaLandSimulation(
```
imreddyTeja, line 48: It might be more convenient for users if the argument ordering, names, and kwargs for the `BucketSimulation` and `ClimaLandSImulation` are unified, but that could probably be done in a later PR.

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,531 @@
...
+    parent(sim.integrator.p.drivers.P_snow) .= parent(field ./ ρ_liq)
+end
+function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:lw_d}, field)
+    parent(sim.integrator.p.drivers.LW_d) .= parent(field)
+end
+function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:sw_d}, field)
+    parent(sim.integrator.p.drivers.SW_d) .= parent(field)
+end
+function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:zenith_angle}, field)
+    parent(sim.integrator.p.drivers.θs) .= parent(field)
```
imreddyTeja, line 488: I think this needs to be 
```suggestion
    parent(sim.integrator.p.drivers.cosθs) .= parent(field)
```
to match the change in ClimaLand

experiments/ClimaEarth/components/land/climaland_bucket.jl
juliasloan25: Oh good point, I don't know why I thought it was okay to remove this function. I'll add it back

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: Good point, I updated this PR to make them the same. The bucket still has some extra arguments, but everything that's shared between the two models is unified now

experiments/ClimaEarth/components/land/climaland_helpers.jl
```diff
@@ -0,0 +1,59 @@
...
+    make_land_domain(
+        atmos_boundary_space::CC.Spaces.SpectralElementSpace2D,
+        zlim::Tuple{FT, FT},
+        nelements_vert::Int,) where {FT}
+
+Creates the land model domain from the horizontal space of the atmosphere, and information
+about the number of elements and extent of the vertical domain.
+"""
+function make_land_domain(
+    atmos_boundary_space::CC.Spaces.SpectralElementSpace2D,
```
kmdeck, line 31: A note that when ClimaLand uses a mask in the space, the atmos still will not. On our side, we will need to both set enable_mask = true and update the atmos space's mask in place with the correct values. The latter we can do with an existing space, but I don't think we can enable the mask once we've made a space without it.

so this function will get more complicated

src/Interfacer.jl
```diff
@@ -247,13 +248,17 @@ update_field!(
     val::Union{
         Val{:air_density},
         Val{:area_fraction},
+        Val{:air_temperature},
```
kmdeck, line 251: Do other surface models need these additional fields? Or just ClimaLand? If so, ClimaLand can make a new method?

experiments/ClimaEarth/components/land/climaland_helpers.jl
kmdeck: Another note to make is that ClimaLand (when running in standalone) will always use GL quadrature and npolynomial = 0. this will be different than Atmos. We may be making some assumptions in ClimaCore mask capabilities and interpolation that this is the case. 

That said, to start with, we can run coupled simulations without ClimaLand using a mask and using whatever space the atmos is using. ClimaLand will be much slower in that case, but it should be okay

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+import ClimaTimeSteppers as CTS
+import ClimaDiagnostics as CD
+import ClimaUtilities.TimeManager: ITime
+
+include("climaland_helpers.jl")
+
+"""
+    ClimaLandSimulation{M, P, D, I, A}
+
+The integrated ClimaLand model simulation object.
```
kmdeck, line 17: Perhaps we could indicate what the types of these fields are in the doc string

kmdeck: We may also wish to place restrictions? e.g. M <: ClimaLand.LandModel

kmdeck: and we could enforce D <: ClimaLand.Domains.SphericalShell instead of the @assert below?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    start_date::Dates.DateTime,
+    output_dir::String,
+    boundary_space,
+    area_fraction;
+    saveat::Vector{TT} = [tspan[1], tspan[2]],
+    domain_type::String = "sphere",
+    surface_elevation = CC.Fields.zeros(boundary_space),
+    land_temperature_anomaly::String = "amip",
+    use_land_diagnostics::Bool = true,
+    stepper = CTS.ARS111(),
```
kmdeck, line 62: is this the first time we use an implicit stepper for a surface model? just curious.

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+) where {FT, TT <: Union{Float64, ITime}}
+    @assert domain_type == "sphere" "Currently only spherical shell domains are supported; single column may be supported in the future."
+
+    # Set up domain
```
kmdeck, line 66: Yes, this is where I am wondering if we eventually will replace this with https://github.com/CliMA/ClimaLand.jl/blob/3685f247d2027f9da73256879bf5867165fc9c52/src/simulations/domains.jl#L35

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    depth = FT(10) # soil depth
+    n_vertical_elements = 10
+    domain = make_land_domain(boundary_space, (-depth, FT(0.0)), n_vertical_elements)
+    surface_space = domain.space.surface
+    subsurface_space = domain.space.subsurface
+
+    # Set up spatially-varying parameters
+    earth_param_set = CL.Parameters.LandParameters(FT)
+    spatially_varying_soil_params = CL.default_spatially_varying_soil_parameters(subsurface_space, surface_space, FT)
```
kmdeck, line 75: I think the land model should be providing an easier way to set this up! otherwise the standard land model and what is used in the coupler will rapidly be out of sync

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    ρc_s = CL.Soil.volumetric_heat_capacity.(Y.soil.ϑ_l, Y.soil.θ_i, soil_args.parameters.ρc_ds, earth_param_set)
+    Y.soil.ρe_int .= CL.Soil.volumetric_internal_energy.(Y.soil.θ_i, ρc_s, orog_adjusted_T, earth_param_set)
+    Y.soilco2.C .= FT(0.000412)
+
+    plant_ν = FT(1.44e-4)
+    Y.canopy.hydraulics.ϑ_l.:1 .= plant_ν
+    @. Y.canopy.energy.T = orog_adjusted_T_surface
+    Y.snow.S .= 0.0
+    Y.snow.S_l .= 0.0
+    Y.snow.U .= 0.0
```
kmdeck, line 153: I think we should switch to these:
https://github.com/CliMA/ClimaLand.jl/blob/3685f247d2027f9da73256879bf5867165fc9c52/experiments/long_runs/snowy_land.jl#L343-L356

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    SAI = FT(0.0) # m2/m2
+    f_root_to_shoot = FT(3.5)
+    RAI = FT(1.0)
+    K_sat_plant = FT(5e-9) # m/s # seems much too small?
+    ψ63 = FT(-4 / 0.0098) # / MPa to m, Holtzman's original parameter value is -4 MPa
+    Weibull_param = FT(4) # unitless, Holtzman's original c param value
+    a = FT(0.05 * 0.0098) # Holtzman's original parameter for the bulk modulus of elasticity
+    conductivity_model = CL.Canopy.PlantHydraulics.Weibull{FT}(K_sat_plant, ψ63, Weibull_param)
+    retention_model = CL.Canopy.PlantHydraulics.LinearRetentionCurve{FT}(a)
+    plant_ν = FT(1.44e-4)
```
kmdeck, line 240: This is currently defined twice, once here, and once on line 148

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    integrator = SciMLBase.init(
+        prob,
+        ode_algo;
+        dt = dt,
+        saveat = saveat,
+        adaptive = false,
+        callback = SciMLBase.CallbackSet(diag_cb),
+    )
+
+    sim = ClimaLandSimulation(model, (; domain = domain, soil_depth = depth), integrator, area_fraction)
```
kmdeck, line 203: Im curious why the `domain` is this `(; domain = domain, soil_depth = depth)`?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+
+    # DSS state to ensure we have continuous fields
+    dss_state!(sim)
```
kmdeck, line 206: the land model does not use dss in standalone runs - are we sure we need this?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:beta}) =
+    CL.surface_evaporative_scaling(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:emissivity}) = sim.integrator.p.ϵ_sfc
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:energy}) = CL.total_energy(sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_direct_albedo}) =
+    CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_diffuse_albedo}) =
+    CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:water}) = CL.total_water(sim.integrator.u, sim.integrator.p)
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:height_disp})
```
kmdeck, line 392: I am not sure we need this function anymore

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    d_canopy = CL.displacement_height(sim.model.canopy, sim.integrator.u, sim.integrator.p)
+    d_snow = CL.displacement_height(sim.model.snow, sim.integrator.u, sim.integrator.p)
+    d_soil = CL.displacement_height(sim.model.soil, sim.integrator.u, sim.integrator.p)
+    return NamedTuple{(:canopy, :snow, :soil)}((d_canopy, d_snow, d_soil))
+end
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:roughness_buoyancy})
```
kmdeck, line 398: I am not sure we need this function anymore

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    z_0b_canopy = sim.model.canopy.parameters.z_0b
+    z_0b_snow = sim.model.snow.parameters.z_0b
+    z_0b_soil = sim.model.soil.parameters.z_0b
+    return NamedTuple{(:canopy, :snow, :soil)}((z_0b_canopy, z_0b_snow, z_0b_soil))
+end
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:roughness_momentum})
```
kmdeck, line 404: Likewise

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    z_0m_canopy = sim.model.canopy.parameters.z_0m
+    z_0m_snow = sim.model.snow.parameters.z_0m
+    z_0m_soil = sim.model.soil.parameters.z_0m
+    return NamedTuple{(:canopy, :snow, :soil)}((z_0m_canopy, z_0m_snow, z_0m_soil))
+end
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_temperature})
```
kmdeck, line 410: Do we only need T_eff here? (same reason as above)

experiments/ClimaEarth/components/land/climaland_integrated.jl
kmdeck: Because the land knows how to access these things from itself (for surface fluxes), and we are computing surface fluxes with a land function

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+end
+
+"""
+    update_field!(sim::ClimaLandSimulation, ::Val{:turbulent_energy_flux}, fields::NamedTuple)
+
+Takes in a NamedTuple having fields `canopy`, `snow`, and `soil`, which in turn
+each contain fields `lhf` and `shf`, and updates each component of the land
+simulation with the corresponding LHF and SHF fields.
+"""
+function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:turbulent_energy_flux}, fields::NamedTuple)
```
kmdeck, line 434: When is this function called?

kmdeck (commented):
I have a couple questions about certain fields used for surface fluxes. Other than that, I think we should expect in the long run that the land model will not have the same space as the atmos model (it will use a mask, GL quadrature, etc, in standalone runs). I also think that we really need to provide a helper function from the land side for setting up global simulations.

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: Yes good points! I'll change these here and for the bucket. We can remove the `domain_type` argument altogether with this change too. It's unused right now and if we want to support running on other domains in the future it should be specified for the overall system rather than just for land via `land_domain_type` like we have right now.

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: Agreed. We can work on it on the land side - I opened an issue here https://github.com/CliMA/ClimaLand.jl/issues/1067

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: Okay! One downside is that this restricts us to run with 50m soil depth, but I think that's fine

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: A couple years ago we had conservation issues because some fields were discontinuous. We maintain continuous fields throughout the simulation, but we also need to ensure that the initial conditions are continuous. To make sure this is the case, we dss everything at the start of the simulation. It's possible that we don't need this for the land model with these ICs, but we do this across all component models so we don't miss anything.

This issue has more details from when we initially found this https://github.com/CliMA/ClimaCoupler.jl/issues/435

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+        hydrology_cm,
+        K_sat,
+        S_s,
+        θ_r,
+        PAR_albedo_dry = PAR_albedo_dry,
+        NIR_albedo_dry = NIR_albedo_dry,
+        PAR_albedo_wet = PAR_albedo_wet,
+        NIR_albedo_wet = NIR_albedo_wet,
+    )
+    return (domain = domain, parameters = soil_params)
```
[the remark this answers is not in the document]
juliasloan25: It seems like this is what's expected by the `canopy_model_args` that we pass into `LandModel`, e.g. see here in the snowy land longrun https://github.com/CliMA/ClimaLand.jl/blob/c0fb86948c460d76870b576f377eeff74ee4bfec/experiments/long_runs/snowy_land.jl#L309-L312

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: I agree, but if in the future we want to compute fluxes by calling SurfaceFluxes directly from the coupler, maybe it would be good to know where to access these things in the land model. We can either keep it, or leave it out for now and add it in if/when we need it

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: This is used by the combined state fluxes option, called here https://github.com/CliMA/ClimaCoupler.jl/blob/710a9f5c703ae40d7caea88b10cbc9339a630b3f/src/FieldExchanger.jl#L176-L181

When using partitioned state fluxes (which is what we're focusing on for the full land model), the component model's fluxes get updated directly

juliasloan25: I guess since we aren't supporting combined surface fluxes for the full land model, we can remove these two functions

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: When I read in the IC from file, the model isn't stable when taking a timestep. When I change back to 0, even with 50m depth, it is stable. This is the error I get
```
/Users/jsloan/clima/ClimaCoupler.jl/experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl:12
  Got exception outside of a @test
  DomainError with -0.10677256:
  Exponentiation yielding a complex result requires a complex argument.
  Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
  Stacktrace:
    [1] throw_exp_domainerror(x::Float32)
      @ Base.Math ./math.jl:41
    [2] ^(x::Float32, y::Float32)
      @ Base.Math ./math.jl:1186
    [3] co2_diffusivity(T_soil::Float32, θ_w::Float32, P_sfc::Float32, θ_a100::Float32, b::Float32, ν::Float32, params::ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters{Float32, ClimaLand.Parameters.LandParameters{Float32, Thermodynamics.Parameters.ThermodynamicsParameters{Float32}, SurfaceFluxes.Parameters.SurfaceFluxesParameters{Float32, SurfaceFluxes.UniversalFunctions.BusingerParams{Float32}, Thermodynamics.Parameters.ThermodynamicsParameters{Float32}}, Insolation.Parameters.InsolationParameters{Float32}}})
      @ ClimaLand.Soil.Biogeochemistry ~/.julia/packages/ClimaLand/Fc8cR/src/standalone/Soil/Biogeochemistry/co2_parameterizations.jl:86
    [4] _broadcast_getindex_evalf
      @ ./broadcast.jl:678 [inlined]
    [5] _broadcast_getindex
      @ ./broadcast.jl:651 [inlined]
    [6] getindex
      @ ./broadcast.jl:610 [inlined]
    [7] copyto!(dest::ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, bc::Base.Broadcast.Broadcasted{ClimaCore.DataLayouts.VIJFHStyle{10, 4, Array{Float32}}, NTuple{5, Base.OneTo{Int64}}, typeof(ClimaLand.Soil.Biogeochemistry.co2_diffusivity), Tuple{ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, ClimaCore.DataLayouts.IJFH{Float32, 4, Array{Float32, 4}}, ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, ClimaCore.DataLayouts.VIJFH{Float32, 10, 4, Array{Float32, 5}}, Tuple{ClimaLand.Soil.Biogeochemistry.SoilCO2ModelParameters{Float32, ClimaLand.Parameters.LandParameters{Float32, Thermodynamics.Parameters.ThermodynamicsParameters{Float32}, SurfaceFluxes.Parameters.SurfaceFluxesParameters{Float32, SurfaceFluxes.UniversalFunctions.BusingerParams{Float32}, Thermodynamics.Parameters.ThermodynamicsParameters{Float32}}, Insolation.Parameters.InsolationParameters{Float32}}}}}}, ::ClimaCore.DataLayouts.ToCPU)
      @ ClimaCore.DataLayouts ~/.julia/packages/ClimaCore/K4gqi/src/DataLayouts/copyto.jl:184
    [8] copyto!
      @ ~/.julia/packages/ClimaCore/K4gqi/src/DataLayouts/copyto.jl:16 [inlined]
    [9] copyto!
      @ ~/.julia/packages/ClimaCore/K4gqi/src/Fields/broadcast.jl:184 [inlined]
   [10] materialize!
      @ ./broadcast.jl:883 [inlined]
   [11] materialize!
      @ ./broadcast.jl:880 [inlined]
   [12] update_aux!
      @ ~/.julia/packages/ClimaLand/Fc8cR/src/standalone/Soil/Biogeochemistry/Biogeochemistry.jl:400 [inlined]
```

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+"""
+Extend Interfacer.add_coupler_fields! to add the fields required for ClimaLandSimulation.
+
+The fields added are:
+- `:SW_d` (for radiative transfer)
+- `:LW_d` (for radiative transfer)
+- `:cos_zenith_angle` (for radiative transfer)
+- `:diffuse_fraction` (for radiative transfer)
+- `:c_co2` (for photosynthesis, biogeochemistry)
+- `:d_sfc` (for turbulent fluxes)
```
kmdeck, line 512: This one is not needed

experiments/ClimaEarth/components/land/climaland_integrated.jl
kmdeck: To do on land side:
- debug NaNs in soil, canopy, snow (stability issues)
- use Nanmath and debug soil co2 so that instability manifests as NaN, not domain error, so that we can run on CPU

For now for coupler: use IC that allow a single step without error, run on CPU and/or GPU

experiments/ClimaEarth/components/land/climaland_integrated.jl
kmdeck: Thank you!

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    orog_adjusted_T = CC.Fields.Field(orog_adjusted_T_data, subsurface_space)
+    orog_adjusted_T_surface = CC.Fields.Field(CC.Fields.level(orog_adjusted_T_data, 1), surface_space)
+
+    @. Y.soil.ϑ_l = θ_r + (ν - θ_r) / 2
+    Y.soil.θ_i .= FT(0.0)
+    ρc_s = CL.Soil.volumetric_heat_capacity.(Y.soil.ϑ_l, Y.soil.θ_i, soil_args.parameters.ρc_ds, earth_param_set)
+    Y.soil.ρe_int .= CL.Soil.volumetric_internal_energy.(Y.soil.θ_i, ρc_s, orog_adjusted_T, earth_param_set)
+    Y.soilco2.C .= FT(0.000412)
+
+    plant_ν = FT(1.44e-4)
```
kmdeck: do you want to get this from the model to avoid defining 2x?
can access with land.canopy.hydraulics.parameters.plant_nu or \nu

experiments/ClimaEarth/components/land/climaland_integrated.jl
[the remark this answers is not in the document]
kmdeck: This is fine, then, but it might be clearer to get the params, and then make the args in the main file. But, since this will hopefully be replaced soon with the easier setup from land, this is OK for now!

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,540 @@
...
+    dss_state!(sim::ClimaLandSimulation))
+
+Perform DSS on the state of a component simulation, intended to be used
+before the initial step of a run. This method acts on ClimaLand simulations.
+The `dss!` function of ClimaLand must be called because it uses either the 2D
+or 3D dss buffer stored in the cache depending on space of each variable in
+`sim.integrator.u`.
+"""
+function dss_state!(sim::ClimaLandSimulation)
+    CL.dss!(sim.integrator.u, sim.integrator.p, sim.integrator.t)
```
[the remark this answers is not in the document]
kmdeck: Just calling this out again, since the understanding was that this was needed for conservation. our dss function does not do anything right now (for the bucket either).

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -0,0 +1,511 @@
...
+    CL.surface_evaporative_scaling(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:emissivity}) = sim.integrator.p.ϵ_sfc
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:energy}) = CL.total_energy(sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_direct_albedo}) =
+    CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_diffuse_albedo}) =
+    CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
+Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:water}) = CL.total_water(sim.integrator.u, sim.integrator.p)
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:height_disp})
+    d_canopy = CL.displacement_height(sim.model.canopy, sim.integrator.u, sim.integrator.p)
```
kmdeck, line 407: I wonder if we could leave these in another section marked off with ###### to note that that in the current implementation, these are not used?

kmdeck: seeing them may make the coupling harder for someone to understand, just a thought

kmdeck (approved):
Looks good to me! This a big step towards coupling :)

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: sorry I forgot to remove these but just did it! let me push now

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: yes! I'm getting it from `canopy_component_args.hydraulics.parameters.ν` now

PR #1205, opened 2025-02-24
Tr/add get fields atmos

imreddyTeja:
## Content

----


src/Utilities.jl
```diff
@@ -36,6 +36,13 @@ or 0 otherwise, keeping the same type.
 """
 binary_mask(var) = binary_mask(var, eps(eltype(var)))
 
+"""
+    safe_div(x, y)
+
+Divides `x` by `y` if `y` is not zero, otherwise returns zero.
+"""
+safe_div(x, y) = y == zero(y) ? zero(x) : x / y
```
juliasloan25, line 44: Instead of returning 0 in this case, it's probably better to divide by a small number close to 0 (i.e. `eps(FT)`). This is done sometimes in land, like [here](https://github.com/CliMA/ClimaLand.jl/blob/c8fb12ce3c7dc36e4d25edff61ac3a83c5a0be3a/src/standalone/Snow/snow_parameterizations.jl#L190-L194)

imreddyTeja: I moved this into an anonymous function within `get_field(::ClimaAtmosSimulation, ::Val{:diffuse_fraction})` because this is probably a one off use. When calculating diffuse_fraction, I think that it makes more sense to return 0 when the divisor is 0.  The division that occurs there is (total_flux_dn - direct_flux_dn) / total_flux_dn, so when the divisor is zero, the dividend should also be zero. 

Is it better to just divide by `eps(FT)`? This should result in the same values, but the behavior will be different in the case where `total_flux_dn == 0 && direct_flux_dn != 0`, which should not happen.

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -171,12 +171,36 @@ moisture_flux(::Union{CA.EquilMoistModel, CA.NonEquilMoistModel}, integrator) =
...
 # extensions required by the Interfacer
+function Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:diffuse_fraction})
+    radiation_model = sim.integrator.p.radiation.rrtmgp_model
+    total_flux_dn = radiation_model.face_sw_flux_dn
+    face_space = axes(sim.integrator.u.f)
+    if radiation_model.radiation_mode isa CA.RRTMGPInterface.GrayRadiation
+        diffuse_fractions = zero(total_flux_dn)
+    else
+        direct_flux_dn = radiation_model.face_sw_direct_flux_dn
+        diffuse_fractions = ((x, y) -> y > zero(y) ? x / y : zero(x)).(total_flux_dn .- direct_flux_dn, total_flux_dn)
```
szy21, line 182: It might be better to only calculate the first half level as that's all you need?

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -171,12 +171,36 @@ moisture_flux(::Union{CA.EquilMoistModel, CA.NonEquilMoistModel}, integrator) =
...
+    end
+    return CC.Fields.level(CC.Fields.array2field(diffuse_fractions, face_space), CC.Utilities.half)
+end
+Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{Symbol("cos(zenith_angle)")}) = CC.Fields.array2field(
```
szy21, line 186: not sure about the standard for coupler but maybe better not to have brackets in the Symbol?

src/Interfacer.jl
```diff
@@ -119,13 +119,19 @@ an atmosphere component model.
 get_field(
     sim::AtmosModelSimulation,
     val::Union{
+        Val{Symbol("cos(zenith_angle)")},
+        Val{:diffuse_fraction},
         Val{:energy},
         Val{:height_int},
         Val{:height_sfc},
         Val{:liquid_precipitation},
+        Val{:LHF},
```
szy21, line 128: do we need this (and SHF)?

src/Utilities.jl
szy21: If total_flux_dn = 0, direct_flux_dn should be 0 too. And diffuse_fraction should be non-negative. If this is not the case we should fix it in RRTMGP.

src/Interfacer.jl
imreddyTeja: we do not. They are removed now

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -171,12 +171,38 @@ moisture_flux(::Union{CA.EquilMoistModel, CA.NonEquilMoistModel}, integrator) =
...
+    radiation_model = sim.integrator.p.radiation.rrtmgp_model
+    # only take the first level
+    total_flux_dn = radiation_model.face_sw_flux_dn[1, :]
+    face_space = CC.Spaces.level(axes(sim.integrator.u.f), CC.Utilities.half)
+    if radiation_model.radiation_mode isa CA.RRTMGPInterface.GrayRadiation
+        diffuse_fractions = zero(total_flux_dn)
+    else
+        direct_flux_dn = radiation_model.face_sw_direct_flux_dn[1, :]
+        FT = eltype(total_flux_dn)
+        diffuse_fractions = (total_flux_dn .- direct_flux_dn) ./ (max.(total_flux_dn, eps(FT)))
```
szy21, line 188: I wonder if we want to limit it between 0 and 1 just to be exra safe. I don't think it should matter, maybe we can also wait and see if there is a problem.

szy21 (approved):
Thanks!

src/Utilities.jl
juliasloan25: Okay, it sounds like it's fine to set the return value to 0 if the denominator is 0 then

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -210,6 +236,11 @@ function Interfacer.update_field!(atmos_sim::ClimaAtmosSimulation, ::Val{:co2},
     end
 end
 # extensions required by the Interfacer
+function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:emissivity}, field)
+    # sets all 16 bands (rows) to the same values
+    sim.integrator.p.radiation.rrtmgp_model.surface_emissivity .=
+        reshape(CC.Fields.field2array(field), 1, length(parent(field)))
+end
```
szy21, line 243: In the long term, maybe we can make the RRTMGPInterface assume every band is the same if only one value is passed in (maybe this is already the behavior not? I'm not sure). This way we don't need to hard code the length(parent) here.

juliasloan25:
We recently changed CO2 from being read in within ClimaCoupler and passed to ClimaAtmos, to being read in within ClimaAtmos. We should have removed the `update_field!` method for co2 but we missed it. Could you remove that method for ClimaAtmos, as well as the mentions of `:co2` in src/Interfacer.jl and the tests, in this PR please?

When we use the full land model, we'll need to read in mean global co2 from the atmosphere and pass it to the land model, so we need a `get_field` for this quantity from atmos. Could you add that in this PR too please?

juliasloan25:
@szy21 @Sbozzolo Will the field `atmos_sim.integrator.p.radiation.rrtmgp_model.volume_mixing_ratio_co2` only be filled if we run atmos with radiation and the `MaunaLoa` option for CO2? That seems to be implied by the changes in the ClimaAtmos PR moving the file read to atmos https://github.com/CliMA/ClimaAtmos.jl/pull/3522/files, e.g. in `src/solver/type_getters.jl`. If so, we'll need to think about what CO2 value to supply to land when atmos doesn't provide CO2

Sbozzolo:
> @szy21 @Sbozzolo Will the field `atmos_sim.integrator.p.radiation.rrtmgp_model.volume_mixing_ratio_co2` only be filled if we run atmos with radiation and the `MaunaLoa` option for CO2? That seems to be implied by the changes in the ClimaAtmos PR moving the file read to atmos https://github.com/CliMA/ClimaAtmos.jl/pull/3522/files, e.g. in `src/solver/type_getters.jl`. If so, we'll need to think about what CO2 value to supply to land when atmos doesn't provide CO2

It is always filled. 

https://github.com/CliMA/ClimaAtmos.jl/blob/cedbf1b6e27ed084d04a66a559451841a44c31b8/src/parameterized_tendencies/radiation/radiation.jl#L93-L96

Here's how it's filled in the fixedco2 case

juliasloan25:
> It is always filled.
> 
> https://github.com/CliMA/ClimaAtmos.jl/blob/cedbf1b6e27ed084d04a66a559451841a44c31b8/src/parameterized_tendencies/radiation/radiation.jl#L93-L96
> 
> Here's how it's filled in the fixedco2 case

What if co2 is set to `nothing` and we aren't using radiation? I.e. [this case](https://github.com/CliMA/ClimaAtmos.jl/blob/cedbf1b6e27ed084d04a66a559451841a44c31b8/src/solver/type_getters.jl#L44-L45):
```
    (isnothing(co2) && !with_rrtgmp) &&
        @warn ("$(co2) does nothing if RRTGMP is not used")
```

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
imreddyTeja: diffuse_fraction is now clamped to [0,1]

docs/src/interfacer.md
```diff
@@ -123,6 +123,31 @@ following properties:
...
+| Coupler name      | Description | Units |
+|-------------------|-------------|-------|
+| `air_pressure`     | boundary air pressure | Pa |
+| `air_temperature`  | boundary air temperature | K |
+| `cos_zenith_angle` | cosine of the zenith angle | |
+| `co2`              | global mean co2 | ppm |
+| `diffuse_fraction` | fraction of downwards shortwave flux that is direct | |
+| `humidity`         | boundary humidity | kg kg^-1 |
+| `LW_d`             | downwards longwave flux | W m^-2 |
+| `SW_d`             | downwards shortwave flux | W m^-2 |
```
imreddyTeja, line 145: Are these descriptions accurate? Do the pressure and temperature values need `air_` prepended to them?

juliasloan25: These look good! Prepending `air_` is a good idea to distinguish from `surface_`. Instead of "boundary air pressure/temp/humidity", I would write "air pressure/temp/humidity at the bottom of the atmosphere", to avoid any confusion between the boundary layer of the atmosphere vs boundary surface

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -171,12 +171,54 @@ moisture_flux(::Union{CA.EquilMoistModel, CA.NonEquilMoistModel}, integrator) =
...
+    end
+    return CC.Fields.array2field(diffuse_fraction, lowest_face_space)
+end
 Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:liquid_precipitation}) =
     surface_rain_flux(sim.integrator.p.atmos.moisture_model, sim.integrator)
+Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:LW_d}) = CC.Fields.level(
+    CC.Fields.array2field(sim.integrator.p.radiation.rrtmgp_model.face_lw_flux_dn, axes(sim.integrator.u.f)),
+    CC.Utilities.half,
+)
+Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:humidity}) =
```
szy21, line 212: I would call it `specific_humidity` to be more accurate, but up to you.

docs/src/interfacer.md
```diff
@@ -123,6 +123,31 @@ following properties:
...
+Coupling with full `ClimaLand` model requires the following functions, in addition
+to the functions required for coupling with a general `SurfaceModelSimulation`.
+
+- `get_field(::AtmosModelSimulation. ::Val{property})`:
+This getter function must be extended
+for the following properties:
+
+| Coupler name      | Description | Units |
+|-------------------|-------------|-------|
+| `air_pressure`     | air pressure at the bottom of the atmosphere | Pa |
```
szy21, line 138:
```suggestion
| `air_pressure`     | air pressure at the bottom cell center of the atmosphere | Pa |
```
(and same for other bottom variables)

szy21 (approved):

src/Interfacer.jl
```diff
@@ -155,12 +155,20 @@ an atmosphere component model.
 get_field(
```
juliasloan25, line 155: I don't think we should error if these `get_field`s aren't implemented - e.g. if we want to couple a simpler/different atmosphere with the bucket, it doesn't need to extend all these methods required only for the full land model. I was going to suggest that we could give warnings if these aren't extended, but that will print a lot of warnings, so maybe the best thing to do is just remove these fields from here and expect users to read what's required to extend from the documentation

juliasloan25 (approved):
I just had one comment about throwing errors if the `get_field`s aren't extended, but this looks good other than that!

src/Interfacer.jl
imreddyTeja: I thought this would only error if the `get_field` is actually called for that value. So if we used some simple atmosphere with the bucket, it would not need to extend `get_field` for all the methods listed here, because `get_field(::SomeSimpleAtmos, ::Val{:cos_zenith})` would never be called. Is that correct? 

I think it might still make sense to split the `get_field(::AtmosModelSimulation)` default error into what is required for coupling with all land models, and just what is required for coupling with ClimaLand.

PR #1208, opened 2025-02-26
Fix ITime issue with saveat

ph-kev:
When `saveat = Inf`, it is not possible to make an `ITime`. When this happens, we will instead pass in an empty array for `saveat`.

### DIFF
--- a/experiments/ClimaEarth/user_io/arg_parsing.jl
+++ b/experiments/ClimaEarth/user_io/arg_parsing.jl
@@ -61,20 +61,24 @@ function get_coupler_args(config_dict::Dict)
     date0 = date = Dates.DateTime(config_dict["start_date"], Dates.dateformat"yyyymmdd")
     Δt_cpl = Float64(Utilities.time_to_seconds(config_dict["dt_cpl"]))
     saveat = Float64(Utilities.time_to_seconds(config_dict["dt_save_to_sol"]))
-    saveat = promote(t_start:saveat:t_end..., t_end)
     if use_itime
         t_end = ITime(t_end, epoch = date0)
         t_start = ITime(t_start, epoch = date0)
         Δt_cpl = ITime(Δt_cpl, epoch = date0)
         times = promote(t_end, t_start, Δt_cpl, ITime.(values(config_dict["component_dt_dict"]))...)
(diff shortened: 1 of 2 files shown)


experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -61,20 +61,24 @@ function get_coupler_args(config_dict::Dict)
...
-        saveat = ITime(Float64(Utilities.time_to_seconds(config_dict["dt_save_to_sol"])))
-        saveat = [promote([t_start:saveat:t_end..., t_end]...)...]
         component_dt_dict =
             Dict(component => first(promote(ITime(dt), t_end)) for (component, dt) in config_dict["component_dt_dict"])
     else
         component_dt_dict = config_dict["component_dt_dict"]
     end
+    if saveat != Inf
+        use_itime && (saveat = ITime(saveat))
+        saveat = [promote([t_start:saveat:t_end..., t_end]...)...]
```
juliasloan25, line 77: Can we just do `promote([t_start:saveat:t_end..., t_end]...)`, or do we need the additional splat and repackaging as a list?

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -61,20 +61,24 @@ function get_coupler_args(config_dict::Dict)
...
+    else
+        saveat = typeof(t_start)[]
```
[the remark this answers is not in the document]
ph-kev: Yes, because `typeof([]) == Vector{Any}` and we can't pass that when initializing one of the struct.

juliasloan25 (commented):
This looks good! Could you also add the following lines to `config/ci_configs/amip_coarse_ft32.yml` please? That way we'll exercise those options in our regular CI and we can catch issues like this before they get to the benchmarks.
```
dt_save_state_to_disk: "Inf"
dt_save_to_sol: "Inf"
```

experiments/ClimaEarth/user_io/arg_parsing.jl
ph-kev: I tried doing this without the additional splat and putting it into a vector, but it didn't work because some structs expect it to be a vector instead of a tuple.

juliasloan25 (approved):

PR #1210, opened 2025-02-27
add default get_field methods

juliasloan25:
## Purpose 
Add default methods for `get_field` methods that are commonly not extended for surface models. These return reasonable default values, and can be extended by surface models that won't use the defaults (e.g. the full land model).

I decided to add these now since adding the full land simulation adds some new `get_field` methods for surface model properties, e.g. emissivity, that have previously been assumed to have default values. This PR reduces code duplication by moving those common default values into default methods, instead of defining the same defaults for multiple surface models.

At first I added this as part of #1199, but I pulled it into a separate PR to make that one a little smaller.

### DIFF
--- a/docs/src/interfacer.md
+++ b/docs/src/interfacer.md
@@ -67,6 +67,18 @@ this function for both `AtmosModelSimulation` and
 the coupler. This function can optionally be extended to include
 additional field updates as desired.
 
+- `get_field(::ComponentModelSimulation, ::Val{property})`:
+Default `get_field` functions are provided for `energy` and `water` fields,
+described in the table below.
+These quantities are used to track conservation, and the defaults
+return `nothing`. To check conservation throughout a simulation, these
+functions must be extended for all models being run.
+
+| Coupler name      | Description | Units | Default value |
+|-------------------|-------------|-------|---------------|
+| `energy` | globally integrated energy | J | `nothing` |
+| `water` | globally integrated water | kg | `nothing` |
+
 ### AtmosModelSimulation - required functions
 In addition to the functions required for a general
 `ComponentModelSimulation`, an `AtmosModelSimulation` requires the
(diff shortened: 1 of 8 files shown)


src/Interfacer.jl
```diff
@@ -200,6 +197,13 @@ get_field(sim::ComponentModelSimulation, val::Val) = get_field_error(sim, val)
 
 get_field_error(sim, val::Val{X}) where {X} = error("undefined field `$X` for " * name(sim))
 
+# Set default values for fields that are not defined in all component models
+get_field(sim::ComponentModelSimulation, ::Val{:beta}) = convert(eltype(sim.integrator.u), 1.0)
+get_field(sim::ComponentModelSimulation, ::Val{:emissivity}) = convert(eltype(sim.integrator.u), 1.0)
+get_field(sim::ComponentModelSimulation, ::Val{:height_disp}) = convert(eltype(sim.integrator.u), 0.0)
+get_field(::ComponentModelSimulation, ::Val{:energy}) = nothing
+get_field(::ComponentModelSimulation, ::Val{:water}) = nothing
+
```
imreddyTeja, line 206: Should get_field for `beta`, `emissivity`, and `height_displacement` be defined for all `ComponentModelSimulations`? I thought those values only come from `SurfaceModelSimulations`.

docs/src/interfacer.md
```diff
@@ -157,6 +156,18 @@ following properties:
...
+For some quantities, default `get_field` functions are provided, which may be
+overwritten or used as-is. These currently include the following:
+
+| Coupler name      | Description | Units | Default value |
+|-------------------|-------------|-------|---------------|
+| `beta` | factor that scales evaporation based on its estimated level of saturation | | 1 |
+| `emissivity` | measure of how much energy a surface radiates | | 1 |
+| `height_disp` | displacement height relative to the surface | m | 0 |
+| `energy` | globally integrated energy | J | `nothing` |
+| `water` | globally integrated water | kg | `nothing` |
```
imreddyTeja, line 169: The default `get_field` functions for these value are for all component models, right? If so, I think we should also move `energy` and `water` from the `AtmosModelSimulation` required functions section into an optional functions section.

src/Interfacer.jl
juliasloan25: You're right, good catch!

docs/src/interfacer.md
juliasloan25: Good point! I'll make this change

imreddyTeja (approved):

PR #1217, opened 2025-03-05
Add optional `toml_files` kwarg to the `BucketSimulation` func

### LINKED ISSUE #1216: `BucketModelParameters` in ClimaEarth are hardcoded
For calibration, the `BucketSimulation` should take in a toml_dict or at least a NamedTuple to pass to the `BucketModelParameters` parameter set. This is already supported in [ClimaLand](https://github.com/CliMA/ClimaLand.jl/blob/main/ext/CreateParametersExt.jl#L442-L467) and should be a quick fix.

nefrathenrici:
This PR adds a `parameter_files` optional keyword argument to `BucketSimulation`. If `toml_files` is not empty, it will be used to construct the `BucketModelParameters`. If it is empty, behavior is unchanged.

`get_coupler_args` now returns `parameter_files` in its tuple, passing it to `BucketSimulation` in `setup_and_run`.

These changes will allow us to pass a calibration parameter file to the `BucketSimulation` in order to calibrate bucket parameters. Closes #1216

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -427,12 +427,9 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
     atmos_toml = joinpath.(pkgdir(CA), atmos_config["toml"])
     coupler_toml = joinpath.(pkgdir(ClimaCoupler), coupler_dict["coupler_toml"])
     toml = isempty(coupler_toml) ? atmos_toml : coupler_toml
-    if haskey(atmos_config, "calibration_toml")
-        push!(toml, atmos_config["calibration_toml"])
-    end
     if !isempty(toml)
         @info "Overwriting Atmos parameters from input TOML file(s): $toml"
-        atmos_config = merge(atmos_config, Dict("toml" => toml))
+        atmos_config["toml"] = toml
     end
 
     # Specify atmos output directory to be inside the coupler output directory
--- a/experiments/ClimaEarth/setup_run.jl
+++ b/experiments/ClimaEarth/setup_run.jl
@@ -200,6 +200,7 @@ function setup_and_run(config_dict::AbstractDict)
         energy_check,
         conservation_softfail,
         output_dir_root,
+        parameter_files,
     ) = get_coupler_args(config_dict)
 
     #=
@@ -341,6 +342,7 @@ function setup_and_run(config_dict::AbstractDict)
             albedo_type = land_albedo_type,
             land_initial_condition,
             energy_check,
+            parameter_files,
         )
 
         ## sea ice model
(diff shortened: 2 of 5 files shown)


experiments/ClimaEarth/setup_run.jl
```diff
@@ -326,6 +326,7 @@ function setup_and_run(config_dict::AbstractDict)
             energy_check = energy_check,
             surface_elevation,
             use_land_diagnostics,
+            toml_files,
```
juliasloan25, line 329: It seems a little weird to me that the `toml_files` we pass to the bucket come from `get_atmos_args`. Can we either specify individual tomls for atmos and for land, or if we want them to be the same for both can we specify the tomls at the coupler level and pass them to both atmos and land?

nefrathenrici: I agree, this is pretty hacky. Ideally we'd have one list of files that the coupler takes in and passes to both atmos and land but I wasn't sure how to accomplish that cleanly.

juliasloan25: We have the input `coupler_toml` that lets us specify a list of tomls at the coupler level. We could pass in multiple files there an pass them to atmos, land, etc. Maybe it could be a dictionary if we want to use the keys to determine which component the value goes to https://github.com/CliMA/ClimaCoupler.jl/blob/710a9f5c703ae40d7caea88b10cbc9339a630b3f/experiments/ClimaEarth/cli_options.jl#L23-L26

nefrathenrici: Thanks, I think  the `coupler_toml` will be perfect but we will need to change the behavior to not prepend the `pkgdir(ClimaCoupler)` path to the toml paths.

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -425,14 +425,11 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
 
     # use atmos toml if coupler toml is not defined
     atmos_toml = joinpath.(pkgdir(CA), atmos_config["toml"])
-    coupler_toml = joinpath.(pkgdir(ClimaCoupler), coupler_dict["coupler_toml"])
+    coupler_toml = coupler_dict["coupler_toml"]
```
juliasloan25, line 428: This imposes the requirement that the filepaths in `coupler_toml` need to be relative to where the simulation is being run from. Is that something that'll be useful to you in calibration? If you don't need it, I slightly prefer to keep the previous behavior where tomls are expected to be specified as a path within `ClimaCoupler.jl/`

juliasloan25 (approved):
Thanks Nat! I just had one question about relative vs. absolute paths. Could you also please update the NEWS.md before you merge to explain the change this PR makes?

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
nefrathenrici: This change turned out to be unnecessary, thanks for catching it. I'll revert this line and update the NEWS

PR #1219, opened 2025-03-06
update ocean, sea ice every 1 day

juliasloan25:
## Purpose 
Change the timestep for ocean and sea ice models to be 1 day by default (this will be used by all simulations except for that specified by `amip_component_dts.yml`. Also changes how component model timesteps are chosen from the input config to facilitate making this change.


juliasloan25:
replaced by #1926

PR #1220, opened 2025-03-11
add compute_surface_fluxes function for component models; test land fluxes

### LINKED ISSUE #1242: Add model-specific flux calculations
Right now, we perform the same flux calculations for all surface models. We need to abstract this out to dispatch on model type, so that we can perform different flux calculations for models that require this (i.e. the integrated land model).

juliasloan25:
## Purpose 
Closes #1242 

Extract most of the code out of the `partitioned_turbulent_fluxes!` function and move it into a function that dispatches on the component model type. For atmosphere models, the function does nothing, and for surface models it performs the flux calculation that was previously in `partitioned_turbulent_fluxes!`.

This design allows us to dispatch on model type instead of checking the type with an `if` statement, and, more importantly, it allows us to extend `compute_surface_fluxes!` for surface models where we want to perform different calculations (e.g. the integrated land model). 

## Content
- [x] Add `compute_surface_fluxes!` function, write methods for atmosphere and surface models
- [x] Write method for integrated land model - requires #1199 and https://github.com/CliMA/ClimaLand.jl/pull/1062
- [x] Make `import_atmos_fields` depend on surface model type, since exchange variables depend on surface model type
- [x] Add a test to compute surface fluxes between integrated land and ClimaAtmos, and check that fluxes stored in the coupler are changed and non-zero

### Dependency updates
This PR requires ClimaLand v0.15.13, so it updates the dependencies. The changes include:
```
    Updating `~/clima/ClimaCoupler.jl/experiments/ClimaEarth/Project.toml`
  [d414da3d] ↑ ClimaCore v0.14.28 ⇒ v0.14.29
  [08f4d4ce] ↑ ClimaLand v0.15.12 ⇒ v0.15.13
  [0bca4576] ↑ SciMLBase v2.79.0 ⇒ v2.81.0
```

## Summary by CodeRabbit

- **New Features**
  - Enhanced simulation integration with improved handling of atmospheric composition and moisture fluxes.
  - Introduced new functionality for updating simulation fields and computing surface fluxes.
  - Added a new configuration file to streamline test setups and parameter tuning.

- **Chores**
  - Updated runtime and dependency versions to boost performance and compatibility.

- **Tests**
  - Expanded test coverage with new scenarios verifying simulation flux calculations.
  - Introduced a new test set for "ClimaLandSimulation flux calculations".

- **Refactor**
  - Streamlined code for simulation field exchanges and flux computations, enhancing overall reliability.

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_integrated.jl
+++ b/experiments/ClimaEarth/components/land/climaland_integrated.jl
@@ -1,5 +1,6 @@
 import ClimaParams
 import ClimaLand as CL
+import ClimaLand.Parameters as LP
 import Dates
 import ClimaUtilities.TimeVaryingInputs: LinearInterpolation, PeriodicCalendar, TimeVaryingInput
 import ClimaCoupler: Checkpointer, FieldExchanger, FluxCalculator, Interfacer, Utilities
@@ -116,7 +117,7 @@ function ClimaLandSimulation(
     Csom = CL.PrescribedSoilOrganicCarbon{FT}(TimeVaryingInput((t) -> 5))
 
     land_input = (
-        atmos = CL.CoupledAtmosphere{FT}(),
+        atmos = CL.CoupledAtmosphere{FT}(boundary_space),
         radiation = CL.CoupledRadiativeFluxes{FT}(),
         runoff = runoff_model,
         soil_organic_carbon = Csom,
@@ -402,8 +403,7 @@ function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:area_fraction
     parent(sim.area_fraction) .= parent(field)
 end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:diffuse_fraction}, field)
-    p = sim.integrator.p
-    parent(p.canopy.radiative_transfer.diffuse_fraction) .= parent(field)
+    parent(sim.integrator.p.drivers.frac_diff) .= parent(field)
 end
 
 # Update fields stored in land drivers
@@ -413,18 +413,20 @@ end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:air_pressure}, field)
     parent(sim.integrator.p.drivers.P) .= parent(field)
 end
-function Interfacer.update_field!(::ClimaLandSimulation, ::Val{:air_humidity}, field)
+function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:air_humidity}, field)
     parent(sim.integrator.p.drivers.q) .= parent(field)
 end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:c_co2}, field)
     sim.integrator.p.drivers.c_co2 .= field
 end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:liquid_precipitation}, field)
+    # Arbitrarily take parameters from the soil (they are the same for all land sub-components)
     ρ_liq = (LP.ρ_cloud_liq(sim.model.soil.parameters.earth_param_set))
     parent(sim.integrator.p.drivers.P_liq) .= parent(field ./ ρ_liq)
 end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:snow_precipitation}, field)
-    ρ_liq = (LP.ρ_cloud_liq(sim.model.parameters.earth_param_set))
+    # Arbitrarily take parameters from the soil (they are the same for all land sub-components)
+    ρ_liq = (LP.ρ_cloud_liq(sim.model.soil.parameters.earth_param_set))
     parent(sim.integrator.p.drivers.P_snow) .= parent(field ./ ρ_liq)
 end
 function Interfacer.update_field!(sim::ClimaLandSimulation, ::Val{:lw_d}, field)
@@ -440,6 +442,39 @@ end
 Interfacer.step!(sim::ClimaLandSimulation, t) = Interfacer.step!(sim.integrator, t - sim.integrator.t, true)
 Interfacer.reinit!(sim::ClimaLandSimulation, t) = Interfacer.reinit!(sim.integrator, t)
 
+function FieldExchanger.update_sim!(sim::ClimaLandSimulation, csf, turbulent_fluxes, area_fraction)
+    # update fields for radiative transfer
+    Interfacer.update_field!(sim, Val(:diffuse_fraction), csf.diffuse_fraction)
+    Interfacer.update_field!(sim, Val(:sw_d), csf.SW_d)
+    Interfacer.update_field!(sim, Val(:lw_d), csf.LW_d)
+    Interfacer.update_field!(sim, Val(:cos_zenith_angle), csf.cos_zenith_angle)
+
+    # update fields for canopy conductance and photosynthesis
+    Interfacer.update_field!(sim, Val(:c_co2), csf.c_co2)
+    Interfacer.update_field!(sim, Val(:air_temperature), csf.T_air)
+    Interfacer.update_field!(sim, Val(:air_pressure), csf.P_air)
+    Interfacer.update_field!(sim, Val(:air_humidity), csf.q_air)
+
+    # precipitation
+    Interfacer.update_field!(sim, Val(:liquid_precipitation), csf.P_liq)
+    Interfacer.update_field!(sim, Val(:snow_precipitation), csf.P_snow)
+end
+
+function FieldExchanger.import_atmos_fields!(csf, sim::ClimaLandSimulation, atmos_sim, turbulent_fluxes)
+    # TODO we should be able to do this in a loop - need to unify coupler field names with get/update fields
+    FieldExchanger.dummmy_remap!(csf.diffuse_fraction, Interfacer.get_field(atmos_sim, Val(:diffuse_fraction)))
+    FieldExchanger.dummmy_remap!(csf.SW_d, Interfacer.get_field(atmos_sim, Val(:SW_d)))
+    FieldExchanger.dummmy_remap!(csf.LW_d, Interfacer.get_field(atmos_sim, Val(:LW_d)))
+    FieldExchanger.dummmy_remap!(csf.cos_zenith_angle, Interfacer.get_field(atmos_sim, Val(:cos_zenith)))
+    FieldExchanger.dummmy_remap!(csf.P_air, Interfacer.get_field(atmos_sim, Val(:air_pressure)))
+    FieldExchanger.dummmy_remap!(csf.T_air, Interfacer.get_field(atmos_sim, Val(:air_temperature)))
+    FieldExchanger.dummmy_remap!(csf.q_air, Interfacer.get_field(atmos_sim, Val(:specific_humidity)))
+    FieldExchanger.dummmy_remap!(csf.P_liq, Interfacer.get_field(atmos_sim, Val(:liquid_precipitation)))
+    FieldExchanger.dummmy_remap!(csf.P_snow, Interfacer.get_field(atmos_sim, Val(:snow_precipitation)))
+    # CO2 is a scalar so it doesn't need remapping TODO store CO2 as a scalar in the coupler fields
+    csf.c_co2 .= Interfacer.get_field(atmos_sim, Val(:co2))
+end
+
 """
 Extend Interfacer.add_coupler_fields! to add the fields required for ClimaLandSimulation.
 
@@ -452,9 +487,12 @@ The fields added are:
 - `:P_air` (for canopy conductance)
 - `:T_air` (for canopy conductance)
 - `:q_air` (for canopy conductance)
+- `P_liq` (for moisture fluxes)
+- `P_snow` (for moisture fluxes)
 """
 function Interfacer.add_coupler_fields!(coupler_field_names, ::ClimaLandSimulation)
-    land_coupler_fields = [:SW_d, :LW_d, :cos_zenith_angle, :diffuse_fraction, :c_co2, :P_air, :T_air, :q_air]
+    land_coupler_fields =
+        [:SW_d, :LW_d, :cos_zenith_angle, :diffuse_fraction, :c_co2, :P_air, :T_air, :q_air, :P_liq, :P_snow]
     push!(coupler_field_names, land_coupler_fields...)
 end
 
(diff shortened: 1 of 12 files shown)


experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,130 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    parent(coupled_atmos.h) .= parent(z_int)
+
+    # set the same atmosphere state for all sub-components
+    @assert sim.model.soil.boundary_conditions.top.atmos ===
+            sim.model.canopy.boundary_conditions.atmos ===
+            sim.model.snow.boundary_conditions.atmos ===
+            coupled_atmos
+
+    # get area fraction (min = 0, max = 1)
+    area_fraction = Interfacer.get_field(sim, Val(:area_fraction))
```
kmdeck, line 554: Maybe a comment explaining that this is the land area fraction

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,130 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    Y, p, t = sim.integrator.u, sim.integrator.p, sim.integrator.t
+
+    # update `csf` in-place using this model's flux calculation
+    fluxes = coupler_land_turbulent_fluxes!(p, sim.model, Y, t, coupled_atmos)
+    (; F_turb_ρτxz, F_turb_ρτyz, F_shf, F_lhf, F_turb_moisture) = fluxes
+
+    # add the flux contributing from this surface to the coupler field
+    # note that the fluxes are area-weighted, so if a surface model is
+    #  not present at this point, the fluxes are zero
+    @. csf.F_turb_ρτxz += F_turb_ρτxz * area_fraction * area_mask
```
kmdeck, line 567: Im not sure I understand why we multiply by area_fraction and area_mask, and what the difference is between those

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,130 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    F_shf = @. canopy_dest.shf +
+       soil_dest.shf * (1 - p.snow.snow_cover_fraction) +
+       p.snow.snow_cover_fraction * snow_dest.shf
+    F_turb_moisture = @. canopy_dest.transpiration +
+       (soil_dest.vapor_flux_liq + soil_dest.vapor_flux_ice) * (1 - p.snow.snow_cover_fraction) +
+       p.snow.snow_cover_fraction * snow_dest.vapor_flux
+    F_turb_ρτxz = @. soil_dest.ρτxz * (1 - p.snow.snow_cover_fraction) + p.snow.snow_cover_fraction * snow_dest.ρτxz
+    F_turb_ρτyz = @. soil_dest.ρτyz * (1 - p.snow.snow_cover_fraction) + p.snow.snow_cover_fraction * snow_dest.ρτyz
+
+    # At locations where this surface model is not evaluated, we get `NaN` for
```
kmdeck, line 614: Is it correct to say that the model is not evaluated, and are these really NaN? Or are they computed but irrelevant/wrong because over ocean? 

Also, for CL right now, I think we may have some NaNs over the ocean, but also other "physical" but meaningless values. This would just replace the NaN values, leaving some ocean values set, right?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,130 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+atmospheric properties required for the flux calculation: `u`, `h`,
+`thermal_state`, and `gustiness`.
+
+This function updates each subcomponent in place with its own turbulent fluxes, and also
+returns the total turbulent fluxes for the land model.
+
+Note that this function is currently allocating. We should rewrite it/allocate
+space so that it can update fields in-place instead.
+"""
+function coupler_land_turbulent_fluxes!(p, model::ClimaLand.LandModel, Y, t, coupled_atmos::ClimaLand.CoupledAtmosphere)
```
kmdeck, line 590: This function is also allocating the combined flux fields. Does it need to return them as a named tuple? Or can it update the named tuple fields in place? 

It also feels like this should be a new method of `ClimaLand.coupler_compute_turbulent_fluxes!` (esp if it is updating something in place)

src/FluxCalculator.jl
```diff
@@ -439,4 +382,95 @@ function water_albedo_from_atmosphere!(atmos_sim::Interfacer.AtmosModelSimulatio
...
+- `csf`: [CC.Fields.Field] containing a NamedTuple of turbulent flux fields: `F_turb_ρτxz`, `F_turb_ρτyz`, `F_turb_energy`, `F_turb_moisture`.
+- `sim`: [Interfacer.ComponentModelSimulation] the surface simulation to compute fluxes for.
+- `atmos_sim`: [Interfacer.AtmosModelSimulation] the atmosphere simulation to compute fluxes with.
+- `boundary_space`: [CC.Spaces.AbstractSpace] the space of the coupler surface.
+- `thermo_params`: [TD.Parameters.ThermodynamicsParameters] the thermodynamic parameters.
+- `surface_scheme`: [AbstractSurfaceFluxScheme] the surface flux scheme.
+"""
+function compute_surface_fluxes!(
+    csf,
+    sim::Interfacer.AtmosModelSimulation,
```
kmdeck, line 410: Are `sim` and `atmos_sim` really the same type? or should the sim be the surface simulation

kmdeck: oh maybe this just simplifies the loop 👍

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,130 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    fluxes = coupler_land_turbulent_fluxes!(p, sim.model, Y, t, coupled_atmos)
+    (; F_turb_ρτxz, F_turb_ρτyz, F_shf, F_lhf, F_turb_moisture) = fluxes
+
+    # add the flux contributing from this surface to the coupler field
+    # note that the fluxes are area-weighted, so if a surface model is
+    #  not present at this point, the fluxes are zero
+    @. csf.F_turb_ρτxz += F_turb_ρτxz * area_fraction * area_mask
+    @. csf.F_turb_ρτyz += F_turb_ρτyz * area_fraction * area_mask
+    @. csf.F_turb_energy += (F_shf .+ F_lhf) * area_fraction * area_mask
+    @. csf.F_turb_moisture += F_turb_moisture * area_fraction * area_mask
```
kmdeck, line 570: This is setting the coupler surface fields, but later on, are those used to update the land cache? Or are we using this `coupler_land_turbulent_fluxes!(p, sim.model, Y, t, coupled_atmos)` to update the land cache with the turb fluxes to be used for the land timestep?

experiments/ClimaEarth/components/land/climaland_integrated.jl
kmdeck: These lines also have the affect of hiding NaNs that appear over land (setting them to 0).

src/FluxCalculator.jl
juliasloan25: Yeah exactly

src/FluxCalculator.jl
```diff
@@ -439,4 +382,95 @@ function water_albedo_from_atmosphere!(atmos_sim::Interfacer.AtmosModelSimulatio
...
+    # update fluxes in the coupler
+    fields = (;
+        F_turb_ρτxz = F_turb_ρτxz,
+        F_turb_ρτyz = F_turb_ρτyz,
+        F_turb_energy = F_shf .+ F_lhf,
+        F_turb_moisture = F_turb_moisture,
+    )
+
+    # update the fluxes of this surface model
+    FluxCalculator.update_turbulent_fluxes!(sim, fields)
```
kmdeck, line 464: This is slightly confusing in terms of names - we "get_surface_fluxes!" (which computes them and allocates `fluxes`), then we "update_turbulent_fluxes!" (which I think is just updating the sim cache with the values), and then we update `csf` with them. 

Maybe a further PR could unify naming (turbulent fluxes vs surface fluxes), and clarify what is computing vs setting?

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
```diff
@@ -1,11 +1,17 @@
 using Test
 import ClimaCore as CC
+import ClimaAtmos as CA
 import ClimaCoupler
+import ClimaCoupler: FluxCalculator, Interfacer
 import Dates
+import ClimaComms
+@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
```
kmdeck, line 8: Would it make sense to move this to the compat and then remove from here?

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
```diff
@@ -46,3 +52,56 @@ FT = Float32
...
+    # Step the atmosphere once to get non-zero wind and humidity
+    Interfacer.step!(atmos_sim, dt)
+
+    # Exchange the initial conditions between atmosphere and land
+    # This also tests the `get_field`, `update_field!` and `update_model_sims!` methods for `ClimaLandSimulation`
+    FieldExchanger.import_combined_surface_fields!(coupler_fields, model_sims, flux_type)
+    FieldExchanger.import_atmos_fields!(coupler_fields, model_sims, boundary_space, flux_type)
+    FieldExchanger.update_model_sims!(model_sims, coupler_fields, flux_type)
+
+    # Step the land once to update cache variables with the updated drivers after the exchange
```
kmdeck, line 95: Im just wondering if, for clarity, we want to only update the drivers here? We can call ClimaLand.update_drivers! I believe for that

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
```diff
@@ -46,3 +52,56 @@ FT = Float32
...
+
+    # Compute the surface fluxes
+    FluxCalculator.compute_surface_fluxes!(coupler_fluxes, land_sim, atmos_sim, boundary_space, nothing, nothing)
+
+    # Check that the fluxes have been changed
+    zero_field = CC.Fields.zeros(boundary_space)
+    @test coupler_fluxes.F_turb_ρτxz != zero_field
+    @test coupler_fluxes.F_turb_ρτyz != zero_field
+    @test coupler_fluxes.F_turb_energy != zero_field
+    @test coupler_fluxes.F_turb_moisture != zero_field
```
kmdeck, line 106: Do we need to also confirm that they are NaN free? (compute_surface_fluxes! did a few other things)

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: ~~I agree this is confusing. `area_fraction` is a value in [0, 1] that tells what percentage of each grid cell is covered by land. `area_mask` is a binary mask containing values of 0 or 1, that is made by rounding the values in `area_fraction` based on a cutoff value (by default `eps(FT)`).~~

~~Basically, `area_mask` tells us whether we want to include the land contribution in this cell, and `area_fraction` tells us how much of this grid cell is covered in land. We need both because e.g. we could have non-zero `area_fraction` but still want to exclude land.~~

Using both is a bug, and we should just be using `area_fraction`. What we have now still works since `area_mask` is 0 if and only if `area_fraction` is 0, and it's otherwise 1. But this is confusing and could easily be incorrect, so we should remove `area_mask` altogether. I'll do it in a separate PR https://github.com/CliMA/ClimaCoupler.jl/issues/1255

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: We can update the named tuple fields in place. I wasn't sure of the best way to structure this since we need to get the land fluxes and multiply them by the `area_fraction`. I wanted this function to be independent of ClimaCoupler ideas (e.g. the coupler fields and `area_fraction`, but you're right that it's allocating. Let me try to rewrite this in a clean way without allocating

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: The land cache is updated in `coupler_land_turbulent_fluxes!`, so the coupler never updates the land's fluxes. This is different from every other component model which isn't very clean, but it would be redundant to update them again with the same values

src/FluxCalculator.jl
```diff
@@ -439,4 +382,95 @@ function water_albedo_from_atmosphere!(atmos_sim::Interfacer.AtmosModelSimulatio
...
+    # `_int` refers to atmos state of center level 1
+    z_int = Interfacer.get_field(atmos_sim, Val(:height_int))
+    uₕ_int = Interfacer.get_field(atmos_sim, Val(:uv_int))
+    thermo_state_int = Interfacer.get_field(atmos_sim, Val(:thermo_state_int))
+    z_sfc = Interfacer.get_field(atmos_sim, Val(:height_sfc))
+
+    # get area fraction (min = 0, max = 1)
+    area_fraction = Interfacer.get_field(sim, Val(:area_fraction))
+    # get area mask [0, 1], where area_mask = 1 if area_fraction > 0
+    area_mask = Utilities.binary_mask.(area_fraction)
```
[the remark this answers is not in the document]
juliasloan25: Yeah you're right. I'm planning to remove `area_mask`, and I'll open an issue to make `thermo_state_sfc` non-allocating https://github.com/CliMA/ClimaCoupler.jl/issues/1256

src/FluxCalculator.jl
juliasloan25: Good point, I opened an issue for this https://github.com/CliMA/ClimaCoupler.jl/issues/1257

src/FluxCalculator.jl
```diff
@@ -439,4 +382,95 @@ function water_albedo_from_atmosphere!(atmos_sim::Interfacer.AtmosModelSimulatio
...
+
+    # set inputs based on whether the surface_scheme is `MoninObukhovScheme` or `BulkScheme`
+    surface_params = FluxCalculator.get_surface_params(atmos_sim)
+    scheme_properties = FluxCalculator.get_scheme_properties(surface_scheme, sim)
+
+    input_args = (; thermo_state_sfc, thermo_state_int, uₕ_int, z_int, z_sfc, scheme_properties, boundary_space)
+    inputs = FluxCalculator.surface_inputs(surface_scheme, input_args)
+
+    # calculate the surface fluxes
+    fluxes = FluxCalculator.get_surface_fluxes!(inputs, surface_params)
```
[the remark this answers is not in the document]
juliasloan25: I agree this is weird but I'm hesitant to change this right now since Gabriele is fixing up the partitioned fluxes to work correctly (some fields aren't updated properly as it is now). For now I just moved this code from `partitioned_surface_fluxes!` into `compute_surface_fluxes!` so we can dispatch on model type and do something different for the integrated land.

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
juliasloan25: This line is like this so we're still compatible with previous versions of ClimaComms, which could be helpful if we ever need to debug/test with a previous version

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
juliasloan25: I like this idea, but when I change it to only update the drivers, the moisture flux remains 0. Do you know why that might be? I'm having a hard time finding other cache variables used for the flux calculations that might not be set.

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,108 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    # set the same atmosphere state for all sub-components
+    @assert sim.model.soil.boundary_conditions.top.atmos ===
+            sim.model.canopy.boundary_conditions.atmos ===
+            sim.model.snow.boundary_conditions.atmos ===
+            coupled_atmos
+
+    # get area fraction of the land model (min = 0, max = 1)
+    area_fraction = Interfacer.get_field(sim, Val(:area_fraction))
+    # get area mask [0, 1], where area_mask = 1 if area_fraction > 0
+    area_mask = Utilities.binary_mask.(area_fraction)
```
kmdeck, line 561: can remove entirely?

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
```diff
@@ -46,3 +52,65 @@ FT = Float32
...
+    # Step the atmosphere once to get non-zero wind and humidity
+    Interfacer.step!(atmos_sim, dt)
+
+    # Exchange the initial conditions between atmosphere and land
+    # This also tests the `get_field`, `update_field!` and `update_model_sims!` methods for `ClimaLandSimulation`
+    FieldExchanger.import_combined_surface_fields!(coupler_fields, model_sims, flux_type)
+    FieldExchanger.import_atmos_fields!(coupler_fields, model_sims, boundary_space, flux_type)
+    FieldExchanger.update_model_sims!(model_sims, coupler_fields, flux_type)
+
+    # Update land cache variables with the updated drivers in the cache after the exchange
```
kmdeck, line 97: I think this is now updating land cache variables (including those needed for SF computation) only

src/FluxCalculator.jl
```diff
@@ -439,4 +383,95 @@ function water_albedo_from_atmosphere!(atmos_sim::Interfacer.AtmosModelSimulatio
...
+        F_turb_moisture = F_turb_moisture,
+    )
+
+    # update the fluxes of this surface model
+    FluxCalculator.update_turbulent_fluxes!(sim, fields)
+
+    # add the flux contributing from this surface to the coupler field
+    # note that the fluxes are area-weighted, so if a surface model is
+    #  not present at this point, the fluxes are zero
+    @. csf.F_turb_ρτxz += F_turb_ρτxz * area_fraction * area_mask
```
kmdeck, line 470: do you also want to remove the area_mask here?

kmdeck (approved):
looks good with a couple of items for future work!

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -440,6 +442,38 @@ end
 Interfacer.step!(sim::ClimaLandSimulation, t) = Interfacer.step!(sim.integrator, t - sim.integrator.t, true)
 Interfacer.reinit!(sim::ClimaLandSimulation, t) = Interfacer.reinit!(sim.integrator, t)
 
+function FieldExchanger.update_sim!(sim::ClimaLandSimulation, csf, turbulent_fluxes, area_fraction)
+    # update fields for radiative transfer
+    Interfacer.update_field!(sim, Val(:diffuse_fraction), csf.diffuse_fraction)
+    Interfacer.update_field!(sim, Val(:sw_d), csf.SW_d)
+    Interfacer.update_field!(sim, Val(:lw_d), csf.LW_d)
+    Interfacer.update_field!(sim, Val(:cos_zenith_angle), csf.cos_zenith_angle)
```
Sbozzolo, line 450: To me, `cos_zenith_angle` is something that should not be updated/exchanged because it is global property of the system that depends only on the current data. ClimaLand can compute this internally.

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,108 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+for this model. These are then summed using area-weighting across all surface
+models to get the total fluxes.
+
+Because the integrated land model is composed of multiple sub-components, the
+fluxes are computed for each sub-component and then combined to get the total for this model.
+The land model cache is updated with the computed fluxes for each sub-component.
+
+Any NaNs present in the fluxes are replaced by 0. Note that we are likely to
+get NaNs in areas where this model is evaluated but not physically present
+(e.g. land model in ocean grid cells).
```
Sbozzolo, line 522: This is quite risky because it prevents possible problems from surfacing and being detected.

src/FieldExchanger.jl
```diff
@@ -67,8 +67,21 @@ have to be defined for the amtospheric component model type.
...
+    for sim in model_sims
+        import_atmos_fields!(csf, sim, model_sims.atmos_sim, turbulent_fluxes)
+    end
+
+end
 
+"""
+    import_atmos_fields!(csf, ::Interfacer.SurfaceModelSimulation, atmos_sim, turbulent_fluxes)
+
+Updates the coupler simulation fields with atmospheric fluxes from the atmosphere simulation.
```
Sbozzolo, line 79:
```suggestion
Update the coupler simulation fields with atmospheric fluxes from the atmosphere simulation.
```

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
juliasloan25: Yeah, that's what I tried to mean with this comment but maybe it isn't clear

src/FluxCalculator.jl
juliasloan25: I'll do this in a separate PR!

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: I see your point, but I don't think it makes sense to have the computation of `cos_zenith_angle` in two places. It's duplicated code, and they could have different ways of computing it and end up with different values. I'd like to keep it like this for now, and we can discuss computing it in ClimaLand for a future PR

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: Yeah I see your point. What we really want to do is filter out NaNs where the area fraction of a model is 0. I'll change it to do this tomorrow

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,108 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+    ClimaLand.coupler_compute_turbulent_fluxes!(snow_dest, coupled_atmos, model.snow, Y, p, t)
+
+    canopy_dest = p.canopy.turbulent_fluxes
+    ClimaLand.coupler_compute_turbulent_fluxes!(canopy_dest, coupled_atmos, model.canopy, Y, p, t)
+
+    # Combine turbulent energy fluxes from each component of the land model
+    # Use temporary variables to avoid allocating
+    @. csf.temp1 = @. canopy_dest.lhf +
+       soil_dest.lhf * (1 - p.snow.snow_cover_fraction) +
+       p.snow.snow_cover_fraction * snow_dest.lhf
```
charleskawczynski, line 579:
```suggestion
    @. csf.temp1 = canopy_dest.lhf +
       soil_dest.lhf * (1 - p.snow.snow_cover_fraction) +
       p.snow.snow_cover_fraction * snow_dest.lhf
```

experiments/ClimaEarth/components/land/climaland_integrated.jl
Sbozzolo: The zenith angle is a physical parameter and we already have a single shared package to ensure that they are computed in the same way (Insolation.jl). 

It will have to be computed internally anyway because we don't want to remap the zenith angle when there's an analytical exact formula to compute it exactly on each grid.

juliasloan25: @kmdeck could we add this to ClimaLand?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -463,3 +500,108 @@ function Checkpointer.get_model_prog_state(sim::ClimaLandSimulation)
...
+)
+    coupled_atmos = sim.model.soil.boundary_conditions.top.atmos
+
+    # `_int` refers to atmos state of center level 1
+    z_int = Interfacer.get_field(atmos_sim, Val(:height_int))
+    uₕ_int = Interfacer.get_field(atmos_sim, Val(:uv_int))
+    thermo_state_int = Interfacer.get_field(atmos_sim, Val(:thermo_state_int))
+
+    # We use parent here because of the mismatch between the lowest atmos level
+    #  and the boundary space, even though the horizontal grids are the same.
```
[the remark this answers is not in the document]
juliasloan25: Thank you!

experiments/ClimaEarth/test/component_model_tests/climaland_tests.jl
juliasloan25: Commenting here for completeness: The drivers are updated by the coupler in `update_model_sims!`. We need to call `update_aux!` to use those driver values to update other aux variables (e.g. vars for photosynthesis calculations)

PR #1225, opened 2025-03-14
Unify Interfacer argname for simulation

### LINKED ISSUE #1212: Unify `get_field` and `update_field` argument names in `climaatmos.jl`
The extended `Interfacer` functions in `experiments/CLimaEarth/components/atmosphere/climaatmos.jl` refer to the `::ClimaAtmosSimulation` argument as both `atmos_sim` and `sim`. It would be nice to unify this to either `atmos_sim` or just `sim`.

imreddyTeja:
The methods for each component in experiments/ClimaEarth/components/ refer to the simulation argument with different names. For the ClimaAtmosSimulation, it is refered to as both `sim` and `atmos_sim`. Similarly, the BucketSimulation is refered to as both `sim` and `bucket_sim`. This commit unifies them by changing everything to `sim`. The opposite could also be done.

## Purpose 
Closes #1212 

----

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -139,20 +139,19 @@ end
 
 
 """
-Interfacer.get_field(atmos_sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux_toa})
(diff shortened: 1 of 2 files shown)


juliasloan25 (approved):
Thanks Teja! The buildkite CUDA failure is from a CUDA problem with julia v1.11.0-2, but earlier today I merged a PR updating buildkite to use 1.11.3 so it should be fixed if you rebase

PR #1228, opened 2025-03-18
Prepend coupler pkgdir to all coupler TOML files

nefrathenrici:
## Purpose 

This PR ensures that the parameter filepaths passed to atmos and the bucket are unified.

## Content

----

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -447,8 +447,11 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
     end
 
     # use atmos toml if coupler toml is not defined
-    atmos_toml = joinpath.(pkgdir(CA), atmos_config["toml"])
-    coupler_toml = joinpath.(pkgdir(ClimaCoupler), coupler_dict["coupler_toml"])
+    # If we can't find the file at the relative path, prepend pkgdir(ClimaAtmos)
+    atmos_toml = map(atmos_config["toml"]) do file
+        isfile(file) ? file : joinpath(pkgdir(CA), file)
+    end
+    coupler_toml = atmos_config["coupler_toml"]
     toml = isempty(coupler_toml) ? atmos_toml : coupler_toml
     if !isempty(toml)
         @info "Overwriting Atmos parameters from input TOML file(s): $toml"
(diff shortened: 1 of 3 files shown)


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -449,7 +449,7 @@ function get_atmos_config_dict(coupler_dict::Dict, job_id::String, atmos_output_
 
     # use atmos toml if coupler toml is not defined
     atmos_toml = joinpath.(pkgdir(CA), atmos_config["toml"])
-    coupler_toml = joinpath.(pkgdir(ClimaCoupler), coupler_dict["coupler_toml"])
+    coupler_toml = atmos_config["coupler_toml"]
```
juliasloan25, line 452: It might be a little confusing to have these two lines together, where we have to prepend `pkgdir` for the atmos toml but we've already done it for the coupler. We could add a comment explaining why, but I don't think it matters too much either way

nefrathenrici: I'll add a comment, thanks!

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -59,6 +59,7 @@ function get_coupler_args(config_dict::Dict)
     FT = config_dict["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
 
     # Vector of TOML files containing model parameters
+    config_dict["coupler_toml"] = joinpath.(pkgdir(ClimaCoupler), config_dict["coupler_toml"])
```
Sbozzolo, line 62: This prevents people from using TOMLs outside of the install pkgdir, which could be in the `.julia` folder, a hidden folder that users should not be expected to touch.

nefrathenrici: This was already the behavior for the TOML files passed to ClimaAtmos, this PR just applies the same behavior further up to ensure the paths are the same for the bucket. 
https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/components/atmosphere/climaatmos.jl#L451

Sbozzolo: Could we fix this in atmos too? One way would be to introduce a hierarchy: First we try resolving `config_dict["coupler_toml"]` as a path, if such path doesn't exist, and we try prepending `pkgdir(ClimaCoupler)` and resolving that path.

PR #1232, opened 2025-03-21
update land energy/water conservation functions

juliasloan25:
## Purpose 
Update integrated land conservation functions to use the ClimaLand functions introduced in https://github.com/CliMA/ClimaLand.jl/pull/1071. Note that these provide the energy and water per area, so we integrate horizontally via `sum`


experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -389,13 +389,37 @@ Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:area_fraction}) = sim.area
...
 Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_direct_albedo}) =
     CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
 Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_diffuse_albedo}) =
     CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
-Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:water}) = CL.total_water(sim.integrator.u, sim.integrator.p)
 Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:surface_temperature}) = sim.integrator.p.T_sfc
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:energy}, level)
+    surface_field = similar(sim.integrator.p.soil.sfc_scratch)
+    return sum(
+        CL.total_energy_per_area!(
```
kmdeck, line 400: I think we need to call total_energy_per_area! and then call sum(surface_field) and return that, since CL.total_energy_per_area! returns nothing

since the `similar` function allocates, is there another surface field we can use?

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -389,13 +389,37 @@ Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:area_fraction}) = sim.area
...
+            sim.integrator.p,
+            sim.integrator.t,
+            sim.integrator.p.soil.sfc_scratch,
+        ),
+    )
+end
+function Interfacer.get_field(sim::ClimaLandSimulation, ::Val{:water}, level)
+    surface_field = similar(sim.integrator.p.soil.sfc_scratch)
+    return sum(
+        CL.total_liq_water_vol_per_area!(
```
kmdeck, line 413: likewise here

kmdeck: and one other thought is that we will need to multiply by density here or somewhere else to get mass

PR #1237, opened 2025-03-25
Fix: Close file after restoring model cache

### LINKED ISSUE #1236: Bug: JLD2File opened but never closed by Checkpointer.jl
We ran into an `IOError` (EBUSY) on Windows machines when working with the cache files created by the Checkpointer, see the issue described [here](https://github.com/valentinaschueller/clima-playground/issues/17). I am quite sure that this is due to the fact that [`Checkpointer.restart_model_cache!()`](https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/Checkpointer.jl#L163) opens but never closes the cache file. On my (Linux) machine this is not a problem, it might be that the resource is freed implicitly. But we consistently get this issue on Windows. 

I would suggest changing this function nonetheless, since it is explicitly recommended in the [JLD2 README](https://github.com/JuliaIO/JLD2.jl?tab=readme-ov-file#file-interface): "When you are done with the file, remember to call `close(f)`."

valentinaschueller:
## Purpose 

Closes #1236

## Content

Explicitly calls `close(file)` after the model cache is restored to avoid IOErrors.

----

### DIFF
--- a/src/Checkpointer.jl
+++ b/src/Checkpointer.jl
@@ -160,7 +160,9 @@ the component models that have a cache.
 function restart_model_cache!(sim, input_file)
     ispath(input_file) || error("File $(input_file) not found")
     # Component models are responsible for defining a method for this
-    restore_cache!(sim, JLD2.jldopen(input_file)["cache"])
+    JLD2.jldopen(input_file) do file
+        restore_cache!(sim, file["cache"])
+    end
 end
 
 """


src/Checkpointer.jl
```diff
@@ -160,7 +160,9 @@ the component models that have a cache.
 function restart_model_cache!(sim, input_file)
     ispath(input_file) || error("File $(input_file) not found")
     # Component models are responsible for defining a method for this
-    restore_cache!(sim, JLD2.jldopen(input_file)["cache"])
+    file = JLD2.jldopen(input_file)
+    restore_cache!(sim, file["cache"])
+    close(file)
```
charleskawczynski, line 165:
```suggestion
    JLD2.jldopen(input_file) do file
        restore_cache!(sim, file["cache"])
    end
```
I kind of prefer using the `do`-syntax, as this will gracefully handle errors, and it's a nice pattern to repeat (in this case, the file should be closed by `JLD2`)

valentinaschueller: Thanks, I was not aware of this! 👍

Sbozzolo (approved):
Thank you!

valentinaschueller:
Thanks for approving @Sbozzolo! Is there something else to do before this can be merged? I saw that the `codecov/patch` check failed, is that a problem I should address somehow? 🙂

Sbozzolo:
> Thanks for approving @Sbozzolo! Is there something else to do before this can be merged? I saw that the `codecov/patch` check failed, is that a problem I should address somehow? 🙂

We need a green light on buildkite, but I suspect buildkite is not starting on this PR because you are an external user and buildkite is disabled on external users to prevent abuse of resources (e.g. PRs mining cryptocurrencies on our severs).

I temporarily disabled this setting to let your PR through. Could you rebase, squash your commits, and push again?

valentinaschueller:
Done as you suggested! Weird, it seems that it's already the upload which fails.

Sbozzolo:
> Done as you suggested! Weird, it seems that it's already the upload which fails.

Okay, this is happening because you are on a fork, so your commit does not exist on this repo (yet). 

I guess we have to clean up our interaction between buildkite and external forks. 

For now, the easiest will be that I cherry pick your commit and manually merge it.

PR #1243, opened 2025-03-27
distinguish 1D and 2D coupler fields

juliasloan25:
## Purpose 
Many coupler exchange fields are 2D fields at the surface, but some are 1D scalars (CO2 concentration, beta, roughness lengths, etc) - at least for now. This PR updates the coupler exchange field initialization to specify whether a variable is 1D or 2D as it's added to the coupler fields, to avoid allocating and performing operations on unneeded fields. The intermediate `coupler_field_names` is now a 2-Tuple of Vectors rather than a single Vector, but `coupler_fields` remains a single NamedTuple

Edit: I just realized that even if we have a 1D field for one surface component model, it can (will) vary between surfaces. So for these we probably still want to have a 2D coupler field for exchange with the atmosphere. This really only leaves CO2 as a 1D field we'll exchange, so I'm not sure this PR is worth it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,8 +6,16 @@ ClimaCoupler.jl Release Notes
 
 ### ClimaCoupler features
 
+#### Identify coupler fields as 1D or 2D at init PR[#1243](https://github.com/CliMA/ClimaCoupler.jl/pull/1243)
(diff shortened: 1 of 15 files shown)


juliasloan25:
We almost always want to exchange 2D fields, so we won't be implementing this

PR #1251, opened 2025-04-01
Split `setup_and_run` in multiple steps

Sbozzolo:
This commit introduces more structure in ClimaEarth. It splits `setup_and_run` in three functions:
- `setup`, which produces a `CoupledSimulation` (hence, it was renamed to be `CoupledSimulation`, since it is just a constructor)
- `run!`, which runs the simulation. This is supported by the introduction of `step!`, to take one single coupling step. This is particularly useful in debugging
- `postprocess`, to make plots

The advantage of this separation is that one can now construct a simulation without running it, and step a simulation one step at a time. Both of these are particularly useful in interactive work and debugging.

This commit also renames `calendar_dt` to `diag_calendar_dt` to make it clearer that it refers to diagnostics.

## Summary by CodeRabbit

- **New Features**
  - Introduced a modular simulation lifecycle that separates initialization, execution, iterative stepping, and postprocessing.
  - Enhanced simulation output handling now provides clearer performance reporting and robust result validation.

- **Refactor**
  - Streamlined the simulation workflow for improved clarity and consistency in diagnostic scheduling and time management.
  - Maintained legacy functionality while transitioning to a more organized and flexible execution approach.
  - Updated variable naming for better clarity regarding diagnostic periods, including the renaming of `calendar_dt` to `diagnostics_dt`.

### DIFF
--- a/experiments/ClimaEarth/setup_run.jl
+++ b/experiments/ClimaEarth/setup_run.jl
@@ -35,6 +35,7 @@ import ClimaCoupler:
 import ClimaCoupler.Interfacer:
     AbstractSlabplanetSimulationMode,
     AMIPMode,
+    CoupledSimulation,
     SlabplanetAquaMode,
     SlabplanetEisenmanMode,
     SlabplanetMode,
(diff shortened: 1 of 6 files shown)


experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -29,11 +28,15 @@ function postprocess_sim(cs, postprocessing_vars)
     # Plot all model states and coupler fields (useful for debugging)
     !CA.is_distributed(cs.comms_ctx) && debug(cs, artifact_dir)
 
-    # If we have monthly data, plot the leaderboard
-    if t_end > 84600 * 31 * 3 && output_default_diagnostics
```
juliasloan25, line 33: `output_default_diagnostics` is now unused in ClimaCoupler. We might get an error from missing variables here if `output_default_diagnostics == false` and we try to plot the leaderboard, but I'm not entirely sure. It's a ClimaAtmos argument, so either way we can still include it in our config files, but if we don't use it here we can also remove it from `get_atmos_args`.

experiments/ClimaEarth/setup_run.jl
```diff
@@ -82,87 +83,18 @@ include("user_io/postprocessing.jl")
...
-    solve_coupler!(cs)
+    CoupledSimulation(config_file)
+    CoupledSimulation(config_dict)
 
-This function performs the coupling loop, which is the main part of the simulation.
-It runs the component models sequentially for one coupling timestep (`Δt_cpl`) at a time,
-and exchanges combined fields and calculates fluxes using the selected turbulent fluxes option.
-Note that we want to implement this in a dispatchable function to allow for
-other forms of timestepping (e.g. leapfrog).
+Set up a `CoupledSimulation` as prescribed by the given input.
```
juliasloan25, line 89: This could be a bit more descriptive, e.g.:
```suggestion
Set up a `CoupledSimulation` as prescribed by the given input.
This struct is defined in the Interfacer module and contains all
information about component models, diagnostics, timestepping,
output directories, etc needed to run a coupled simulation.
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -658,24 +596,26 @@ function setup_and_run(config_dict::AbstractDict)
...
-        solve_coupler!(cs)
+Keyword arguments
+==================
 
-        ## update the timespan to the correct values
-        cs.tspan[1] = tspan[1] + Δt_cpl * 2
-        cs.tspan[2] = tspan[2]
-    end
+`precompile`: If `true`, run the coupled simulations for two steps, so that most functions
+              are precompiled and timing are more accurate.
```
juliasloan25, line 611:
```suggestion
              are precompiled and subsequent timing will be more accurate.
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -687,37 +627,56 @@ function setup_and_run(config_dict::AbstractDict)
...
-    ## Use ClimaAtmos calculation to show the simulated years per day of the simulation (SYPD)
-    es = CA.EfficiencyStats((tspan[1], tspan[2]), walltime)
-    sypd = CA.simulated_years_per_day(es)
-    n_atmos_steps = atmos_sim.integrator.step
-    walltime_per_atmos_step = es.walltime / n_atmos_steps
+    simulated_seconds_per_second = float(cs.tspan[end] - cs.tspan[begin]) / walltime
+    simulated_years_per_day = simulated_seconds_per_second / (365.25 * walltime)
+    sypd = simulated_years_per_day
+    n_atmos_steps = cs.model_sims.atmos_sim.integrator.step
+    walltime_per_atmos_step = walltime / n_atmos_steps
```
juliasloan25, line 644: I think we should instead get the walltime per coupled step here. Of course atmos will dominate the time taken, but it's not the coupler's responsibility to track atmos performance. The only reason I can think we might want to keep it as walltime per atmos step would be for more direct comparisons to ClimaAtmos standalone runs

juliasloan25 (approved):

experiments/ClimaEarth/user_io/arg_parsing.jl
```diff
@@ -99,7 +99,7 @@ function get_coupler_args(config_dict::Dict)
     # Diagnostics information
     use_coupler_diagnostics = config_dict["use_coupler_diagnostics"]
     use_land_diagnostics = config_dict["use_land_diagnostics"]
-    (_, calendar_dt) = get_diag_period(t_start, t_end)
+    (_, diag_calendar_dt) = get_diag_period(t_start, t_end)
```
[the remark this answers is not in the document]
Sbozzolo: Changed

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -29,11 +28,15 @@ function postprocess_sim(cs, postprocessing_vars)
...
-        leaderboard_base_path = artifact_dir
-        compute_leaderboard(leaderboard_base_path, atmos_output_dir)
-        compute_pfull_leaderboard(leaderboard_base_path, atmos_output_dir)
+    # If we have enough data, plot the leaderboard
+    if "pfull" in CAN.available_vars(CAN.SimDir(atmos_output_dir))
```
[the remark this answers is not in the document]
Sbozzolo: "pfull" is pressure and it is needed to compute the leaderboard. I added a comment about this.

experiments/ClimaEarth/user_io/postprocessing.jl
Sbozzolo: Yes, you could get an error if `output_default_diagnostics == false` and don't have all the variables. But you could also have enough variables with `output_default_diagnostics == false`, meaning that this is not a good flag to check if we can plot the leaderboard or not. 

I removed it from `get_atmos_args`.

experiments/ClimaEarth/setup_run.jl
Sbozzolo: I think the reason was here was ineed to compare with Atmos standalone. I'll turn it into coupling step. I agree that it makes more sense.

PR #1253, opened 2025-04-01
add OceananigansSimulation template

juliasloan25:
## Purpose 
Add a file `oceananigans.jl` to implement the `OceananigansSimulation` in.

## To-do
- [x] Make a template file with function stubs
- [ ] Clarify exchange fields, where fluxes will be calculated/updated
- [ ] Fill in constructor
- [ ] Fill in setters and getters
- [ ] Fill in flux calculation function


Sbozzolo:
Implemented in #1325

PR #1268, opened 2025-04-10
remove `area_mask`

### LINKED ISSUE #1255: remove `area_mask`
We currently have both `area_fraction` and `area_mask` for each surface model, and use these e.g. to multiply the fluxes between a surface component and the atmosphere to get this component's contribution in this grid cell. This is logically flawed and potentially incorrect depending on the threshold used to generate `area_mask`.

For example:
A grid cell is 40% ocean, 60% land. Use a binary mask with a threshold of 50%, so we get `ocean_area_mask = 0`, `land_area_mask = 1` . When we combine fluxes, we'll get 0 contribution from land, and multiply the land contribution by 0.6 since it's covering 60% of the land. This defeats the purpose of the binary mask; we wanted to count the entire cell as land, but we do so incorrectly.

For now, our fluxes are still correct since we use a threshold of `eps(FT)` when creating the `area_mask`, but if we were to use a different threshold this would be incorrect. We should remove the idea of an `area_mask` and just use `area_fraction.` If we later on want an option to use the dominant surface model entirely in a grid cell, we can add that in as needed.

juliasloan25:
## Purpose 
closes #1255

## Content
- remove `area_mask` and use `area_fraction` instead, which is maintained to sum to 1 across components at each point
- remove the `mono_surface` option, which was rarely used and did not do what it was documented to do
- don't multiply by `area_fraction` in RHS calculations; only use `area_fraction` for flux calculations/exchanges
  - While multiplying the RHS by area fraction should be okay when using a binary area fraction, it may lead to incorrect values along coasts when using fractional area fractions because we multiply by the area fraction multiple times (fluxes and RHS). Now, the use of area fractions is limited to flux calcuations.

Note that simulations which previously used `mono_surface: true` will have changed results because we're no longer using fractional area fractions along coastlines. Among other things, this results in conservation changes - conservation is slightly improved after this PR.

### DIFF
--- a/config/longrun_configs/slabplanet_aqua_target_evolve_ocn.yml
+++ b/config/longrun_configs/slabplanet_aqua_target_evolve_ocn.yml
@@ -9,7 +9,6 @@ energy_check: false
 evolving_ocean: true
 land_albedo_type: "function"
 mode_name: "slabplanet_aqua"
-mono_surface: false
 start_date: "20100101"
 surface_setup: "PrescribedSurface"
 t_end: "120days"
--- a/experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
+++ b/experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
@@ -1,5 +1,6 @@
 import Test: @test, @testset
 import Dates
+import NCDatasets
 import ClimaCore as CC
 import Thermodynamics.Parameters as TDP
 import ClimaParams # required for TDP
@@ -68,8 +69,7 @@ for FT in (Float32, Float64)
         # check that extracting expected T due to input atmopsheric fluxes
         dY, Y, p = test_sea_ice_rhs(F_radiative = 1.0)
         dT_expected = -1.0 / (p.params.h * p.params.ρ * p.params.c)
-        @test minimum(dY) ≈ FT(dT_expected)
-        @test maximum(dY) ≈ FT(0)
+        @test all(extrema(dY) .≈ FT(dT_expected))
 
         # check that tendency will not result in above freezing T
         dY, Y, p = test_sea_ice_rhs(F_radiative = 0.0, T_base = 330.0) # Float32 requires a large number here!
--- a/experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
+++ b/experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
@@ -41,12 +41,11 @@ for FT in (Float32, Float64)
             # Get initial SIC values and use them to calculate ice fraction
             SIC_init = CC.Fields.zeros(space)
             evaluate!(SIC_init, SIC_timevaryinginput, t_start)
-            ice_fraction = get_ice_fraction.(SIC_init, false)
 
             cache = (;
                 F_turb_energy = CC.Fields.zeros(space),
                 F_radiative = CC.Fields.zeros(space) .+ FT(F_radiative),
-                area_fraction = ice_fraction,
+                area_fraction = SIC_init,
                 SIC_timevaryinginput = SIC_timevaryinginput,
                 land_fraction = CC.Fields.zeros(space),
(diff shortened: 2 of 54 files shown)


juliasloan25:
`Slabplanet: default` conservation

before this PR ([build](https://buildkite.com/clima/climacoupler-ci/builds/6145#_))
```
[ Info: Conservation Check Plots
[ Info: ClimaCoupler.ConservationChecker.EnergyConservationCheck
[ Info: 0.00010401177251925538
[ Info: ClimaCoupler.ConservationChecker.WaterConservationCheck
[ Info: 0.0003231120128117549
```

this PR (build)
```
[ Info: Conservation Check Plots
[ Info: ClimaCoupler.ConservationChecker.EnergyConservationCheck
[ Info: 0.00010316326319205847
[ Info: ClimaCoupler.ConservationChecker.WaterConservationCheck
[ Info: 0.00019193406452284358
```

src/FluxCalculator.jl
```diff
@@ -492,51 +479,70 @@ function compute_surface_fluxes!(
...
     # calculate the surface fluxes
-    fluxes = FluxCalculator.get_surface_fluxes!(inputs, surface_params, area_fraction)
-    (; F_turb_ρτxz, F_turb_ρτyz, F_shf, F_lhf, F_turb_moisture) = fluxes
+    fluxes = FluxCalculator.get_surface_fluxes!(inputs, surface_params)
+    (; F_turb_ρτxz, F_turb_ρτyz, F_shf, F_lhf, F_turb_moisture, L_MO, ustar, buoyancy_flux) = fluxes
+
+
+    # Zero out fluxes where the area fraction is zero
+    # Multiplying by `area_fraction` is not sufficient because the area fraction
+    # may be approximately, but not exactly, zero.
```
Sbozzolo, line 488: Maybe we should just set `area_fraction` to 0 when it is close to zero.

config/longrun_configs/slabplanet_aqua_target_evolve_ocn.yml
```diff
@@ -9,7 +9,7 @@ energy_check: false
 evolving_ocean: true
 land_albedo_type: "function"
 mode_name: "slabplanet_aqua"
-mono_surface: false
+
```
Sbozzolo, line 12: Do you want to add a blank line to all these files?

juliasloan25: no 😅

src/FluxCalculator.jl
juliasloan25: Something else must be going on because when I do that and just multiply fluxes by `area_fraction` instead of explicitly making these checks, the atmosphere is stable when it first steps

Sbozzolo: I think that this is worth understanding better

juliasloan25: I looked into it more. Multiplying by 0 doesn't work because we sometimes have NaNs in the fluxes where the area fraction is 0. We will need to manually substitute 0 anywhere the area fraction is 0, and if/when we want to support fractional area fractions, we'll need to also multiply by the area fraction

Sbozzolo: Okay, thanks for looking into it!

experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
```diff
@@ -1,5 +1,6 @@
 import Test: @test, @testset
 import Dates
+import NCDatasets
```
Sbozzolo, line 3: Is this needed?

Sbozzolo (approved):

experiments/ClimaEarth/test/component_model_tests/prescr_seaice_tests.jl
juliasloan25: Yes, it's needed because we call `TimeVaryingInput` in this file. I guess we had only ever run this test script after including other files that imported NCDatasets, but now this file is self-contained

PR #1273, opened 2025-04-14
ED only for target longrun tests

trontrytel:
Use EDonly sub-grid scale ClimaAtmos model in `TARGET IDEALIZED` long runs tests.


szy21 (approved):

szy21:
I'm going to merge this manually as it only changes longrun configs, so we don't need to run ci.

PR #1277, opened 2025-04-17
Remove hierarchy tests and unused argument

Sbozzolo:
Hierarchy tests and experiments have become stale and started being more burdensome to maintain. This commit removes these experiments. These experiments can be reproduced in some form from the AMIP driver.

This commit also removes an unused argument in the `import_atmos_fields!` function

The last commit that contains hierarchy tests is: a6557a3b


juliasloan25 (approved):

juliasloan25:
Could you also please remove the `ClimaCoupler-Hierarchies` entry in `.buildkite/README.md`? And maybe add a NEWS.md entry for this change

PR #1278, opened 2025-04-18
remove `nans_to_zero`

### LINKED ISSUE #1267: remove `nans_to_zero`
We're currently using `nans_to_zero` to zero out component model fields where that component model is masked out (or at least that was the intention). In reality, we zero out all NaNs independent of where they appear on the grid. We should instead set NaN to zero only where the area fraction for a given model is zero.

This was recently done for the [integrated land model](https://github.com/CliMA/ClimaCoupler.jl/blob/cfa9bad92395a8551826a3412f3cd51dd051aa00/experiments/ClimaEarth/components/land/climaland_integrated.jl#L565-L567) - we can use this pattern more generally.

juliasloan25:
## Purpose 
Instead of zeroing out all NaNs in a surface field, we zero out all values where the area fraction is 0. This ensures that a surface model does indeed not contribute where it isn't present, and that meaningful (i.e. bug- or instability-indicating) NaNs are not removed.

closes #1267

This PR is similar in intention to #1268

### DIFF
--- a/src/FieldExchanger.jl
+++ b/src/FieldExchanger.jl
@@ -266,13 +266,6 @@ function dummmy_remap!(target, source)
 end
 
 """
-    nans_to_zero(v)
-
-Replaces NaNs with zeros, otherwise returns the value.
-"""
-nans_to_zero(v) = isnan(v) ? typeof(v)(0) : v
-
-"""
     combine_surfaces!(combined_field::CC.Fields.Field, sims, field_name::Val)
 
 Sums the fields, specified by `field_name`, weighted by the respective area fractions of all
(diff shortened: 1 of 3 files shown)


src/FieldExchanger.jl
```diff
@@ -287,11 +280,15 @@ surface simulations. THe result is saved in `combined_field`.
...
+    combined_field .= 0
     for sim in sims
         if sim isa Interfacer.SurfaceModelSimulation
+            # Zero out the contribution from this surface if the area fraction is zero
+            # Note that multiplying by `area_fraction` is not sufficient in the case of NaNs
+            area_fraction = Interfacer.get_field(sim, Val(:area_fraction))
             combined_field .+=
-                Interfacer.get_field(sim, Val(:area_fraction)) .* nans_to_zero.(Interfacer.get_field(sim, field_name)) # this ensures that unitialized (masked) areas do not affect (TODO: move to mask / remove)
+                area_fraction .*
+                ifelse.(area_fraction .== 0, zero(combined_field), Interfacer.get_field(sim, field_name))
```
Sbozzolo, line 291: You probably want to use approx instead of ==

Sbozzolo (approved):

PR #1279, opened 2025-04-18
Remove conservation_softfail

Sbozzolo:
Now the job will just fail when conservation is not respected.


experiments/ClimaEarth/user_io/debug_plots.jl
```diff
@@ -23,8 +23,7 @@ relative to the initial value;
 """
 function plot_global_conservation(
```
juliasloan25, line 24: docstring needs to be updated too

juliasloan25 (approved):

Sbozzolo:
I think I'll change my approach here and isolate a function to compute conservation (from the plotting routine)

Sbozzolo:
I tried coming up with a cleaner way to do this, but I could not find a clean way to have that some jobs fail upon failing conservation and others don't.

PR #1283, opened 2025-04-21
Fix calibration test, lower resource requirements

### LINKED ISSUE #1265: Longrun calibration test fails
[Here](https://buildkite.com/clima/climacoupler-longruns/builds/850#01960a1e-3185-4722-843d-dd45d539c78f) is the most recent failing build. This is caused by NaNs somewhere in the observation map, I am not exactly sure where since the model runs themselves do not fail.

nefrathenrici:
## Purpose 

This PR fixes the longrun calibration test, closes #1265

## To-do

- Re-run the longrun test from this commit to ensure that the tests pass. In-progress build [here](https://buildkite.com/clima/climacoupler-longruns/builds/861). An existing passing build is [here](https://buildkite.com/clima/climacoupler-longruns/builds/857)

## Content

- Fixed the parameter filepath argument in `model_interface.jl` and the filepath in the observation generation. These changes alone fix the calibration test.
- Use Unscented Kalman Inversion instead of the default scheme, reducing the number of ensemble members and resources required to run the test
- Set a lower horizontal resolution and a higher timestep for the model config

----

### DIFF
--- a/experiments/calibration/run_calibration.jl
+++ b/experiments/calibration/run_calibration.jl
@@ -5,14 +5,16 @@ using ClimaAnalysis
 import ClimaAnalysis: SimDir, get, slice, average_xy
 import ClimaComms
 import EnsembleKalmanProcesses: I, ParameterDistributions.constrained_gaussian
+import EnsembleKalmanProcesses as EKP
+using Statistics
 
 # Ensure ClimaComms doesn't use MPI
 ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
 ClimaComms.@import_required_backends
 
-single_member_dims = (1,)
+single_member_dims = 1
 function CAL.observation_map(iteration)
-    G_ensemble = Array{Float64}(undef, single_member_dims..., ensemble_size)
+    G_ensemble = Array{Float64}(undef, single_member_dims, ensemble_size)
 
     for m in 1:ensemble_size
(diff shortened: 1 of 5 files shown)


experiments/calibration/run_calibration.jl
```diff
@@ -68,14 +68,20 @@ if !isfile(obs_path)
...
+    observations .= process_member_data(SimDir(joinpath(obs_output_dir, "model_config/output_active/clima_atmos")))
     JLD2.save_object(obs_path, observations)
 end
-
-# Initialize experiment data
-@everywhere observations = JLD2.load_object(obs_path)
-
-eki = CAL.calibrate(CAL.WorkerBackend, ensemble_size, n_iterations, observations, noise, prior, output_dir)
+observations = JLD2.load_object(obs_path)
+@show observations
```
juliasloan25, line 75: Is this `show` meant to be here, or was it for debugging?

juliasloan25 (approved):

experiments/calibration/run_calibration.jl
nefrathenrici: It was for debugging, but it would be nice to keep around to debug quickly from a buildkite log if this fails in the future.

juliasloan25: Okay we can keep it!

PR #1286, opened 2025-04-21
allow < 3 surface models

### LINKED ISSUE #1069: Don't require all 4 component models
Or equivalently, allow < 3 surface models.

Currently, running a coupled simulation requires creating atmosphere, land, ocean, _and_ ice models. This means that if we run a simulation with only one or two models (e.g. atmosphere and land), we still have to initialize all 4, then we zero out the area fraction of the unused models.

It seems pretty straightforward to transition to require atmos and >= 1 surface model. In the future we may also want to be able to run the atmosphere without any surface model, but we need to think a little bit more about what the fluxes would look like in this case, and similarly for 1 surface model only. These could also be use cases we don't support.

juliasloan25:
## Purpose 
Don't hardcode 4 component models (atmos, land, ocean, sea ice).
closes #1069

## To-do
- [x] generalize output writer closing via `Interfacer.close_output_writers!` function
- [x] in `model_sims`, only include surface models that are present
- [x] write `update_surface_fractions` to handle case of a surface model not present

## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,12 @@ ClimaCoupler.jl Release Notes
 
 ### ClimaCoupler features
 
+#### Allow < 3 surface models. PR[#1286](https://github.com/CliMA/ClimaCoupler.jl/pull/1286)
+
+Previously, land, ocean, and sea ice were all required to be defined, and the
+area fraction of a model would be set to 0 to exclude it. Now, a `CoupledSimulation`
+can be created with fewer than 4 components in `model_sims`.
+
 #### Change signature for `turbulent_fluxes!`. PR[#1327](https://github.com/CliMA/ClimaCoupler.jl/pull/1327)
 
 `FluxCalculator.turbulent_fluxes!` can now be called in two ways:
@@ -21,7 +27,7 @@ The new signature simplifies calling the function, ensures that the mutating
 convention is respected (`turbulent_fluxes!` mutates `coupler_fields`), and
 removes unnecessary arguments and type restrictions.
 
-Similarly, `get_surface_fluxes!` was renamed to `get_surface_fluxes`. 
+Similarly, `get_surface_fluxes!` was renamed to `get_surface_fluxes`.
 
 #### Simplify initial component model exchange. PR[#1305](https://github.com/CliMA/ClimaCoupler.jl/pull/1305)
 
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -97,7 +97,8 @@ end
 
 # Extension of CA.set_surface_albedo! to set the surface albedo to 0.38 at the beginning of the simulation,
 # so the initial callback initialization doesn't lead to NaNs in the radiation model.
-# Subsequently, the surface albedo will be updated by the coupler, via water_albedo_from_atmosphere!.
+# Subsequently, the surface albedo will be updated by the coupler, via
+# water_albedo_from_atmosphere!.
 function CA.set_surface_albedo!(Y, p, t, ::CA.CouplerAlbedo)
     if float(t) == 0
         FT = eltype(Y)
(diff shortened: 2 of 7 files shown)


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -393,10 +392,17 @@ function Interfacer.add_coupler_fields!(coupler_field_names, atmos_sim::ClimaAtm
     push!(coupler_field_names, atmos_coupler_fields...)
 end
 
+"""
+    Interfacer.close_output_writers!(sim::ClimaAtmosSimulation)
+
+Close all output writers used by the atmos simulation.
+"""
+Interfacer.close_output_writers!(sim::ClimaAtmosSimulation) =
```
Sbozzolo, line 400: I don't think this should be a mutating function. In Julia, operating on the enviroment (e.g., IO) is not considered mutating. (It's `close` not `close!`)

experiments/ClimaEarth/components/ocean/slab_ocean.jl
```diff
@@ -203,3 +218,78 @@ Perform DSS on the state of a component simulation, intended to be used
 before the initial step of a run. This method acts on slab ocean model sims.
 """
 dss_state!(sim::SlabOceanSimulation) = CC.Spaces.weighted_dss!(sim.integrator.u, sim.integrator.p.dss_buffer)
+
+"""
+    FieldExchanger.update_sim!(::SlabOceanSimulation, csf, area_fraction)
+
+Updates the air density (needed for the turbulent flux calculation)
+and the direct and diffuse albedos of the ocean.
```
Sbozzolo, line 226:
```suggestion
Update air density and velocity (needed for the turbulent flux calculation), 
and the direct and diffuse albedos of the ocean.
```

experiments/ClimaEarth/components/ocean/slab_ocean.jl
```diff
@@ -203,3 +218,78 @@ Perform DSS on the state of a component simulation, intended to be used
...
+    update_field!(sim, Val(:v_atmos), csf.v_int)
+
+    # Update the direct and diffuse albedos with the new atmospheric wind
+    set_albedos!(sim, sim.integrator.t)
+end
+
+"""
+    FieldExchanger.import_atmos_fields!(csf, sim::SlabOceanSimulation, atmos_sim)
+
+Imports quantities from the coupled simulation fields into the ocean simulation.
```
Sbozzolo, line 240:
```suggestion
Import quantities from the coupled simulation fields into the ocean simulation.
```

experiments/ClimaEarth/components/ocean/slab_ocean.jl
```diff
@@ -203,3 +218,78 @@ Perform DSS on the state of a component simulation, intended to be used
...
+
+Set the direct and diffuse albedos of the ocean based on the current date and
+the atmospheric wind. The albedos are calculated using the `surface_albedo_direct`
+and `surface_albedo_diffuse` functions from the `ClimaAtmos` module, so this
+is dependent on running with `ClimaAtmosSimulation` as the atmosphere simulation.
+"""
+function set_albedos!(sim::SlabOceanSimulation, t)
+    u = sim.integrator.u
+    p = sim.integrator.p
+    FT = eltype(u)
```
Sbozzolo, line 271:
```suggestion
    FT = CC.Spaces.undertype(axes(u))
```
This ensures that we get the `FT` even when `u` is not a scalr

Sbozzolo (approved):

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
juliasloan25: I didn't know that, thank you!

experiments/ClimaEarth/setup_run.jl
```diff
@@ -384,7 +369,10 @@ function CoupledSimulation(config_dict::AbstractDict)
     =#
 
     ## model simulations
-    model_sims = (atmos_sim = atmos_sim, ice_sim = ice_sim, land_sim = land_sim, ocean_sim = ocean_sim)
+    model_sims = (; atmos_sim, ice_sim, land_sim, ocean_sim)
+
+    # remove models that are not used
+    model_sims = filter(sim -> !isnothing(sim), model_sims)
```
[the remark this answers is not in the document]
juliasloan25: This works in Julia 1.11, but Julia 1.10 doesn't support filtering NamedTuples so I had to change this anyway :/

NEWS.md
```diff
@@ -6,6 +6,12 @@ ClimaCoupler.jl Release Notes
 
 ### ClimaCoupler features
 
+#### Allow < 3 surface models. PR[#1286](https://github.com/CliMA/ClimaCoupler.jl/pull/1286)
+
```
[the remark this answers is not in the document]
juliasloan25: < 3 surface models and < 4 component models are equivalent because the atmosphere is a component but not a surface model

PR #1293, opened 2025-04-23
create atmos, surface thermo states from remapped T, q, rho

### LINKED ISSUE #1291: construct thermo states for atmos and surface from T, q, rho
Currently, we get `thermo_state_int` from the atmosphere and `thermo_state_sfc` from the surface, and use these to compute surface fluxes. This works now only because we don't do any remapping. In a more remapping-safe way, we can instead get `T_atmos`, `q_atmos`, and `rho_atmos` from the atmosphere, and `T_sfc`, `q_sfc`, and `rho_sfc` and use these to compute the two thermo states that are used to compute fluxes.

juliasloan25:
## Purpose 
closes #1291

As of #1305, we have a generalized function to compute `q`, choosing phase based on `T_sfc`. This PR uses that function to compute the atmosphere and surface thermo states as needed, rather than exchanging these objects. This will allow us to remap T, q, rho, and construct the thermo states on the exchange grid when computing fluxes.

This unified approach allows us to remove a couple functions, including `surface_thermo_state` and `FluxCalculator.calculate_surface_air_density`.

Previously, we computed radiation before setting the surface ("skin") temperature of the integrated land, which led to immediate instability. Now, the land surface temperature is set at initialization, and we compute radiation + exchange before calling `set_initial_cache` for the land, so the correct radiation drivers have been set when we finish filling the land cache. I expect that this is also a bug on main but wasn't as pronounced since here we use the initial `T_sfc` more extensively to compute `rho_sfc`, `q_sfc`, atmos thermo state, etc.

Visually inspecting the plots from buildkite, they look identical to main.

## Content
- [x] add generalized function to compute q, choosing phase based on T_sfc - done in #1305
- [x] store atmos T, q, rho in coupler fields so we can easily remap them
- [x] store surface T, q in coupler fields; compute rho as needed
- [x] create atmosphere and surface thermo states from the respective T, q, rho on the exchange grid when computing fluxes
- [x] remove `surface_thermo_state` function
- [x] remove FluxCalculator `calculate_surface_air_density`
- [x] [additional] rename integrated land atmosphere properties to unify with the rest of the codebase

----

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -226,6 +226,8 @@ Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:air_temperature}) =
         sim.integrator.p.params.thermodynamics_params,
         CC.Fields.level(sim.integrator.p.precomputed.ᶜts, 1),
     )
+Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:air_density}) =
+    TD.air_density.(sim.integrator.p.params.thermodynamics_params, CC.Fields.level(sim.integrator.p.precomputed.ᶜts, 1))
 function Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:cos_zenith})
     hasradiation(sim.integrator) || return nothing
     return CC.Fields.array2field(
@@ -271,7 +273,6 @@ function Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:LW_d})
 end
 Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:specific_humidity}) =
     CC.Fields.level(ρq_tot(sim.integrator.p.atmos.moisture_model, sim.integrator) ./ sim.integrator.u.c.ρ, 1)
-Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:air_density}) = CC.Fields.level(sim.integrator.u.c.ρ, 1)
 Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:radiative_energy_flux_sfc}) =
     surface_radiation_flux(sim.integrator.p.atmos.radiation_mode, sim.integrator)
 Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:snow_precipitation}) =
@@ -293,11 +294,21 @@ function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:surface_temp
     # it.
     # The rrtmgp_model.surface_temperature field is updated by the RRTMGP callback using the
     # sfc_conditions.ts, so all we need to do is update sfc_conditions.ts
+
+    # Note also that this method updates the entire thermo state, not just the temperature
+
+    ρ_sfc =
+        FluxCalculator.extrapolate_ρ_to_sfc.(
+            get_thermo_params(sim),
+            sim.integrator.p.precomputed.sfc_conditions.ts,
+            csf.T_sfc,
+        )
+
     if sim.integrator.p.atmos.moisture_model isa CA.DryModel
-        sim.integrator.p.precomputed.sfc_conditions.ts .= TD.PhaseDry_ρT.(get_thermo_params(sim), csf.ρ_sfc, csf.T_sfc)
+        sim.integrator.p.precomputed.sfc_conditions.ts .= TD.PhaseDry_ρT.(get_thermo_params(sim), ρ_sfc, csf.T_sfc)
     else
         sim.integrator.p.precomputed.sfc_conditions.ts .=
-            TD.PhaseNonEquil_ρTq.(get_thermo_params(sim), csf.ρ_sfc, csf.T_sfc, TD.PhasePartition.(csf.q_sfc))
+            TD.PhaseNonEquil_ρTq.(get_thermo_params(sim), ρ_sfc, csf.T_sfc, TD.PhasePartition.(csf.q_sfc))
     end
 end
 Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:height_int}) =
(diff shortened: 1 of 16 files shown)


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -290,17 +292,26 @@ Interfacer.get_field(sim::ClimaAtmosSimulation, ::Val{:thermo_state_int}) =
...
     ρq_tot(sim.integrator.p.atmos.moisture_model, sim.integrator)
+# TODO this is updating the ts- rename
 function Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:surface_temperature}, csf)
     # NOTE: This update_field! takes as argument the entire coupler fields struct, instead
     # of a single field. This is unlike most of other functions, so we may want to revisit
     # it.
     # The rrtmgp_model.surface_temperature field is updated by the RRTMGP callback using the
     # sfc_conditions.ts, so all we need to do is update sfc_conditions.ts
+
+    ρ_sfc =
```
juliasloan25, line 303: we should store this in an atmosphere temp field if we can

also, csf.T_sfc will need to be regridded to the atmosphere space here

experiments/ClimaEarth/setup_run.jl
```diff
@@ -529,12 +529,10 @@ function CoupledSimulation(config_dict::AbstractDict)
         # called below at reinit). Turbulent fluxes in atmos are updated in `update_model_sims`.
         Interfacer.update_field!(atmos_sim, Val(:surface_temperature), cs.fields)
 
-
         # 5.reinitialize models + radiative flux: prognostic states and time are set to their initial conditions. For atmos, this also triggers the callbacks and sets a nonzero radiation flux (given the new sfc_conditions)
         FieldExchanger.reinit_model_sims!(cs.model_sims)
 
         # 6.update all fluxes.
-        FieldExchanger.import_atmos_fields!(cs.fields, cs.model_sims)
```
juliasloan25, line 537: I don't know why we were importing atmos fields here. Maybe because combined fluxes were computed in atmos and we needed to retrieve them. But with partitioned fluxes I don't see a need for it

src/FluxCalculator.jl
```diff
@@ -293,17 +260,26 @@ function compute_surface_fluxes!(
...
 
+    # construct the atmospheric thermo states
+    thermo_state_atmos = TD.PhaseEquil_ρTq.(thermo_params, csf.ρ_atmos, csf.T_atmos, csf.q_atmos)
+
+    # construct the surface thermo state
+    # get surface air density by extrapolating atmospheric density to the surface
+    ρ_sfc = extrapolate_ρ_to_sfc.(Ref(thermo_params), thermo_state_atmos, T_sfc)
+
+    # compute surface humidity from the surface temperature, surface density, and phase
+    q_sfc = TD.q_vap_saturation_generic.(thermo_params, csf.T_sfc, ρ_sfc, get_field(sim, Val(:phase), thermo_params))
```
juliasloan25: use `FieldExchanger.compute_surface_humidity!` instead (move to FluxCalculator?)

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -156,7 +156,9 @@ function ClimaLandSimulation(
     Y.snow.S_l .= FT(0)
     Y.snow.U .= FT(0)
 
-    set_initial_cache! = CL.make_set_initial_cache(model)
+    # Initialize the surface temperature so the atmosphere can compute radiation
```
Sbozzolo, line 159: Let's maybe add a little bit more detail here (so that people can get more context in the future)

experiments/ClimaEarth/setup_run.jl
```diff
@@ -483,22 +483,22 @@ function CoupledSimulation(config_dict::AbstractDict)
...
             FieldExchanger.exchange!(cs)
         end
 
+        # 3. Set all initial cache values for the land model, now that we have updated drivers and radiative fluxes
+        land_set_initial_cache! = CL.make_set_initial_cache(cs.model_sims.land_sim.model)
+        land_set_initial_cache!(
+            cs.model_sims.land_sim.integrator.p,
+            cs.model_sims.land_sim.integrator.u,
+            cs.model_sims.land_sim.integrator.t,
+        )
```
Sbozzolo, line 500: Maybe we can group these two and have a function `set_caches!(cs)`. This would also be a good place where to explain the intricacies of the inter-dependencies.

juliasloan25:
I triggered a longrun from a branch rebased on this one, and AMIP + integrated land was stable for 6+ months, whereas before it was stable for only ~4 months

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -420,20 +427,32 @@ function FieldExchanger.update_sim!(sim::ClimaAtmosSimulation, csf)
...
+Finish setting the cache for the atmosphere, after the initial exchange.
+Radiation must be called here because it requires the surface model initial
+temperatures, which are set in the atmosphere during the initial exchange.
+Any other callbacks which modify the cache should be called here as well.
 """
-function FluxCalculator.calculate_surface_air_density(sim::ClimaAtmosSimulation, T_sfc::CC.Fields.Field)
-    thermo_params = get_thermo_params(sim)
-    ts_int = Interfacer.get_field(sim, Val(:thermo_state_int))
-    FluxCalculator.extrapolate_ρ_to_sfc.(Ref(thermo_params), ts_int, Utilities.swap_space!(axes(ts_int.ρ), T_sfc))
+function Interfacer.set_cache!(sim::ClimaAtmosSimulation)
```
Sbozzolo, line 441: Let's note in the comments that this function does not set all the cache variables, because many are computed as part of the tendendencies.

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -420,20 +427,32 @@ function FieldExchanger.update_sim!(sim::ClimaAtmosSimulation, csf)
...
+        CA.rrtmgp_model_callback!(cs.model_sims.atmos_sim.integrator)
+        if pkgversion(CA) == v"0.30" &&
+           !isnothing(cs.model_sims.atmos_sim.integrator.p.atmos.non_orographic_gravity_wave)
+            # In version 0.30, nogw_model_callback crashes when there are no gravity waves,
+            # see CA #3792
+            CA.nogw_model_callback!(cs.model_sims.atmos_sim.integrator)
+        else
+            pkgversion(CA) > v"0.30" && CA.nogw_model_callback!(cs.model_sims.atmos_sim.integrator)
+        end
+    end
```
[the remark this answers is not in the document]
Sbozzolo: This comment is correct

Sbozzolo (approved):

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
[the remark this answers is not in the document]
Sbozzolo: Yes, make a PR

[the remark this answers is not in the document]
juliasloan25: I'll fix this in this PR since it won't run as is

PR #1303, opened 2025-04-24
use extra fluxes in ClimaLandSim flux calculation

juliasloan25:
## Purpose 
To couple with ClimaAtmos using EDMF, the integrated land model needs to provide L_MO, ustar, and buoyancy flux in its flux calculation. As of ClimaLand v0.16.0, the integrated land model returns the buoyancy flux. L_MO and ustar are computed here, using the combined momentum fluxes across land model components

Also updates dependencies:
```
    Updating `~/clima/ClimaCoupler.jl/experiments/ClimaEarth/Project.toml`
  [29b5916a] ↑ ClimaAnalysis v0.5.16 ⇒ v0.5.17
  [b2c96348] ↑ ClimaAtmos v0.29.1 ⇒ v0.30.0
  [d414da3d] ↑ ClimaCore v0.14.31 ⇒ v0.14.33
  [5c42b081] ↑ ClimaParams v0.10.27 ⇒ v0.10.28
  [595c0a79] ↑ ClimaTimeSteppers v0.8.3 ⇒ v0.8.4
  [0bca4576] ↑ SciMLBase v2.85.0 ⇒ v2.86.2
```

## Content
- [x] get buoyancy flux from ClimaLand flux calculation functions (requires ClimaLand v0.16.0)
- [x] compute ustar from the buoyancy flux combined across land surfaces
- [x] compute monin obukhov length from ustar, buoyancy flux following [SurfaceFluxes.jl](https://github.com/CliMA/SurfaceFluxes.jl/blob/48fe3012755c14fc2910d6e5f36dd2d57be036ed/src/SurfaceFluxes.jl#L459-L461)

----


experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -594,13 +596,36 @@ function FluxCalculator.compute_surface_fluxes!(
...
+
+    # Combine the buoyancy flux from each component of the land model
+    # Note that we exclude the canopy component here for now, since ClimaLand doesn't
+    #  include its extra resistance term in the buoyancy flux calculation.
+    @. csf.temp1 =
+        soil_dest.buoy_flux * (1 - p.snow.snow_cover_fraction) + p.snow.snow_cover_fraction * snow_dest.buoy_flux
+    @. csf.temp1 = ifelse(area_fraction == 0, zero(csf.temp1), csf.temp1)
+    @. csf.buoyancy_flux += csf.temp1 * area_fraction
+
+    # Compute ustar from the momentum fluxes and surface air density
```
kmdeck, line 617: Maybe add a formula?

kmdeck (approved):
This looks correct to me. It might be good to look at some output values and make sure they look reasonable

juliasloan25:
> This looks correct to me. It might be good to look at some output values and make sure they look reasonable

I looked over the outputs with Akshay and he confirmed that they look fine, though it's a little hard to tell what we should expect because the atmosphere starts with 0 moisture, which is very different from an equilibrated state. We agreed it looks okay to merge, and we can check the values in the longruns this weekend

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: I added the formulas we're using for ustar and L_MO

Issue #1261, opened 2025-04-03
remove `get_new_cache` function for Bucket


juliasloan25:
For the Bucket, we use `get_new_cache` to allocate space in the cache when we want to track energy conservation. We should instead just always allocate space for this in the bucket model, and we can remove `get_new_cache`. See https://github.com/CliMA/ClimaLand.jl/issues/1091

PR #1304, opened 2025-04-24
Use bucket cache `e_per_area`

juliasloan25:
## Purpose 
https://github.com/CliMA/ClimaLand.jl/pull/1104 adds `e_per_area` to the Bucket cache, independent of whether conservation checks are enabled. This allows us to access that field directly and simplify the coupler code. It's tested out here.

closes https://github.com/CliMA/ClimaCoupler.jl/issues/1261

## To-do
- [x] access `p.e_per_area` directly in `get_field(::BucketSimulation, Val(:energy))`
- [x] remove `get_new_cache` method for Bucket
- [x] don't pass `energy_check` to Bucket constructor


juliasloan25:
superseded by #1437

PR #1305, opened 2025-04-25
simplify initial exchange

### LINKED ISSUE #331: simplify initial component model exchange
Because we cannot easily set the `thermo_state` and `sfc_conditions` in Atmos or abstract the calculation of `q_sfc` from the bucket, we need to initialize the respective component models in a complicated way. i.e. we have additional `step!` calls of surface sims before solving simulations (step 2 in driver).

Currently, the bucket `q_sfc` depends on atmos density, so we need to pass atmos info to bucket then step bucket once to calculate its `q_sfc`. We want to be able to do this calculation independently (without stepping model).

Ideally we'd like something like:
- get integrators of all component models with states already initialized (this includes `thermo_state` in atmos)
- update coupler with rho_sfc from atmos 
- calculate q_sfc in surface models (the `q_sfc` calculating functions should be abstractable)
- exchange surface states (incl. `q_sfc`) with the atmosphere
- calculate radiative and turbulent surface fluxes
- exchange surface fluxes

This will require collaboration with Atmos and ClimaLSM. We should only implement this after partial fluxes are implemented. 

Note: `reinit!` updates the state to be at its initial conditions, and then (by default) updates the cache and callbacks based on these values.

### Specific changes needed
- [ ] isolate `q_sfc` calculation in ClimaLand from solve/step
  - maybe we can call `update_aux` directly to update `q_sfc` in-place in `p`, but we need to make sure that this is being evaluated with the correct state values
- [ ] run coupled model interactively to make sure that at each step of the process, all values look as we would expect
- [ ] might need a wrapper function in atmos that calls `set_precomputed_quantities` and radiation callback (and potentially other surface condition callbacks) - to replace `reinit!`

juliasloan25:
## Purpose 
This PR does 2 things which depend on each other:
- compute `q_sfc` from atmosphere state and surface temperature, rather than sometimes computing and sometimes storing/retrieving in surface model cache
- simplifies the initial component model exchange, and reorders the `step!` function to align with it

tested longruns pipeline [here](https://buildkite.com/clima/climacoupler-longruns/builds/865)
compare to previous longrun build [here](https://buildkite.com/clima/climacoupler-longruns/builds/864)
Most of the longruns fail in both cases, but we see that `GPU AMIP FINE: new target amip: topo` reaches OOM shortly after day 20 in both cases, and `GPU AMIP FINE: new target amip: topo + diagedmf` is stable past 34 weeks 4 days in both cases.

closes #331

## Content
- [x] add a function to compute `q_sfc` from the atmosphere thermo properties and surface temperature
- [x] simplify initial component model exchange (don't need a step to compute bucket `q` anymore; call radiation callback directly instead of `reinit!`)
- [x] reorder `step!` to align with initial exchange
- [x] remove surface model `get_field(_, Val(:surface_humidity))` functions

## Next steps (to be done in #1293)
- store atmos T, q, rho in coupler fields so we can easily remap them
- store surface T, q in coupler fields; compute rho as needed
- remove `surface_thermo_state` function
- create atmosphere, surface thermo states from the respective T, q, rho when computing fluxes

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_bucket.jl
+++ b/experiments/ClimaEarth/components/land/climaland_bucket.jl
@@ -250,8 +250,6 @@ Interfacer.get_field(sim::BucketSimulation, ::Val{:surface_direct_albedo}) =
     CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
 Interfacer.get_field(sim::BucketSimulation, ::Val{:surface_diffuse_albedo}) =
     CL.surface_albedo(sim.model, sim.integrator.u, sim.integrator.p)
-Interfacer.get_field(sim::BucketSimulation, ::Val{:surface_humidity}) =
-    CL.surface_specific_humidity(nothing, sim.model, sim.integrator.u, sim.integrator.p, sim.integrator.t)
 Interfacer.get_field(sim::BucketSimulation, ::Val{:surface_temperature}) =
     CL.surface_temperature(sim.model, sim.integrator.u, sim.integrator.p, sim.integrator.t)
 
@@ -336,16 +334,27 @@ end
 function FluxCalculator.surface_thermo_state(
     sim::BucketSimulation,
     thermo_params::TD.Parameters.ThermodynamicsParameters,
-    thermo_state_int,
+    atmos_sim::Interfacer.AtmosModelSimulation,
 )
-
     T_sfc = Interfacer.get_field(sim, Val(:surface_temperature))
+    FieldExchanger.compute_surface_humidity!(
+        sim.integrator.p.bucket.q_sfc,
+        Interfacer.get_field(atmos_sim, Val(:air_temperature)),
+        Interfacer.get_field(atmos_sim, Val(:specific_humidity)),
+        Interfacer.get_field(atmos_sim, Val(:air_density)),
+        T_sfc,
+        thermo_params,
+    )
     # Note that the surface air density, ρ_sfc, is computed using the atmospheric state at the first level and making ideal gas
     # and hydrostatic balance assumptions. The land model does not compute the surface air density so this is
     # a reasonable stand-in.
-    ρ_sfc = FluxCalculator.extrapolate_ρ_to_sfc.(thermo_params, thermo_state_int, T_sfc) # ideally the # calculate elsewhere, here just getter...
-    q_sfc = Interfacer.get_field(sim, Val(:surface_humidity)) # already calculated in rhs! (cache)
-    @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, q_sfc)
+    ρ_sfc =
+        FluxCalculator.extrapolate_ρ_to_sfc.(
+            thermo_params,
+            Interfacer.get_field(atmos_sim, Val(:thermo_state_int)),
+            T_sfc,
+        )
+    return @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, sim.integrator.p.bucket.q_sfc)
 end
 
 """
--- a/experiments/ClimaEarth/setup_run.jl
+++ b/experiments/ClimaEarth/setup_run.jl
@@ -496,16 +496,15 @@ function CoupledSimulation(config_dict::AbstractDict)
         The concrete steps for proper initialization are:
         =#
 
-        # 1.coupler updates surface model area fractions
+        # 1. Coupler updates surface model area fractions
         FieldExchanger.update_surface_fractions!(cs)
 
-        # 2.surface density (`ρ_sfc`): calculated by the coupler by adiabatically extrapolating atmospheric thermal state to the surface.
-        # For this, we need to import surface and atmospheric fields. The model sims are then updated with the new surface density.
+        # 2. Import atmospheric and surface fields into the coupler fields, then broadcast them back out to all components.
         FieldExchanger.import_combined_surface_fields!(cs.fields, cs.model_sims)
         FieldExchanger.import_atmos_fields!(cs.fields, cs.model_sims)
         FieldExchanger.update_model_sims!(cs.model_sims, cs.fields)
 
-        # Set all initial cache values for the land model, now that we have updated drivers
+        # 3. Set all initial cache values for the land model, now that we have updated drivers
         land_set_initial_cache! = CL.make_set_initial_cache(cs.model_sims.land_sim.model)
         land_set_initial_cache!(
             cs.model_sims.land_sim.integrator.p,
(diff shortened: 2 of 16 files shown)


experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -335,16 +333,27 @@ end
...
     # a reasonable stand-in.
-    ρ_sfc = FluxCalculator.extrapolate_ρ_to_sfc.(thermo_params, thermo_state_int, T_sfc) # ideally the # calculate elsewhere, here just getter...
-    q_sfc = Interfacer.get_field(sim, Val(:surface_humidity)) # already calculated in rhs! (cache)
-    @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, q_sfc)
+    ρ_sfc =
+        FluxCalculator.extrapolate_ρ_to_sfc.(
+            thermo_params,
+            Interfacer.get_field(atmos_sim, Val(:thermo_state_int)),
+            T_sfc,
+        ) # ideally the # calculate elsewhere, here just getter...
```
Sbozzolo, line 355: I don't understand this comment

experiments/ClimaEarth/setup_run.jl
```diff
@@ -496,46 +496,33 @@ function CoupledSimulation(config_dict::AbstractDict)
...
 
-        # 3.surface vapor specific humidity (`q_sfc`): step surface models with the new surface density to calculate their respective `q_sfc` internally
-        ## TODO: the q_sfc calculation follows the design of the bucket q_sfc, but it would be neater to abstract this from step! (#331)
-        Interfacer.step!(land_sim, tspan[1] + Δt_cpl)
-        Interfacer.step!(ocean_sim, tspan[1] + Δt_cpl)
-        Interfacer.step!(ice_sim, tspan[1] + Δt_cpl)
-
-        # 4.turbulent fluxes: now we have all information needed for calculating the initial
-        # turbulent surface fluxes
+        # 4. Compute radiative fluxes and update the coupler fields and model simulations with the new fluxes
```
Sbozzolo, line 515: Maybe let's mention that one should run all the other callbacks that change p as well

src/FieldExchanger.jl
```diff
@@ -248,4 +250,32 @@ function combine_surfaces!(combined_field, sims, field_name)
     end
 end
 
+"""
+    compute_surface_humidity!(q_sfc, T_atmos, q_atmos, ρ_atmos, T_sfc, thermo_params)
+Computes the surface specific humidity based on the atmospheric state and surface temperature.
```
Sbozzolo, line 255:
```suggestion

Compute the surface specific humidity based on the atmospheric state and surface temperature.
```

src/FieldExchanger.jl
```diff
@@ -248,4 +250,32 @@ function combine_surfaces!(combined_field, sims, field_name)
...
+The phase of the surface is determined by the surface temperature, and the saturation
+specific humidity is computed accordingly.
+All fields should be on the exchange grid.
+# Arguments
+- `q_sfc`: [CC.Fields.Field] output field for surface specific humidity.
+- `T_atmos`: [CC.Fields.Field] atmospheric temperature. G
```
Sbozzolo, line 261:
```suggestion
- `T_atmos`: [CC.Fields.Field] atmospheric temperature. 
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -681,29 +668,25 @@ function step!(cs::CoupledSimulation)
...
+    FieldExchanger.import_atmos_fields!(cs.fields, cs.model_sims)
+    ## update the model simulations with the new coupler fields
+    FieldExchanger.update_model_sims!(cs.model_sims, cs.fields)
+
+    ## calculate turbulent fluxes in the coupler and update the model simulations with them
+    FluxCalculator.turbulent_fluxes!(cs.model_sims, cs.fields, cs.boundary_space, cs.thermo_params)
 
+    ## update water albedo from wind at dt_water_albedo
+    ## (this will be extended to a radiation callback from the coupler)
+    TimeManager.maybe_trigger_callback(cs.callbacks.water_albedo, cs)
```
Sbozzolo, line 689: Why is this needed now?

Sbozzolo (commented):
Can you also add a NEWS entry?

test/flux_calculator_tests.jl
```diff
@@ -236,7 +243,8 @@ for FT in (Float32, Float64)
         atmos_sim = TestAtmos((; FT = FT), [], (; T = _ones .* FT(300), ρ = _ones .* FT(1.2), q = _ones .* FT(0.01)))
         thermo_params = get_thermo_params(atmos_sim)
         thermo_state_int = Interfacer.get_field(atmos_sim, Val(:thermo_state_int))
-        @test FluxCalculator.surface_thermo_state(surface_sim, thermo_params, thermo_state_int).ρ == thermo_state_int.ρ
+        ρ_expected = FluxCalculator.extrapolate_ρ_to_sfc.(thermo_params, thermo_state_int, surface_sim.integrator.T)
+        @test FluxCalculator.surface_thermo_state(surface_sim, thermo_params, atmos_sim).ρ == thermo_state_int.ρ
```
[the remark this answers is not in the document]
juliasloan25: Good point

experiments/ClimaEarth/components/land/climaland_bucket.jl
juliasloan25: I'm not sure either. I removed it

experiments/ClimaEarth/setup_run.jl
juliasloan25: Before, it was called at the beginning of each step. I moved it here so it's called alongside the checkpoint callback. I don't think it matters when during a step we update the albedo since we update it so often anyway

Sbozzolo: ok!

experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -335,16 +333,27 @@ end
...
-    ρ_sfc = FluxCalculator.extrapolate_ρ_to_sfc.(thermo_params, thermo_state_int, T_sfc) # ideally the # calculate elsewhere, here just getter...
-    q_sfc = Interfacer.get_field(sim, Val(:surface_humidity)) # already calculated in rhs! (cache)
-    @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, q_sfc)
+    ρ_sfc =
+        FluxCalculator.extrapolate_ρ_to_sfc.(
+            thermo_params,
+            Interfacer.get_field(atmos_sim, Val(:thermo_state_int)),
+            T_sfc,
+        )
+    @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, sim.integrator.p.bucket.q_sfc)
```
Sbozzolo, line 356:
```suggestion
    @. TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, sim.integrator.p.bucket.q_sfc)
    return nothing
```

Sbozzolo (approved):

Sbozzolo:
I guess we can also remove `reinit` once this is merged

juliasloan25:
This PR is behavior changing, but the longrun stability hasn't worsened, and the plots in regular CI look comparable to main, so I'll go ahead and merge

juliasloan25:
> I guess we can also remove `reinit` once this is merged

Good point, I opened #1317 for that

PR #1310, opened 2025-04-28
Change Base.show to print the type name

Sbozzolo:
While working interactively , it is often useful to know what type a given `CoupledModelSimulation`. This commit changes the default `Base.show` method to include the type short name, so that different objects can be more easily distinguished.


juliasloan25 (approved):
Right now we have a function `Interfacer.name` which basically does the same thing, but we can remove that and replace it with `nameof`.

E.g. `Interfacer.name(::BucketSimulation) = "BucketSimulation"`

Sbozzolo:
> Right now we have a function `Interfacer.name` which basically does the same thing, but we can remove that and replace it with `nameof`.
> 
> E.g. `Interfacer.name(::BucketSimulation) = "BucketSimulation"`



> Right now we have a function `Interfacer.name` which basically does the same thing, but we can remove that and replace it with `nameof`.
> 
> E.g. `Interfacer.name(::BucketSimulation) = "BucketSimulation"`

I reimplemented `Interfacer.name` to use `nameof(typeof))`

juliasloan25:
> > Right now we have a function `Interfacer.name` which basically does the same thing, but we can remove that and replace it with `nameof`.
> > E.g. `Interfacer.name(::BucketSimulation) = "BucketSimulation"`
> 
> I reimplemented `Interfacer.name` to use `nameof(typeof))`

I think it would be simpler to directly use `nameof` when we need it, instead of having our own wrapper that doesn't add any functionality

Sbozzolo:
> > > Right now we have a function `Interfacer.name` which basically does the same thing, but we can remove that and replace it with `nameof`.
> > > E.g. `Interfacer.name(::BucketSimulation) = "BucketSimulation"`
> > 
> > 
> > I reimplemented `Interfacer.name` to use `nameof(typeof))`
> 
> I think it would be simpler to directly use `nameof` when we need it, instead of having our own wrapper that doesn't add any functionality

Sounds good!

PR #1312, opened 2025-04-29
Fix insolation update

Sbozzolo:
Switching to PartitionedStateFluxes introduced a bug where insolation was not updated. This is because the correct `update_sim!` method was not being called.

This commit fixes that. In parallel, I also fixed the update of insolation in ClimaAtmos so that it is automatically updated every time RRTGMP is called.

https://github.com/CliMA/ClimaAtmos.jl/pull/3788

As I discuss in that issue, I think the only reason for updating the insolation tuple in ClimaCoupler is to do it more often that the radiation timestep. I think this is because the `cos_zenith_angle` is one of the exchanged field. So, if we remove `cos_zenith_angle` from the exchange fields, we can remove this specialized method in `climaatmos.jl` and use the default one in `FieldExchanger`.

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -398,7 +398,9 @@ function Interfacer.add_coupler_fields!(coupler_field_names, atmos_sim::ClimaAtm
     push!(coupler_field_names, atmos_coupler_fields...)
 end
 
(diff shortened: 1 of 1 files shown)


imreddyTeja (approved):
Will this cause `set_insolation_variables!` to be called twice (once from the FieldExchanger and once in the atmos callback)?

Sbozzolo:
> Will this cause `set_insolation_variables!` to be called twice (once from the FieldExchanger and once in the atmos callback)?

Yes, but the function is idempotent, so there's issue with that (and again, the goal would be to remove this call from here eventually)

PR #1319, opened 2025-04-30
Add `subsurface_face` to `ClimaLandSimulation` domain

imreddyTeja:
## Purpose 
This [PR](https://github.com/CliMA/ClimaLand.jl/pull/1121#issuecomment-2831436988) in ClimaLand introduces
a new domain, which needs to be added to the ClimaLandSimulation.

## Content
- adds `subsurface_face` to `ClimaLandSimulation` domains

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_helpers.jl
+++ b/experiments/ClimaEarth/components/land/climaland_helpers.jl
@@ -51,7 +51,8 @@ function make_land_domain(
     verttopology = CC.Topologies.IntervalTopology(vertmesh)
     vert_center_space = CC.Spaces.CenterFiniteDifferenceSpace(verttopology)
     subsurface_space = CC.Spaces.ExtrudedFiniteDifferenceSpace(atmos_boundary_space, vert_center_space)
-    space = (; surface = atmos_boundary_space, subsurface = subsurface_space)
+    subsurface_face_space = CC.Spaces.face_space(subsurface_space)
+    space = (; surface = atmos_boundary_space, subsurface = subsurface_space, subsurface_face = subsurface_face_space)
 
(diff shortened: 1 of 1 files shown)


experiments/ClimaEarth/components/land/climaland_helpers.jl
```diff
@@ -51,7 +51,8 @@ function make_land_domain(
     verttopology = CC.Topologies.IntervalTopology(vertmesh)
     vert_center_space = CC.Spaces.CenterFiniteDifferenceSpace(verttopology)
     subsurface_space = CC.Spaces.ExtrudedFiniteDifferenceSpace(atmos_boundary_space, vert_center_space)
-    space = (; surface = atmos_boundary_space, subsurface = subsurface_space)
+    subsurface_face_space = CC.Spaces.face_space(subsurface_space)
```
juliasloan25, line 54:
```suggestion
    subsurface_face_space = CL.Domains.obtain_face_space(subsurface_space)
```
In land they use `obtain_face_space` - we should probably do the same to be consistent

imreddyTeja: I opened a [PR](https://github.com/CliMA/ClimaLand.jl/pull/1134#pullrequestreview-2809989443) in ClimaLand that removes `obtain_face_space` and replaces it with `face_space`. I'm fine with keeping `obtain_face_space` if you see a reason to.

juliasloan25: Oh perfect! No need to keep that function in ClimaLand if it exists in ClimaCore

juliasloan25 (approved):
Thank you!

PR #1323, opened 2025-05-01
Remove unnecessary update_surface_fractions call

Sbozzolo:
At the very beginning our a simulation, `update_surface_conditions!` should do nothing, because the surface fractions were already initialized at the correct time. This commir removes that call.

This commit also adds a second atmos callback that modifies the cache.


juliasloan25 (approved):
Thank you! From buildkite, it also looks like there are no behavioral changes from this (as expected)

PR #1324, opened 2025-05-01
add `exchange!` function

### LINKED ISSUE #1320: combine coupler field imports and model sim update into one `exchange!` function
`import_combined_surface_fields!`, `import_atmos_fields!`, and `update_model_sims!` should always be called together in that order. We should make a function `exchange!` that calls these 3 together. We can also combine `import_combined_surface_fields!` and `import_atmos_fields!` into a function `update_coupler_fields!`.

juliasloan25:
## Purpose 
Wrap the following 3 functions into one function `exchange!`:
- import atmos fields into coupler fields
- import surface fields into coupler fields
- update component models with coupler fields

closes #1320

### DIFF
--- a/experiments/ClimaEarth/setup_run.jl
+++ b/experiments/ClimaEarth/setup_run.jl
@@ -498,9 +498,7 @@ function CoupledSimulation(config_dict::AbstractDict)
         FieldExchanger.update_surface_fractions!(cs)
 
         # 2. Import atmospheric and surface fields into the coupler fields, then broadcast them back out to all components.
-        FieldExchanger.import_combined_surface_fields!(cs.fields, cs.model_sims)
-        FieldExchanger.import_atmos_fields!(cs.fields, cs.model_sims)
-        FieldExchanger.update_model_sims!(cs.model_sims, cs.fields)
+        FieldExchanger.exchange!(cs)
 
         # 3. Set all initial cache values for the land model, now that we have updated drivers
         land_set_initial_cache! = CL.make_set_initial_cache(cs.model_sims.land_sim.model)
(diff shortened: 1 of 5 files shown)


src/FieldExchanger.jl
```diff
@@ -258,5 +263,26 @@ function compute_surface_humidity!(q_sfc, T_atmos, q_atmos, ρ_atmos, T_sfc, the
     return nothing
 end
 
+"""
+    exchange!(cs::Interfacer.CoupledSimulation)
+
+Exchange fields between the surface and atmosphere models.
+This is done in 2 steps:
+1. Import the atmosphere fields and surface fields into the coupler.
+3. Update the component model simulations with the coupler fields.
```
Sbozzolo, line 272:
```suggestion
2. Update the component model simulations with the coupler fields.
```

experiments/ClimaEarth/setup_run.jl
```diff
@@ -673,12 +671,8 @@ function step!(cs::CoupledSimulation)
     ## update the surface fractions for surface models
     FieldExchanger.update_surface_fractions!(cs)
 
-    ## update the coupler with the new surface properties
-    FieldExchanger.import_combined_surface_fields!(cs.fields, cs.model_sims)
-    ## update the coupler with the new atmospheric properties
-    FieldExchanger.import_atmos_fields!(cs.fields, cs.model_sims)
-    ## update the model simulations with the new coupler fields
-    FieldExchanger.update_model_sims!(cs.model_sims, cs.fields)
```
Sbozzolo, line 681: Let's also maybe remove these functions from the exported list and public APIs (if they are never called individually)

juliasloan25: In the initial exchange in `setup_run.jl`, we still call `import_atmos_fields!` + `update_model_sims!` individually right after the radiation callback, to update the coupler fields and surface models with variables set in the atmos callback. Since the surface models haven't changed, there's no need to call `import_combined_surface_fields!` (unless maybe we want to recompute surface humidity)

Sbozzolo: If we can substitute that with `exchange!`, I'd do it for the sake of simplicy, even if it is less efficient (and performance doesn't as much at init)

test/field_exchanger_tests.jl
```diff
@@ -308,4 +312,72 @@ for FT in (Float32, Float64)
...
+        coupler_field_names =
+            (:surface_direct_albedo, :surface_diffuse_albedo, :P_liq, :P_snow, :ρ_sfc, :T_sfc, :F_radiative, :temp1)
+        coupler_fields = NamedTuple{coupler_field_names}(
+            ntuple(i -> CC.Fields.ones(boundary_space) .* FT(0.5), length(coupler_field_names)),
+        )
+
+        # model cache setup
+        atmos_names =
+            (:albedo_direct, :albedo_diffuse, :liquid_precipitation, :snow_precipitation, :radiative_energy_flux_sfc)
+        atmos_fields = NamedTuple{atmos_names}(ntuple(i -> CC.Fields.ones(boundary_space), length(atmos_names)))
```
Sbozzolo, line 330: This fails because you are constructing a NamedTuple of Fields instead of a Field of NamedTuples.

experiments/ClimaEarth/setup_run.jl
juliasloan25: okay! I'll use `exchange!` there too and remove the other functions from API docs

Sbozzolo (approved):

PR #1327, opened 2025-05-02
Change signature for `turbulent_fluxes!`

### LINKED ISSUE #1295: use mutating-function syntax for `turbulent_fluxes!`
The function `turbulent_fluxes!` (formerly `partitioned_turbulent_fluxes!`) alters its input `csf`, but this is the second argument to the function. `csf` should be the first argument to follow convention

Also `get_surface_fluxes!` uses the mutating function syntax, but doesn't actually alter its inputs.

Sbozzolo:
`FluxCalculator.turbulent_fluxes!` can now be called in two ways:
```julia
turbulent_fluxes!(coupled_simulation::CoupledSimulation)
turbulent_fluxes!(coupler_fields, model_sims, thermo_params)
```
The previous signature was
```julia
turbulent_fluxes!(model_sims, coupler_fields, boundary_space, thermo_params)
```
The new signature simplifies calling the function, ensures that the mutating convention is respected (`turbulent_fluxes!` mutates `coupler_fields`), and removes unnecessary arguments and type restrictions.

Closes #1295

### DIFF
--- a/src/FluxCalculator.jl
+++ b/src/FluxCalculator.jl
@@ -29,11 +29,13 @@ function calculate_surface_air_density(atmos_sim::Interfacer.AtmosModelSimulatio
     error("this function is required to be dispatched on $(nameof(atmos_sim)), but no method defined")
 end
 
+function turbulent_fluxes!(cs::Interfacer.CoupledSimulation)
+    return turbulent_fluxes!(cs.fields, cs.model_sims, cs.thermo_params)
+end
+
 """
-    turbulent_fluxes!(model_sims::NamedTuple,
-                                  csf::CC.Fields.Field,
-                                  boundary_space::CC.Spaces.AbstractSpace,
-                                  thermo_params::TD.Parameters.ThermodynamicsParameters)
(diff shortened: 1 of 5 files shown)


src/FluxCalculator.jl
```diff
@@ -53,12 +54,8 @@ TODO:
 
 (NB: Radiation surface fluxes are calculated by the atmosphere.)
 """
-function turbulent_fluxes!(
-    model_sims::NamedTuple,
-    csf::CC.Fields.Field,
-    boundary_space::CC.Spaces.AbstractSpace,
-    thermo_params::TD.Parameters.ThermodynamicsParameters,
-)
+function turbulent_fluxes!(csf, model_sims, thermo_params)
```
imreddyTeja, line 57: Not important, but `csf` is referred to as `fields` in the doc string.

imreddyTeja (approved):
This does not close the second part of #1295, which is about `get_surface_fluxes!`, so maybe a new issue should be made for that

juliasloan25 (approved):

PR #1330, opened 2025-05-02
Tr/remove zenith angle

imreddyTeja:
## Purpose 
This PR removes zenith angle exchange from the coupler. This follows the change in this ClimaLand [PR](https://github.com/CliMA/ClimaLand.jl/pull/1135). The `CoupledRadiativeFluxes` driver now takes a start date, latitudes, and longitudes as an argument, and the driver updates cosine zenith during its update function.

Removing the zenith exchange requires bumping the lowest compatible ClimaLand version. Alternatively, we could check what version of ClimaLand is being used, and update cosine zenith in the coupler with older versions of ClimaLand

## Content
removes cos_zenith from the interfacer

----


juliasloan25 (approved):
Since supporting backwards compatibility would require keeping a large chunk of code that can otherwise be removed, and the current version of the land model is very stable, I think it's okay to drop backwards compatibility here and require ClimaLand v0.16.2+.

Could you add another commit to this PR that removes the 2 checks `pkgversion(CL) < vX` that we still have in the code? Since we're dropping that compatibility we can remove them, but it's still good to keep as a separate commit so we can easily revert it in the future.

Also, we should visually compare the results of the AMIP + integrated land runs to the previous main before merging to make sure this PR doesn't significantly change the physics.

PR #1335, opened 2025-05-06
Replace zero multiply op with `fill!`

akshaysridhar:
This should fix the issue in the current AMIP runs where we see `csf.L_MO` == `NaN` despite the solver output not having NaNs. (Inf is a valid output for L_MO, but the zero multiplication leads to issues).

### DIFF
--- a/src/FluxCalculator.jl
+++ b/src/FluxCalculator.jl
@@ -61,18 +61,22 @@ function turbulent_fluxes!(csf, model_sims, thermo_params)
 
     # Reset the coupler fields will compute. We need to do this because we will compute
     # area-weighted averages
-    csf.F_turb_ρτxz .*= FT(0)
-    csf.F_turb_ρτyz .*= FT(0)
-    csf.F_lh .*= FT(0)
-    csf.F_sh .*= FT(0)
-    csf.F_turb_moisture .*= FT(0)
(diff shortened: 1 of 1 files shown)


akshaysridhar:
Why not just `fill!(csf, 0)` (I'll check what other fields `csf` contains).

src/FluxCalculator.jl
```diff
@@ -61,18 +61,18 @@ function turbulent_fluxes!(csf, model_sims, thermo_params)
...
+    fill!(csf.F_lh, FT(0)) 
+    fill!(csf.F_sh, FT(0)) 
+    fill!(csf.F_turb_moisture, FT(0))
+    fill!(csf.z0m_sfc, FT(0))
+    fill!(csf.z0b_sfc, FT(0)) 
+    fill!(csf.beta, FT(0)) 
+    fill!(csf.q_sfc, FT(0)) 
+    fill!(csf.L_MO, FT(0)) 
+    fill!(csf.ustar, FT(0)) 
+    fill!(csf.buoyancy_flux, FT(0))
```
Sbozzolo, line 75: You could write
```suggestion
for p in (:F_turb_.....)
fill!(getproperty(csf, p), 0)
end
```
and maybe 0 works (instead of FT(0))

Sbozzolo (approved):

PR #1336, opened 2025-05-06
Add parameter_files arg to ClimaLandSimulation

imreddyTeja:
## Purpose 

The ClimaLandSimulation constructor currently creates the land parameters using the default dict. 

## Content
This PR replicates the functionality of the bucket model setup by adding an argument for parameters_files, which are merged with overriding into the default parameters toml. 

----


juliasloan25 (approved):
Thank you!

PR #1342, opened 2025-05-09
Remove uneeded usage of parent in tests

### LINKED ISSUE #1322: Remove parent from tests
We have tests like:
```
        @test Array(parent(model_sims.atmos_sim.cache.albedo_direct))[1] == results[2]
        @test Array(parent(model_sims.atmos_sim.cache.albedo_diffuse))[1] == results[3]

        # test variables without updates
        @test Array(parent(model_sims.atmos_sim.cache.surface_temperature))[1] == results[1]
        @test Array(parent(model_sims.atmos_sim.cache.beta))[1] == results[1]
        @test Array(parent(model_sims.atmos_sim.cache.roughness_buoyancy))[1] == results[1]
```

In these tests, all the values are identical. So, there's no need to use `parent` when we can compare directly the field.

imreddyTeja:
## Purpose 
closes #1322 

## To-do
`parent` is still used in the following test:

```
        land_fraction = CC.Fields.ones(test_space)
        dims = size(parent(land_fraction))
        m = dims[1]
        n = dims[2]
        parent(land_fraction)[1:(m ÷ 2), :, :, :] .= FT(0)

        # Construct ice fraction of 0s on left, 0.5s on right
        ice_d = CC.Fields.zeros(test_space)
        parent(ice_d)[:, (n ÷ 2 + 1):n, :, :] .= FT(0.5)
```

It could probably be replaced with `CC.Fields.field2array`, but I was not sure
if the top half/ bottom half and left/right relationships are important for the test. 

## Content
Replace `parent` usage with field comparisons in tests.

----

### DIFF
--- a/test/field_exchanger_tests.jl
+++ b/test/field_exchanger_tests.jl
@@ -121,18 +121,16 @@ end
 for FT in (Float32, Float64)
     @testset "test update_surface_fractions!" begin
         test_space = CC.CommonSpaces.CubedSphereSpace(FT; radius = FT(6371e3), n_quad_points = 4, h_elem = 4)
-        # Construct land fraction of 0s in top half, 1s in bottom half
+        # Construct land fraction of 0s in first half, 1s in second half
         land_fraction = CC.Fields.ones(test_space)
-        dims = size(parent(land_fraction))
-        m = dims[1]
-        n = dims[2]
-        parent(land_fraction)[1:(m ÷ 2), :, :, :] .= FT(0)
+        nelements = length(CC.Fields.field2array(land_fraction))
+        CC.Fields.field2array(land_fraction)[1:(nelements ÷ 2)] .= FT(0)
 
-        # Construct ice fraction of 0s on left, 0.5s on right
+        # Construct ice fraction of 0.5s on first half, 0s on second half
         ice_d = CC.Fields.zeros(test_space)
-        parent(ice_d)[:, (n ÷ 2 + 1):n, :, :] .= FT(0.5)
+        CC.Fields.field2array(ice_d)[1:(nelements ÷ 2)] .= FT(0.5)
 
-        # Construct ice fraction of 0s on left, 0.5s on right
+        # Construct ice fraction of 0s. During the update, the first half should be set to 0.5
         ocean_d = CC.Fields.zeros(test_space)
 
         # Fill in only the necessary parts of the simulation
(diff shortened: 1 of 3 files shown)


test/field_exchanger_tests.jl
```diff
@@ -161,7 +161,7 @@ for FT in (Float32, Float64)
         # Test that sum of fractions is 1 everywhere
         ice_fraction = Interfacer.get_field(cs.model_sims.ice_sim, Val(:area_fraction))
         ocean_fraction = Interfacer.get_field(cs.model_sims.ocean_sim, Val(:area_fraction))
-        @test all(parent(ice_fraction .+ ocean_fraction .+ land_fraction) .== FT(1))
+        @test ice_fraction .+ ocean_fraction .+ land_fraction == fill(FT(1), test_space)
```
Sbozzolo, line 164:
```suggestion
        @test ice_fraction .+ ocean_fraction .+ land_fraction == ones(test_space)
```

Sbozzolo (approved):

test/field_exchanger_tests.jl
```diff
@@ -215,9 +215,10 @@ for FT in (Float32, Float64)
             (; atmos_sim = DummySimulation(component_fields), land_sim = TestSurfaceSimulation1(component_fields))
 
         FieldExchanger.import_atmos_fields!(coupler_fields, model_sims)
-        @test Array(parent(coupler_fields.F_lh))[1] == FT(0)
-        @test Array(parent(coupler_fields.F_sh))[1] == FT(0)
-        @test Array(parent(coupler_fields.F_turb_moisture))[1] == FT(0)
+        zero_field = fill(FT(0), boundary_space)
```
juliasloan25, line 218:
```suggestion
        zero_field = zeros(boundary_space)
```

juliasloan25 (approved):

juliasloan25:
> `parent` is still used in the following test:
> 
> ```
>         land_fraction = CC.Fields.ones(test_space)
>         dims = size(parent(land_fraction))
>         m = dims[1]
>         n = dims[2]
>         parent(land_fraction)[1:(m ÷ 2), :, :, :] .= FT(0)
> 
>         # Construct ice fraction of 0s on left, 0.5s on right
>         ice_d = CC.Fields.zeros(test_space)
>         parent(ice_d)[:, (n ÷ 2 + 1):n, :, :] .= FT(0.5)
> ```
> 
> It could probably be replaced with `CC.Fields.field2array`, but I was not sure if the top half/ bottom half and left/right relationships are important for the test.

The left/right partition isn't super necessary, we just wanted to test that if we use an area fraction with 1s in some places and 0 in others for one model, then the inverse for another model, they'll correctly sum to 1 at each point. If you can easily switch to use `field2array` that would be nice, but I don't think it's worth spending a lot of time on.

PR #1343, opened 2025-05-09
Update leaderboard code for use in calibration

ph-kev:
The leaderboard code assumes that all the variables defined in `get_sim_var_dict()` are in the diagnostics, but this isn't always the case, especially with calibration. Hence, this PR only adds the variables if they exist in the diagnostics. Furthermore, this PR clean up the leaderboard code a bit using the new features in ClimaAnalysis.

### DIFF
--- a/experiments/ClimaEarth/leaderboard/leaderboard.jl
+++ b/experiments/ClimaEarth/leaderboard/leaderboard.jl
@@ -6,13 +6,14 @@ import Dates
 include("data_sources.jl")
 
 """
-    compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
+    compute_leaderboard(leaderboard_base_path, diagnostics_folder_path, spinup)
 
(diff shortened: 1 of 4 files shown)


experiments/ClimaEarth/leaderboard/leaderboard.jl
```diff
@@ -136,6 +142,7 @@ function compute_leaderboard(leaderboard_base_path, diagnostics_folder_path)
     end
 
     # Add RMSE for the CliMA model and for each season
+    rmse_var_dict = filter(((k, _),) -> k in keys(sim_var_dict), rmse_var_dict)
```
Sbozzolo, line 145: This takes a minute to process. Could you maybe move the anonymous function outside of the filter and give it a name so that the intent is a little bit more clear?

Sbozzolo (approved):

PR #1345, opened 2025-05-09
fix up longruns

### LINKED ISSUE #1315: fix longruns
Almost all of the longruns are failing for various reasons, including:
- out of memory (GPU AMIP FINE new target amip: topo)
- time limit (anything that runs for 1 day walltime)
- `ERROR: LoadError: ArgumentError: invalid index: "short_name" of type String` (GPU AMIP SUPERFINE: dyamond_target)

None of these involve stability, and we should be able to fix them relatively easily.

juliasloan25:
## Purpose 
closes #1315

example of failing longrun build before this PR: https://buildkite.com/clima/climacoupler-longruns/builds/867#01969a50-35a6-4700-be16-1a292a97b6b2

After this PR, the problems from all failing runs should be addressed.

## Content
- [x] `Slabplanet_aqua: nocouple` - removed run; at this point in time I don't think running component models without coupling is very useful to us. We can always add this back if we need information about an uncoupled run.
- [x] `ClimaAtmos standalone target` - fixed path to config file, unified config file with ClimaAtmos equivalent
- [x] `TARGET IDEALIZED: new target aqua - fixed ocean T, nocouple, atmos flux calc` - decreased `t_end` 120 -> 90 days
- [x] `TARGET IDEALIZED: new target slab - evolving slab ocean T, bucket` - removed `vert_diff` option, which aligns this more closely with the fixed ocean equivalent
- [x] `MPI AMIP FINE: new target amip` - moved to run on GPU; it should be fast enough for the time limit now
- [x] `MPI AMIP FINE: new target amip: topo` - removed since we run the same setup on GPU
- [x] `MPI AMIP FINE: new target amip: topo + diagedmf` - removed since we run the same setup on GPU
- [x] `GPU AMIP FINE: new target amip: topo` - increased memory limit, changed to run on clima
- [x] `GPU AMIP SUPERFINE: dyamond target` - changed atmos/coupler diagnostics merging to handle case where coupler list is empty; tested constructing ClimaAtmosSimulation works now
- [x] all runs - don't softfail

----


experiments/ClimaEarth/setup_run.jl
```diff
@@ -97,16 +97,6 @@ function CoupledSimulation(config_file = joinpath(pkgdir(ClimaCoupler), "config/
 end
 
 function CoupledSimulation(config_dict::AbstractDict)
-    # Make a copy so that we don't modify the original input
-    config_dict = copy(config_dict)
```
Sbozzolo, line 101: We need this, otherwise we cannot run the same simulation twice in the same session (as we do in the restarts).

Sbozzolo: I think I had problems with `parse_component_dts!(config_dict)`, because it changes the dict

charleskawczynski: I think that @Sbozzolo is saying that, although it may work, it's not easy to reason about because `Dict`s are passed by reference, so removing the `copy` makes this a mutating method. If we did remove the `copy`, then we should change `CoupledSimulation` to `CoupledSimulation!`. I'm not exactly sure how this method mutates the dict, but we would want `CoupledSimulation!` to be idempotent for it to be used correctly in that way.

charleskawczynski: Keeping the `copy` makes it simpler in this way.

juliasloan25: In this PR I changed the `CoupledSimulation` function so it doesn't modify `config_dict`. Now, the config dict is created in `get_coupler_config_dict`, and is not modified outside of that function. So as long as we use the config dict output from `get_coupler_config_dict`, our runs will be reproducible.

Previously, the restart test was manually parsing the config dict, duplicating what's done in `get_coupler_config_dict`. Now it just calls that function directly, so on subsequent runs we can call `get_coupler_config_dict(config_file)` and reproduce the config dict that was used

experiments/ClimaEarth/setup_run.jl
```diff
@@ -134,6 +124,9 @@ function CoupledSimulation(config_dict::AbstractDict)
         parameter_files,
     ) = get_coupler_args(config_dict)
 
+    # Initialize communication context (do this first so all printing is only on root)
+    comms_ctx = Utilities.get_comms_context(config_dict)
```
juliasloan25, line 128: I'll move this up to be the first thing done in this function, so that any printing in the arg parsing is only on root

Sbozzolo (approved):

PR #1353, opened 2025-05-14
add ED only, diagnostic EDMF longruns

juliasloan25:
## Purpose 
Add longruns using ED only and diagnostic EDMF in the atmosphere, with both bucket and integrated land models.

build using previous ustar limiter (`max(momentum flux, gustiness)`) https://buildkite.com/clima/climacoupler-longruns/builds/874
build using current main ustar limiter (`ustar > eps(FT)`) https://buildkite.com/clima/climacoupler-longruns/builds/876

## Content

----


juliasloan25:
done in #1361

PR #1354, opened 2025-05-14
use one ocean remapper

juliasloan25:
## Purpose 
Use one ocean remapper instead of 3. Remap all fields onto `Center, Center` coordinates of the ocean grid, then use Oceananigans functions to remap them to the appropriate locations (`Face, Center`, `Center, Face`) as needed.

## Content
- [x] only use center/center remapper (remove other 2)
- [x] use `set_from_extrinsic_vectors!` to rotate and interpolate once on ocean grid (this requires filling halo regions, just for u and v)
- [x] don't fill halo regions for other fluxes, since it isn't needed
- [x] define `_rotate_vectors!`, `_interpolate_vectors!`
- [x] make sure `set_from_extrinsic_vectors!` is launching 2D kernels

----


experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -220,28 +201,27 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
...
 
-    view(oc_flux_u, :, :, 1) .= remapped_F_turb_ρτxz
-    view(oc_flux_v, :, :, 1) .= remapped_F_turb_ρτyz
+    # Set the momentum flux BCs at the correct locations
+    set_from_extrinsic_vectors!(
+        (; u = oc_flux_u, v = oc_flux_v),
+        sim.model.grid,
+        remapped_F_turb_ρτxz,
+        remapped_F_turb_ρτyz,
+    )
```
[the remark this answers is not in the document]
juliasloan25: `remapped_F_turb_ρτxz` and `remapped_F_turb_ρτyz` are already remapped onto the ocean grid.

You are right about the `sim.model.grid` access, thank you!

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -251,41 +222,71 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
...
+    _interpolate_velocities!(u, v, grid, u_cc, v_cc)
+
+Interpolate the input velocities `u_cc` and `v_cc`, which are Center/Center
+Fields to Face/Center and Center/Face coordinates, respectively.
+"""
+@kernel function _interpolate_velocities!(u, v, grid, u_cc, v_cc)
+    i, j = @index(Global, NTuple)
+    @inbounds begin
+        u[i, j] = OC.Operators.ℑxyᶠᶜᵃ(i, j, 1, grid, u_cc)
+        v[i, j] = OC.Operators.ℑxyᶜᶠᵃ(i, j, 1, grid, v_cc)
```
juliasloan25, line 287: Is this the right way to make this kernel 2D? (k=1)

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -251,41 +222,71 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
...
+Rotate the velocities from the extrinsic coordinate system to the intrinsic
+coordinate system.
+"""
+@kernel function _rotate_velocities!(u, v, grid)
+    i, j = @index(Global, NTuple)
+    # Rotate u, v from extrinsic to intrinsic coordinate system
+    ur, vr = OC.Operators.intrinsic_vector(i, j, 1, grid, u, v)
+    @inbounds begin
+        u[i, j] = ur
+        v[i, j] = vr
```
juliasloan25, line 273:
```suggestion
        u[i, j, 1] = ur
        v[i, j, 1] = vr
```

experiments/ClimaEarth/components/ocean/oceananigans.jl
juliasloan25:
```suggestion
        u[i, j, 1] = OC.Operators.ℑxyᶠᶜᵃ(i, j, 1, grid, u_cc)
        v[i, j, 1] = OC.Operators.ℑxyᶜᶠᵃ(i, j, 1, grid, v_cc)
```

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -218,30 +200,19 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
...
-    # onto Center, Center fields on the Oceananingans side, and then remap from ClimaCoupler to this
-    # See also comment below next to the function
-
-    # set_from_extrinsic_vectors!((; u = oc_flux_u, v = oc_flux_v),
-    #                             sim.model.grid,
-    #                             remapped_F_turb_ρτxz,
-    #                             remapped_F_turb_ρτyz
-    #                             )
+    # Remap momentum fluxes onto Center, Center, Center fields
+    remapped_F_turb_ρτxz = OC.Field{OC.Center, OC.Center, OC.Center}(grid)
```
juliasloan25, line 206:
```suggestion
    remapped_F_turb_ρτxz = OC.Field{OC.Center, OC.Center, Nothing}(grid)
```

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -218,30 +200,19 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
...
+    OC.set!(remapped_F_turb_ρτxz, CC.Remapping.interpolate(sim.remapping.remapper_cc, F_turb_ρτxz))
+    remapped_F_turb_ρτyz = OC.Field{OC.Center, OC.Center, OC.Center}(grid)
```
juliasloan25, line 208:
```suggestion
    remapped_F_turb_ρτyz = OC.Field{OC.Center, OC.Center, Nothing}(grid)
```
reduced Field

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -343,7 +346,7 @@ function FieldExchanger.update_sim!(sim::OceananigansSimulation, csf, area_fract
     oc_flux_S = surface_flux(sim.sim.model.tracers.S)
     precipitating_fresh_water_flux = (remapped_P_liq .+ remapped_P_snow) ./ ocean_fresh_water_density
     surface_salinity_flux = OC.interior(sim.sim.model.tracers.S, :, :, 1) .* precipitating_fresh_water_flux
-    view(oc_flux_S, :, :, 1) .= .-surface_salinity_flux
+    oc_flux_S .= .-surface_salinity_flux
```
glwagner, line 349:
```suggestion
    @. oc_flux_S = -surface_salinity_flux
```

glwagner (approved):
Looks good. In the longer term, it may be better to avoid broadcasting and write a kernel instead if possible. A kernel can be easier to read and involve simpler code that is easier to maintain / less bug prone, since you avoid the trickiness of many `.`

experiments/ClimaEarth/components/ocean/oceananigans.jl
juliasloan25: `oc_flux_S` is an Oceananigans Field. What would you recommend doing instead of broadcasting?

glwagner: Then unwrap:

```julia
interior(oc_flux_S, :, :, 1) .= .-surface_salinity_flux
```

juliasloan25:
> Looks good. In the longer term, it may be better to avoid broadcasting and write a single kernel / a few kernels for this entire operation instead if possible. A kernel can be easier to read and involve simpler code that is easier to maintain / less bug prone, since you avoid the trickiness of many "`.`". This is the approach generally used in ClimaOcean as well.

Okay! ClimaCore is optimized to be performant with broadcasting so that's typically the paradigm we use. Maybe we could meet sometime and you can show me how you would write some things as kernels instead?

glwagner:
For sure. You will also benefit from writing kernels with ClimaCore because it will allow you to fuse more operations than is typically possible with broadcasting alone.

experiments/ClimaEarth/components/ocean/oceananigans.jl
juliasloan25: Ahh okay, is this why we were using `view` before?

Abdulrahmansalg:
مرحبا

PR #1356, opened 2025-05-15
Use ECCO date for ocean initial conditions

Sbozzolo:
This commit changes the initial conditions for the ocean when the simulation is non-trivial (in terms of duration).

ECCO data requires authentication to log in. This is automatic when using ClimaModules.

### DIFF
--- a/experiments/ClimaEarth/components/ocean/oceananigans.jl
+++ b/experiments/ClimaEarth/components/ocean/oceananigans.jl
@@ -3,6 +3,7 @@ import ClimaOcean as CO
 import ClimaCoupler: Checkpointer, FieldExchanger, FluxCalculator, Interfacer, Utilities
 import ClimaComms
 import Thermodynamics as TD
+import ClimaOcean.ECCO: download_dataset
 
 """
     OceananigansSimulation{SIM, A, OPROP, REMAP}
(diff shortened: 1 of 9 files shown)


Sbozzolo:
Passing nightly build: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/359

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -62,14 +74,25 @@ function OceananigansSimulation(area_fraction; output_dir, comms_ctx = ClimaComm
 
     grid = OC.ImmersedBoundaryGrid(underlying_grid, OC.GridFittedBottom(bottom_height); active_cells_map = true)
 
-    # Create ocean simulation
-    ocean = CO.ocean_simulation(grid)
+    if use_ecco
```
juliasloan25, line 77: Could you add an inline comment explaining that if we use ECCO, we have forcings and don't need to set ICs, otherwise we do need to set ICs?

experiments/ClimaEarth/setup_run.jl
```diff
@@ -327,7 +327,9 @@ function CoupledSimulation(config_dict::AbstractDict)
         ocean_fraction = FT(1) .- ice_fraction .- land_fraction
 
         if sim_mode <: CMIPMode
-            ocean_sim = OceananigansSimulation(ocean_fraction; output_dir = ocean_output_dir, comms_ctx)
+            stop_date = date(tspan[end])
```
juliasloan25, line 330: `tspan[end] == t_end` is usually a period of time (e.g. `10days`). I believe taking the date like this assumes `t_start == 0`, and we instead want to do something like `start_date + tspan[end]`

juliasloan25 (approved):

experiments/ClimaEarth/setup_run.jl
Sbozzolo: Ah yeah, thank you!

PR #1359, opened 2025-05-21
Lower Pr_max

trontrytel:
Some of the nightly AMIP runs are not stable. Lowering maximum Prandtl number fixes it.


szy21 (approved):
I think that the random failure of amip jobs might be something else given they crash at the very top, but we can change this to see if it makes things better.

costachris (approved):

trontrytel:
> I think that the random failure of amip jobs might be something else given they crash at the very top, but we can change this to see if it makes things better.

👍 We will keep monitoring. If nightly amip keeps randomly failing I will investigate 🤞

PR #1364, opened 2025-05-23
shorten regular CI

juliasloan25:
## Purpose 
Our CI takes 1 hour, and it would be helpful to development time if we can shorten it.

The goals of our regular CI are to make sure PRs don't break functionality, and to have some results we can compare back to (e.g. to ensure results haven't changed if we don't want them to). For the first goal, running for 2 timesteps is sufficient, and for the second, running for 1 day should be sufficient (plots should be made with diagnostics from the first day).

Especially now that we have the nightly AMIP runs I think it's okay to have slightly less strict (and also less expensive) regular CI.

## To-do
- [x] decrease max simulation length in regular CI from 10 days to 1 day
- [x] call radiation once every 2dt instead of 1dt in the restart test

## Content

----


juliasloan25:
> The restart test is removed from `runtests.jl`, but I only see a buildkite step for the restart test with MPI and CUDA. Does this also need to be tested with a single threaded cpu device, as the ci GitHub workflow will no longer run it?
> 
> Other than that, this seems like a nice quality of life improvement.

I was thinking that if it runs with MPI and CUDA, it will likely be fine single threaded too. Since running single threaded on CPU will be slower I was thinking of leaving it out, but if you think it might cause problems down the line we can add it in too

PR #1369, opened 2025-05-26
use nq=4 for boundary space

### LINKED ISSUE #1365: make boundary_space independent of atmos sim when not being shared
Right now, the boundary space construction still accesses the `atmos_sim` space to get `n_quad_points`. It's very helpful to be able to create the space without creating the atmosphere sim, especially when debugging. We should just use a default value for `n_quad_points`, and if it's something users want to change we can expose it as an option.

juliasloan25:
## Purpose 
When we construct a boundary space _not_ based on the atmosphere space, it should be entirely independent of the atmosphere simulation/space. It's helpful to be able to create the space without creating the atmosphere sim, especially when debugging.

This PR changes the independent boundary space to use `n_quad_points = 4`. I chose this value because it's the default used by both ClimaCore and ClimaAtmos, and it's higher resolution than `n_quad_points = 0` used by the land. We can expose this as an option later on if users want to change it.

closes #1365


juliasloan25:
> Is the comment from line 215 to 228 still accurate? It looks like it may have been written before the `if share_surface_space` conditional was added.

Good point, it's out of date. I'll update it.

PR #1374, opened 2025-05-27
Extend diagnostics for Oceananigans sim

### LINKED ISSUE #1368: improve diagnostics for Oceananigans simulation
Right now, we only output a temperature diagnostic for the ocean simulation. We should add more things, potentially including:
- salinity, vertical velocity, eddy viscosity - see [this example](https://clima.github.io/OceananigansDocumentation/stable/literated/ocean_wind_mixing_and_convection/)
- vorticity, speed - see [this example](https://clima.github.io/OceananigansDocumentation/stable/literated/two_dimensional_turbulence/)
- diagnostics for global simulation - see [this example](https://clima.github.io/ClimaOceanDocumentation/stable/literated/near_global_ocean_simulation/#Initial-setup-with-package-imports)

juliasloan25:
## Purpose 
closes #1368

## To-do
- [x] use ClimaAnalysis in `make_ocean_diagnostics_plots` to unify with the other `make_diagnostics_plots` methods
  - I still kept this as a different function since we need to do some things differently because Oceananigans uses a different output saving infrastructure.
- [x] add new ocean diagnostics: S, u, v

## To Do (later)
- [ ] add animations for the ocean simulation and ideally atmosphere too
  - it wasn't clear to me how to do this in a straightforward way, so I left it for a separate PR

## Content
Example of new diagnostics from a simulation run for 2 timesteps locally:
Screenshot 2025-05-27 at 3 41 48 PM

----

### DIFF
--- a/experiments/ClimaEarth/components/ocean/oceananigans.jl
+++ b/experiments/ClimaEarth/components/ocean/oceananigans.jl
@@ -97,7 +97,6 @@ function OceananigansSimulation(area_fraction, start_date, stop_date; output_dir
         OC.set!(ocean.model, T = T_init, S = S_init)
     end
 
-    # TODO: Handle halos directly here instead of filling them later
     long_cc = OC.λnodes(grid, OC.Center(), OC.Center(), OC.Center())
     lat_cc = OC.φnodes(grid, OC.Center(), OC.Center(), OC.Center())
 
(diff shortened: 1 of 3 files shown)


experiments/ClimaEarth/user_io/diagnostics_plots.jl
```diff
@@ -200,23 +201,57 @@ function make_diagnostics_plots(output_path::AbstractString, plot_path::Abstract
...
+
+        vars[i] = output_var
+    end
+
+    # Filter out any variables that are not available
+    vars = filter(!isnothing, vars)
+    isempty(vars) && return nothing
+
+    # Make plots for each variable, saved in one PDF file
+    !isempty(vars) && make_plots_generic(
```
imreddyTeja, line 250: The `!isempty(vars)` will always be true here because of line 247. Was this intentionally left for code clarity?

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -97,7 +97,6 @@ function OceananigansSimulation(area_fraction, start_date, stop_date; output_dir
         OC.set!(ocean.model, T = T_init, S = S_init)
     end
 
-    # TODO: Handle halos directly here instead of filling them later
```
imreddyTeja, line 100: Was this done in a different PR?

experiments/ClimaEarth/user_io/diagnostics_plots.jl
juliasloan25: Oops, good catch!

experiments/ClimaEarth/components/ocean/oceananigans.jl
juliasloan25: Yeah it was fixed in a recent PR where I changed from 3 remappers to 1, but I forgot to remove this comment

PR #1376, opened 2025-05-28
reduce the frequency of amip checkpoint

szy21:
## Purpose 

The checkpoint frequency is too high right now and produces a lot of data which fills the scratch space very fast.

## Content

----

### DIFF
--- a/config/amip_configs/amip.yml
+++ b/config/amip_configs/amip.yml
@@ -1,7 +1,7 @@
 FLOAT_TYPE: "Float32"
 albedo_model: "CouplerAlbedo"
 atmos_config_file: "config/longrun_configs/amip_target_diagedmf.yml"
-checkpoint_dt: "720hours"
+checkpoint_dt: "12months"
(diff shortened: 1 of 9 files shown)


szy21:
Btw does `months` work for checkpoint frequency?

juliasloan25:
> Btw does `months` work for checkpoint frequency?

yeah, `nmonths` will work

juliasloan25 (approved):
thank you!

juliasloan25:
I just noticed the default for `checkpoint_dt` is 20 days. For most of the longruns we don't specify anything so we're using this default. Could you please change the default to something like "90days", and set it to "365days" in these configs:
- config/longrun_configs/amip_decaywithheight_integrated_land.yml
- config/longrun_configs/amip_decaywithheight.yml
- config/longrun_configs/amip_diagedmf_topo.yml (this one should be "300days")
- config/longrun_configs/amip_edonly_topo.yml

PR #1401, opened 2025-06-10
Enable restart using a new sim start time

nefrathenrici:
## Purpose 

This PR enables us to use restart files with a new simulation start time. This is needed to enable steady initial states for calibration.

Sample output:
```
[ Info: Restarting from time 31536000 and directory experiments/calibration/coarse_amip/output/model_config/checkpoints
┌ Warning: Time value differs in restart
│   field = ".dt.epoch"
│   original = 2001-09-01T00:00:00
│   new = 2002-09-01T00:00:00
└ @ Main ~/clima/ClimaCoupler.jl/experiments/ClimaEarth/components/shared/restore.jl:94
┌ Warning: Time value differs in restart
│   field = ".atmos.insolation.start_date"
│   original = 2001-09-01T00:00:00
│   new = 2002-09-01T00:00:00
└ @ Main ~/clima/ClimaCoupler.jl/experiments/ClimaEarth/components/shared/restore.jl:94
```

----

### DIFF
--- a/experiments/ClimaEarth/components/shared/restore.jl
+++ b/experiments/ClimaEarth/components/shared/restore.jl
@@ -10,6 +10,7 @@ import ClimaCore.Spaces: AbstractSpace
 import ClimaUtilities.TimeVaryingInputs: AbstractTimeVaryingInput
 import StaticArrays
 import NCDatasets
+import Dates
 
 """
     restore!(v1, v2, comms_ctx; ignore)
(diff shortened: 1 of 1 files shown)


experiments/ClimaEarth/components/shared/restore.jl
```diff
@@ -77,3 +78,24 @@ function restore!(v1::Dict, v2::Dict, comms_ctx; name, ignore)
...
+Special case for time-related types to allow different timestamps during restore.
+"""
+function restore!(
+    v1::T1,
+    v2::T2,
+    comms_ctx;
+    name,
+    ignore,
+) where {
+    T1 <: Union{Dates.DateTime, Dates.UTInstant, Dates.Millisecond},
```
imreddyTeja, line 94: Should this also include `ITime`?

imreddyTeja (approved):
Looks good. I left one question about using this with `ITime`

experiments/ClimaEarth/components/shared/restore.jl
nefrathenrici: I haven't had any issues when restarting `ITime` simulations so I think it's safe to exclude. I'd rather minimize the list of exceptions for restarts.

@ph-kev Do you have any thoughts?

PR #1402, opened 2025-06-11
Add check for RMSEs

ph-kev:
This PR adds a check for RMSEs, so that regression does not happen.

### DIFF
--- a/experiments/ClimaEarth/user_io/postprocessing.jl
+++ b/experiments/ClimaEarth/user_io/postprocessing.jl
@@ -34,14 +34,16 @@ function postprocess_sim(cs, postprocessing_vars)
     # If we have enough data (in time, but also enough variables), plot the leaderboard.
     # We need pressure to compute the leaderboard.
     simdir = CAN.SimDir(atmos_output_dir)
-    pressure_in_output = "pfull" in CAN.available_vars(simdir)
-    times = CAN.times(get(CAN.SimDir(atmos_output_dir), first(CAN.available_vars(simdir))))
-    t_end = times[end]
-    if t_end > 84600 * 31 * 3 # 3 months for spin up
-        leaderboard_base_path = artifact_dir
-        compute_leaderboard(leaderboard_base_path, atmos_output_dir, 3)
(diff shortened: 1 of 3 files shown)


experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -40,6 +41,7 @@ function postprocess_sim(cs, postprocessing_vars)
             leaderboard_base_path = artifact_dir
             compute_leaderboard(leaderboard_base_path, atmos_output_dir, 3)
             compute_pfull_leaderboard(leaderboard_base_path, atmos_output_dir, 6)
+            test_rmse_thresholds(atmos_output_dir, 3)
```
juliasloan25, line 44: does this function also require `pressure_in_output`, or can it be outside of that check?

experiments/ClimaEarth/user_io/postprocessing.jl
```diff
@@ -33,14 +34,13 @@ function postprocess_sim(cs, postprocessing_vars)
...
     pressure_in_output = "pfull" in CAN.available_vars(CAN.SimDir(atmos_output_dir))
-    if pressure_in_output
-        times = CAN.times(get(CAN.SimDir(atmos_output_dir), "pfull"))
-        t_end = times[end]
-        if t_end > 84600 * 31 * 3 # 3 months for spin up
-            leaderboard_base_path = artifact_dir
-            compute_leaderboard(leaderboard_base_path, atmos_output_dir, 3)
-            compute_pfull_leaderboard(leaderboard_base_path, atmos_output_dir, 6)
-        end
+    times = CAN.times(get(CAN.SimDir(atmos_output_dir), "pfull"))
```
juliasloan25, line 37: Is there another variable we can get `times` from? This assumes that "pfull" is an available var

ph-kev: I can change it to just get any variable from `SimDir`, but there is an assumption that the diagnostics is monthly. However, I think this part of the code assumes the diagnostics are monthly anyway.

PR #1404, opened 2025-06-17
standardize atmos configs

juliasloan25:
## Purpose 
All config files now use one of the following 2 config files: `"config/longrun_configs/amip_target_diagedmf.yml"` or `"config/longrun_configs/amip_target_edonly.yml"` , with the exception of benchmark runs which use atmos config files defined within ClimaCoupler.jl. I think this is fine since we look at those runs for performance more than for physics. The DYAMOND run is also unchanged because we expect to remove it this week. Some variables are overwritten, so we don't always have the same final atmosphere configuration. But we will now use a standardized set of TOML parameters, which should help with debugging.

Config files are also alphabetized in this PR (in a separate commit).


config/longrun_configs/longrun_aquaplanet_allsky_tvinsol_0M_slabocean.yml
```diff
@@ -1,3 +1,9 @@
-atmos_config_file: "config/longrun_configs/longrun_aquaplanet_allsky_tvinsol_0M_slabocean.yml"
+albedo_model: "RegressionFunctionAlbedo"
+approximate_linear_solve_iters: 2
+atmos_config_file: "config/longrun_configs/amip_target_edonly.yml"
```
akshaysridhar, line 3: If the intention here is to retain one aquaplanet longrun (based on the filename) we could revert the `atmos_config_file` to the original slabocean version.

config/longrun_configs/slabplanet_aqua_evolve_ocean.yml
```diff
@@ -1,4 +1,4 @@
-atmos_config_file: "config/longrun_configs/longrun_aquaplanet_allsky_0M.yml"
+atmos_config_file: "config/longrun_configs/amip_target_edonly.yml"
```
akshaysridhar, line 1: Same here as above (aqua_evolve_ocean might suggest the aquaplanet config file, and the corresponding tomls)

config/longrun_configs/longrun_aquaplanet_allsky_tvinsol_0M_slabocean.yml
juliasloan25: In the coupler, running aquaplanet is controlled by specifying `mode_name: "slabplanet_aqua"` (which I should have added here and will update). Is there a problem with using the "AMIP" atmos config, but setting up aquaplanet in the coupler via this config option?

akshaysridhar: Ah it uses `toml: [toml/longrun_aquaplanet.toml]` in the atmos model; I think in this case with `mode_name: slabplanet_aqua` we can leave it as is i.e. `atmos_config_file = config/longrun_configs/longrun_aquaplanet_allsky_tvinsol_0M_slabocean.yml` (both use ED-only, so `decaywithheightdiffusion` is the only other viable option which has a clear identifier elsewhere)

PR #1405, opened 2025-06-20
Window elevations to avoid pressure inversions

Sbozzolo:
In some configurations, we observed pressure inversions at low elevations. These are problematic because they prevent the conversion between pressure and altitudes. This commit remove the region of space that we empirically found being the culprit. If pressure inversions occur somewhere else, an error will sill be thrown.


juliasloan25 (approved):
Thank you!

PR #1406, opened 2025-06-23
Fix typo in leaderboard

ph-kev:
It should be `sim_var` and not `pfull_var`.

Here are the plots produced from the leaderboard.
bias_lat_pfull_heatmaps
bias_vars_in_pfull


ph-kev:
> Thank you! Will this fix the unit mismatch we saw in the most recent longruns?

Yes

PR #1409, opened 2025-06-24
construct TripolarGrid

### LINKED ISSUE #1382: use TripolarGrid with OceananigansSimulation
Right now, we use `LatitudeLongitudeGrid`, but it would be preferred to use the `TripolarGrid`. This is used in the ClimaOcean "one-degree ocean simulation" example, which we can use as a template (see [here](https://clima.github.io/ClimaOceanDocumentation/stable/literated/one_degree_simulation/)).

juliasloan25:
## Purpose 
Use a TripolarGrid instead of a capped lat/lon grid when constructing an Oceananigans simulation.

closes #1382

## To-do
- [x] construct TripolarGrid
- [ ] update interpolation objects for new grid
- [ ] rotate vectors during exchange


glwagner:
@juliasloan25 how do I run the coupled simulation ? Is there a way to run it outside the caltech cluster (eg on a GPU system we have at MIT)?

juliasloan25:
> @juliasloan25 how do I run the coupled simulation ? Is there a way to run it outside the caltech cluster (eg on a GPU system we have at MIT)?

Sorry I just saw this - I get a lot of Github notifications so they get missed often. We have a [docs page](https://github.com/CliMA/ClimaCoupler.jl/pull/1409#issuecomment-3014660680) on running the coupled simulation. It shows how to run interactively or from the command line. You should be able to run it locally or on a different cluster, and it'll pick up the device to run on automatically (i.e. if a GPU is available it will run on GPU). Let me know if there are problems.

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -95,17 +82,19 @@ function OceananigansSimulation(area_fraction, start_date, stop_date; output_dir
...
+    # Construct a remapper from the exchange grid to `Center, Center` fields
     long_cc = OC.λnodes(grid, OC.Center(), OC.Center(), OC.Center())
     lat_cc = OC.φnodes(grid, OC.Center(), OC.Center(), OC.Center())
 
-    # TODO: Go from 0 to Nx+1, Ny+1 (for halos) (for LatLongGrid)
-
     # Construct a remapper from the exchange grid to `Center, Center` fields
     long_cc = reshape(long_cc, length(long_cc), 1)
     lat_cc = reshape(lat_cc, 1, length(lat_cc))
     target_points_cc = @. CC.Geometry.LatLongPoint(lat_cc, long_cc)
```
simone-silvestri, line 92: Looks like the memory error caused by huge allocation is hit here. What is the command doing?

juliasloan25: This creates a Matrix of size length(lat_cc) x length(long_cc) containing a (lat, long) pair at each point

simone-silvestri: this is probably the issue, `lat_cc` and `long_cc` are 2D matrices for a tripolar grid. We can use those ones directly

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -108,96 +102,185 @@ function OceananigansSimulation(
...
+- `boundary_space`: The boundary space (ClimaCore SpectralElementSpace2D).
+
+Returns:
+- `remapper_oc_to_cc`: The remapper object to go from the Oceananigans grid to the Cubed sphere nodes.
+"""
+function construct_remappers(grid, boundary_space)
+    ## Remapper: Oceananigans `Center, Center` to Cubed sphere nodes
+    # Get the Oceananigans coordinates and put them on CPU
+    coords_oc =
+        OceananigansXESMFExt.xesmf_coordinates(grid, OC.Center(), OC.Center(), OC.Center())
```
juliasloan25:
```suggestion
        XESMF.xesmf_coordinates(grid, OC.Center(), OC.Center(), OC.Center())
```

PR #1411, opened 2025-06-26
some longrun fixes

juliasloan25:
## Purpose 
This PR does a few things to the longruns:
- Add aquaplanet setups testing the atmosphere with ED only and diagnostic EDMF setups
- Remove the atmos-only longrun, since it isn't the coupler's responsibility to run standalone atmosphere simulations.
- Increase the conservation limit, since recent slabplanet longruns have failed the conservation check. It's not clear when exactly the conservation worsened, since these runs have been failing for different reasons for 4 months. It should be independent of recent changes to regridding and the independent boundary space, since these runs still use the atmos horizontal space as the boundary space.

## Content

----


.buildkite/longruns/pipeline.yml
```diff
@@ -55,14 +55,7 @@ steps:
   - group: "Coupler integration and conservation tests"
 
     steps:
-      - label: "ClimaAtmos standalone: aquaplanet, allsky rad, TV insol, 0M"
```
akshaysridhar, line 58: Removing this is fine - this test has served its purpose through early ClimaCoupler development.

experiments/ClimaEarth/user_io/debug_plots.jl
```diff
@@ -75,7 +75,7 @@ function plot_global_conservation(
     if !softfail
         @info typeof(cc)
         @info rse[end]
-        @assert rse[end] < 0.0065
+        @assert rse[end] < 0.035
```
akshaysridhar, line 78: Do we need to make this threshold visible in the docs ? e.g. leaderboard / diagnostic output section?

juliasloan25: To me, this value is more for internal use as we develop the coupled model. E.g. I expect it will change as we change our regridding approach moving forward, so I'm not sure we need to broadcast it yet. But if you think it would be useful, I can add it to the docs too

akshaysridhar: Yeah since this is just in `debug_plots` we perhaps don't need to include this in docs.

akshaysridhar:
LGTM to rebase + merge. Thanks @juliasloan25

PR #1413, opened 2025-06-26
use nonequil moisture with 1M

juliasloan25:
## Purpose 
Add `moist: "nonequil"` to the 1M microphysics longrun to make a compatible atmosphere setup. This should fix the error we saw in the [most recent longrun](https://buildkite.com/clima/climacoupler-longruns/builds/904#019796a7-d8df-4a7b-a6f3-73bc921680ed/171-270).


trontrytel (approved):
Thank you!

PR #1432, opened 2025-07-10
use amip.toml for all longruns

juliasloan25:
## Purpose 
Switches 3 remaining longrun configs to use `amip.toml`, which is used by our 3-day AMIP longrun and is the toml we want to use going forward. Also removes ` toml/amip_target_topo_diagedmf.toml` since it's no longer used.


akshaysridhar (approved):
Thanks @juliasloan25. We can monitor the nightly / weekly runs following + the CMIP fixes to see if these are very sensitive to resolution.

PR #1437, opened 2025-07-15
remove bucket get_new_cache

### LINKED ISSUE #1261: remove `get_new_cache` function for Bucket

juliasloan25:
## Purpose 
Remove the bucket `get_new_cache` function, and instead access the cached fields for total energy and water directly, which were added in https://github.com/CliMA/ClimaLand.jl/pull/1104.

closes #1261


experiments/ClimaEarth/components/land/climaland_bucket.jl
```diff
@@ -262,30 +245,24 @@ Interfacer.get_field(sim::BucketSimulation, ::Val{:surface_temperature}) =
...
+the snow water equivalent `σS`, and surface water content `Ws`.
+
+This method is required by the ConservationChecker to check water conservation.
 """
-function Interfacer.get_field(sim::BucketSimulation, ::Val{:water})
-    ρ_cloud_liq = CL.LP.ρ_cloud_liq(sim.model.parameters.earth_param_set)
-    return
-    @. (sim.integrator.u.bucket.σS + sim.integrator.u.bucket.W + sim.integrator.u.bucket.Ws) * ρ_cloud_liq  # kg water / m2
-end
+Interfacer.get_field(sim::BucketSimulation, ::Val{:water}) = sim.integrator.p.bucket.total_water
```
imreddyTeja, line 265: In `ClimaLand.total_liq_water_vol_per_area!`, the sum of (σS, W, and Ws) is not multiplied by ρ_cloud_liq. Is that intentional?

juliasloan25: I missed that! I'll fix it now

imreddyTeja (approved):
Looks good to me. 

I don't think this an issue, but why did we choose to update `total_water` with volume per area in ClimaLand, rather than doing the multiplication by `ρ_cloud_liq` there?

juliasloan25:
> Looks good to me.
> 
> I don't think this an issue, but why did we choose to update `total_water` with volume per area in ClimaLand, rather than doing the multiplication by `ρ_cloud_liq` there?

I'm not sure, but I think it's because we always store water as a volumetric fraction in ClimaLand so we're keeping it consistent there and we only convert to kg/m3 in the coupler where it's used that way.

PR #1448, opened 2025-08-04
remove atmos_config_repo and update atmos config

### LINKED ISSUE #1443: Define ClimaAtmos configs in ClimaCoupler
ClimaCoupler mainly uses 2 atmos configs, and overwrites options when necessary. Currently, we access these configs from the ClimaAtmos repo. It would be better to define them in ClimaCoupler and become independent of ClimaAtmos configs. This has the benefits that we can see the entire config in one place, and any changes in atmos won't unintentionally change coupled runs.

The two files we use are:
```
"config/longrun_configs/amip_target_diagedmf.yml"
"config/longrun_configs/amip_target_edonly.yml"
```
(paths correspond to ClimaAtmos.jl repo)

szy21:
## Purpose 

Removes atmos_config_repo and adds two atmos config for diagnostic edmf and ed-only case. ClimaCoupler now doesn't depend on the configuration files in ClimaAtmos anymore.

Closes #1443 

## Content

----


config/amip_configs/climaatmos_diagedmf.yml
szy21, line 1: Do you prefer to make this in alphabetical order? Is there an easy way to do this?

szy21:
I also removed a few duplicate atmos configs in config/amip_configs/, but let me know if you want to have them back.

PR #1457, opened 2025-08-13
Bump Thermodynamics compat; Bump ClimaEarth project compats

imreddyTeja:
This only bumps the compats. It does not modify any Manifests. This should allow the nightly amip to run

## Content

----


szy21 (approved):
Thanks! I don't know if it's better to update Manifest too? @juliasloan25 should know better.

PR #1463, opened 2025-08-20
add option detect_restart_files

### LINKED ISSUE #1462: Add an option for detecting restart files
Currently, the coupler is always detecting restart files automatically. It would be good to have an option for not restarting the simulation automatically, similar to the `detect_restart_file` in atmos. We should change it here: https://github.com/CliMA/ClimaCoupler.jl/blob/eeb4790ccb5b6848a4ebf8adc315224dd84bcf25/experiments/ClimaEarth/setup_run.jl#L167

juliasloan25:
## Purpose 
Add a CLI option to signal whether restart files should be automatically used to restart a simulation. This is true by default, but can be set to false if a user wishes to run a simulation from the beginning, ignoring present restart files.

Closes #1462


juliasloan25:
@szy21 I updated so the default is false, and the only runs with this option true are the ones that test restart capabilities. Let me know if you want me to enable it for any other configs!

PR #1475, opened 2025-09-04
fix land interpolation method; increase longrun timeout

juliasloan25:
## Purpose 
Fixes for the most recent [longruns](https://buildkite.com/clima/climacoupler-longruns/builds/927#0198ff24-fd0d-43ca-b26d-20e9d8a268f6)

## Content
- [x] Increase longrun timeout limit from 24 -> 30 hours. A lot of our runs get very close to the 24h time limit, and it's annoying for them to fail to often from it.
- [x] Don't use PeriodicCalendar for the integrated land LAI artifact. Since we switched to use the multiyear artifact we don't need to anymore, and PeriodicCalendar errors.
- [x] Extend CMIP runs to 6 months of simulation time, so we can see if they're stable for that long (not a fix, but something we want to know).

----


experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -321,12 +321,7 @@ function create_canopy_args(::Type{FT}, domain, earth_param_set, start_date, sto
     # Set up photosynthesis
     photosynthesis_args = (; parameters = CL.Canopy.FarquharParameters(FT, is_c3; Vcmax25 = Vcmax25))
     # Set up plant hydraulics
-    LAIfunction = CL.prescribed_lai_modis(
-        surface_space,
-        start_date,
-        stop_date;
-        time_interpolation_method = LinearInterpolation(PeriodicCalendar()),
-    )
+    LAIfunction = CL.prescribed_lai_modis(surface_space, start_date, stop_date)
```
imreddyTeja, line 324: It's odd that this doesn't work with the PeriodicCalendar. Does the multiyear lai have a non uniform time dimension?

imreddyTeja (approved):

PR #1484, opened 2025-09-23
use JuliaFormatter; remove .dev

juliasloan25:
## Purpose 
Remove .dev folder and update to use JuliaFormatter, as most other CliMA packages do by now. 

## Content
- [x] remove .dev folder
- [x] add JuliaFormatter toml, update formatter GHA
- [x] add "Contributing" docs page explaining how to use the formatter
- [x] apply the formatter to all files (the margin has changed from 120 to 92, which makes a lot of changes)

----


.JuliaFormatter.toml
```diff
@@ -0,0 +1,6 @@
+margin = 92
+indent = 4
+always_for_in = true
+whitespace_typedefs = true
+whitespace_ops_in_indices = true
+remove_extra_newlines = false
```
nefrathenrici, line 6: You could add `join_lines_based_on_source = true` to prevent spreading lists of arguments onto separate lines. https://domluna.github.io/JuliaFormatter.jl/dev/#join_lines_based_on_source

nefrathenrici (approved):
This looks good! I like the new contributing page and think having a margin of 92 makes the code easier to read.

.JuliaFormatter.toml
nefrathenrici: If you’re past the margin, join_lines_based_on_source = true allows you to preserve your own grouping rather than the formatter splitting one argument per line. With false, it ignores your grouping and normalizes to one-per-line.
Here's an example:
```
function f(long_args1, long_args2, long_args3, long_args4) # larger than margin
    return 2
end
```
With `join_lines_based_on_source = false`, each argument is split onto a new line when they collectively exceed the margin:
```
function f(
    long_args1, 
    long_args2, 
    long_args3, 
    long_args4,
)
    return 2
end
```

With `join_lines_based_on_source = true` you can group arguments onto the same lines without the formatter splitting them up. This doesn't happen automatically but gives you more control.
```
function f(
    long_args1, long_args2, 
    long_args3, long_args4
)
    return 2
end
```

juliasloan25: Actually I tried the formatter with this option and I prefer without, so I'll leave the formatter options as-is

PR #1495, opened 2025-10-02
Modify sea ice surface fluxes and bulk initialization

costachris:
Zero LHF over sea ice through beta. Pass surface temperature estimate (rather than bulk) to coupler. Initialize bulk temperature below freezing.


szy21 (approved):
Thanks!

PR #1496, opened 2025-10-02
fix time interpolation method for prescribed lai

szy21:
## Purpose 

To fix the error here: https://buildkite.com/clima/climacoupler-longruns/builds/949#01998f57-1299-4d1a-8d77-b5672019dd92/171-286

## Content

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_integrated.jl
+++ b/experiments/ClimaEarth/components/land/climaland_integrated.jl
@@ -144,7 +144,7 @@ function ClimaLandSimulation(
         surface_space,
         start_date,
         stop_date;
-        time_interpolation_method = LinearInterpolation(PeriodicCalendar()),
+        time_interpolation_method = LinearInterpolation(),
     )
 
     model = CL.LandModel{FT}(forcing, LAI, toml_dict, domain, dt)


juliasloan25 (commented):
This is okay but will restrict us to running in the years 2000-2020 with the integrated land model since the default `extrapolation_bc = Throw()` errors outside of the data range. The only other option would be `extrapolation_bc = Flat()` to reuse the first/last available LAI when outside of the data range. That would make us lose the seasonal cycle

szy21:
> This is okay but will restrict us to running in the years 2000-2020 with the integrated land model since the default `extrapolation_bc = Throw()` errors outside of the data range. The only other option would be `extrapolation_bc = Flat()` to reuse the first/last available LAI when outside of the data range. That would make us lose the seasonal cycle

@kmdeck what do you prefer?

kmdeck:
> > This is okay but will restrict us to running in the years 2000-2020 with the integrated land model since the default `extrapolation_bc = Throw()` errors outside of the data range. The only other option would be `extrapolation_bc = Flat()` to reuse the first/last available LAI when outside of the data range. That would make us lose the seasonal cycle
> 
> @kmdeck what do you prefer?

ack, idk. I feel like for AMIP simulations, Throw() is good, and 20 years of sim time is plenty to work with for now. 

What about for weather quest, though? It's a little annoying we cant use the periodic calendar with MODIS data. I dont totally understand why. Maybe the best alternative is to create a climatological/typical year LAI with regular sampling and repeat that year over and over

juliasloan25 (approved):

PR #1498, opened 2025-10-04
use individual surface model temps for turbulent fluxes

### LINKED ISSUE #1493: compute surface fluxes using the individual T of each surface model
In our current "partitioned surface fluxes", we compute surface fluxes for each component model, but this is done using a combined `T_sfc`. This means that along a coastline we'd be using the same T (and rho, q) to compute fluxes for both surface models.

What we should do instead is store a surface temperature for each surface model in the coupler fields. Then we compute fluxes using that surface temperature, and average the fluxes afterwards to get the surface fluxes (this part we already do now).

Note 1: We'll still need to store `T_sfc` in the coupler fields for radiative fluxes, but it should also be combined differently - see #1492

Note 2: Our current implementation works fine only because we currently use binary area fractions (i.e. area masks), so no grid cell is ever split between multiple surfaces. After we fix this, we should re-enable fractional coastlines.

## To do
- [x] use individual surface model temps for surface flux calculations (no need to store in coupler fields, but store in temp fields within `compute_surface_fluxes!`)
- [x] as we fix this, we should add more temp fields to the coupler fields to remove allocations in the surface flux calculation [here](https://github.com/CliMA/ClimaCoupler.jl/blob/d2bf7d687c3f73dc1e47b8518336786959a6a2c5/src/FluxCalculator.jl#L241-L247) - we should time a step before this change (on main) and after (on this branch) to tell if allocating or recomputing is faster
- [x] remove coupler fields that are used only in turbulent flux calculations. Instead, we can retrieve the field for each component model and store in a temp field within `compute_turbulent_fluxes!` (T_sfc, q_sfc, z0m_sfc, z0b_sfc, beta)
- [x] update docs
  -  what do surface models need to provide? anything used for turbulent flux calculations, radiation computation, anything else?
  - what should be stored in coupler fields vs accessed when needed? access flux calculation fields as-needed since this is done per component model (no need for global combination across surfaces in coupler fields)

juliasloan25:
## Purpose 
closes #1493

## Content

----

### DIFF
--- a/src/FluxCalculator.jl
+++ b/src/FluxCalculator.jl
@@ -237,7 +237,9 @@ function compute_surface_fluxes!(
     atmos_sim::Interfacer.AtmosModelSimulation,
     thermo_params,
 )
+    # TODO we need temp fields for: T_sfc, q_sfc, ρ_sfc, uh_int (SVector-valued Field), area_fraction, z0m, z0b, beta
     boundary_space = axes(csf)
+    # TODO store atmos fields in coupler fields so we only regrid once per timestep instead of 3x
     # TODO lots of allocations here
     # `_int` refers to atmos state of center level 1
     z_int = Interfacer.get_field(atmos_sim, Val(:height_int), boundary_space)
@@ -252,6 +254,7 @@ function compute_surface_fluxes!(
 
     # construct the surface thermo state
     # get surface air density by extrapolating atmospheric density to the surface
+    # TODO store T_sfc in a temp field so we get it per-component
     ρ_sfc = extrapolate_ρ_to_sfc.(thermo_params, thermo_state_atmos, csf.T_sfc)
 
     # compute surface humidity from the surface temperature, surface density, and phase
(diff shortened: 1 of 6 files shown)


src/FluxCalculator.jl
```diff
@@ -238,43 +266,64 @@ function compute_surface_fluxes!(
...
+
     # get surface air density by extrapolating atmospheric density to the surface
-    ρ_sfc = extrapolate_ρ_to_sfc.(thermo_params, thermo_state_atmos, csf.T_sfc)
+    ρ_sfc = extrapolate_ρ_to_sfc.(thermo_params, thermo_state_atmos, T_sfc)
 
-    # compute surface humidity from the surface temperature, surface density, and phase
-    thermo_state_sfc = TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, csf.T_sfc, csf.q_sfc)
+    # construct the surface thermo state
+    # after this we can reuse `scalar_temp1` and `scalar_temp2` again
+    thermo_state_sfc = TD.PhaseEquil_ρTq.(thermo_params, ρ_sfc, T_sfc, q_sfc)
```
szy21: Same here for TD.PhaseNonEquil_ρTq

PR #1501, opened 2025-10-09
add ClimaSeaIce model

### LINKED ISSUE #1391: use ClimaSeaIce for a more complex sea ice model
We should begin using ClimaSeaIce.jl. One consideration is that we'll need to compute ocean/sea ice fluxes in this more complex case. We have two options here: add these flux calculations to SurfaceFluxes.jl or ClimaCoupler.jl so that each component in ClimaCoupler is distinct, or use ClimaOcean (which already implemented ocean/ice fluxes) and have a coupled ocean/sea ice within the larger coupled simulation. We need to decide which approach is preferred.

As of 9/25/25 we'll try implementing them separately, and call ClimaOcean functions to compute the ocean/sea ice fluxes.

juliasloan25:
## Purpose 
Integrate ClimaSeaIce into the ClimaCoupler ecosystem.

Closes #1391

## Content
- [x] add `struct ClimaSeaIceSimulation` in components/ocean/clima_seaice.jl which inherits a lot of setup from the ocean
- [x] call `CO.compute_sea_ice_ocean_fluxes` from ClimaCoupler to compute fluxes - requires https://github.com/CliMA/ClimaOcean.jl/pull/627
- [x] only `use_restoring` in ocean model when not using clima sea ice model (i.e. when using prescribed sea ice)
- [ ] separate drivers for each simulation mode to avoid so many if/else branches (separate PR)

----


juliasloan25:
> How would I run a coupled simulation from this PR? I think it would be nice to post some videos/visualizations in this PR.

Sorry I forgot you had asked me to explain how to run it in the PR. You can run this (it'll get cleaner once we have a config file for this setup):
```
cd ClimaCoupler.jl/experiments/ClimaEarth
julia --project
include("setup_run.jl")

config_file = "../../config/ci_configs/my_first_cmip.yml"
config_dict = get_coupler_config_dict(config_file)
config_dict["ice_model"] = "clima_seaice"
cs = CoupledSimulation(config_file)
run!(cs) (or step!(cs))
```

experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -287,6 +323,13 @@ function FluxCalculator.update_turbulent_fluxes!(sim::OceananigansSimulation, fi
     F_turb_ρτxz_cc = sim.remapping.scratch_cc1
     F_turb_ρτyz_cc = sim.remapping.scratch_cc2
 
+    # Weight by (1 - sea ice concentration)
+    # TODO does this work with OC fields?
+    OC.interior(F_turb_ρτxz_cc, :, :, 1) .=
+        OC.interior(F_turb_ρτxz_cc, :, :, 1) .* (1.0 .- ice_concentration)
+    OC.interior(F_turb_ρτyz_cc, :, :, 1) .=
+        OC.interior(F_turb_ρτyz_cc, :, :, 1) .* (1.0 .- ice_concentration)
```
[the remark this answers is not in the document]
simone-silvestri: If you want to be cleaner, we can write a kernel for it.

experiments/ClimaEarth/components/ocean/clima_seaice.jl
```diff
@@ -0,0 +1,521 @@
...
+
+"""
+    ClimaSeaIceSimulation()
+
+Creates an ClimaSeaIceSimulation object containing a model, an integrator, and
+a surface area fraction field.
+This type is used to indicate that this simulation is an ocean simulation for
+dispatch in coupling.
+
+Initially, no sea ice is present.
```
szy21: This is not the case if a start_date is provided? (and we always have a start_date for CMIP / AMIP?)

juliasloan25: oops, the docstring isn't updated. I'll fix it

experiments/ClimaEarth/components/ocean/clima_seaice.jl
```diff
@@ -0,0 +1,477 @@
...
+
+    oc_flux_S = surface_flux(ocean_sim.ocean.model.tracers.S)
+    OC.interior(oc_flux_S, :, :, 1) .+=
+        OC.interior(ice_concentration, :, :, 1) .*
+        OC.interior(ice_sim.ocean_ice_fluxes.salt, :, :, 1)
+
+    return nothing
+end
+
+@kernel function _add_ocean_ice_stress!(
```
juliasloan25: add docstring

PR #1506, opened 2025-10-14
use PhaseNonEquil instead of PhaseEquil

### LINKED ISSUE #1502: use `TD.PhaseNonEquil_ρTq` instead of `TD.PhaseEquil_ρTq`
According to @szy21 : it will be more correct to use `TD.PhaseNonEquil_ρTq` instead of `TD.PhaseEquil_ρTq` in `compute_surface_humidity!` and for the atmos and surface thermo states in `compute_surface_fluxes!`

juliasloan25:
## Purpose 
for improved correctness

PhaseEquil is still used for the integrated land model fluxes because it requires changes in ClimaLand.jl - see https://github.com/CliMA/ClimaLand.jl/issues/1492

closes #1502

CI runs look unchanged, surprisingly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@ ClimaCoupler.jl Release Notes
 
 ### ClimaCoupler features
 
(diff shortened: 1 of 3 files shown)


szy21 (approved):
Thanks!

PR #1511, opened 2025-10-17
use update_turbulent_fluxes; don't area weight rad flux

### LINKED ISSUE #1494: rename atmos `update_field!` for turbulent fluxes to `update_turbulent_fluxes!`
The surface models each have a function `update_turbulent_fluxes!`. We should update the naming used by the atmosphere model to match this too. See [this TODO](https://github.com/CliMA/ClimaCoupler.jl/blob/a2bf8c79975b7838125192c5bdd584541b7b9cfc/src/FluxCalculator.jl#L73) in the code.

We can also remove update_field for turbulent fluxes from the atmosphere `update_sim!` method. This will also make the atmosphere more consistent with the surface models.

Update 10/17/25: Actually I decided to go the other way and rename `update_turbulent_fluxes!` to be an `update_field!`. This way the turbulent fluxes are handled in a consistent way with the rest of the coupler updates.

### LINKED ISSUE #1509: Don't weight `F_radiative` by `area_fraction`
This isn't correct when we have fractional area fractions. F_radiative and precipitations shouldn't be weighted

juliasloan25:
## Purpose 
Closes #1494
Closes #1509

## To-do
- [x] Rename the `update_field!` atmosphere method for turbulent fluxes to `update_turbulent_fluxes!`
- [x] Remove the area fraction weighting in FluxCalculator.jl default radiative flux update function

### DIFF
--- a/src/FieldExchanger.jl
+++ b/src/FieldExchanger.jl
@@ -195,7 +195,6 @@ function update_sim!(atmos_sim::Interfacer.AtmosModelSimulation, csf)
     )
     Interfacer.update_field!(atmos_sim, Val(:emissivity), csf.emissivity)
     Interfacer.update_field!(atmos_sim, Val(:surface_temperature), csf)
-    Interfacer.update_field!(atmos_sim, Val(:turbulent_fluxes), csf)
     return nothing
 end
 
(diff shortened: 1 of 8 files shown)


src/FieldExchanger.jl
```diff
@@ -209,14 +208,8 @@ Updates the surface component model cache with the current coupler fields beside
 - `csf`: [NamedTuple] containing coupler fields.
 """
 function update_sim!(sim::Interfacer.SurfaceModelSimulation, csf, area_fraction)
```
szy21: We don't need `area_fraction` here anymore?

juliasloan25: I don't think so. This is the case we talked about the other day where we get the radiative flux computed by the atmosphere and provide it to the surface models in `csf.F_radiative`. We decided we should pass the unweighted flux to each surface

PR #1512, opened 2025-10-17
Do not add title again for debug plots

### LINKED ISSUE #1477: labels for integrated land debug plots are messed up
The first time I see this happen is this PR: https://github.com/CliMA/ClimaCoupler.jl/pull/1467

For example:
Image

ph-kev:
closes #1477 - This PR adds an empty title when it is a ClimaCore Field, because `_heatmap_cc_field!` will add the title.

This is an image of the debug plots from Buildkite.
image


juliasloan25:
Thank you! I think we want to remove the title from `_heatmap_cc_field!` instead, so that the original extrema get used in the title. Then we can tell if there are NaNs/Infs in the data, and still be able to see the plot.

juliasloan25 (approved):

PR #1524, opened 2025-10-22
change Oceananigans model closures

juliasloan25:
## Purpose 
Change the physics of the OceananigansSimulation model, per Simone's suggestions. This is more similar to the model setup in the ClimaOcean [one degree example](https://clima.github.io/ClimaOceanDocumentation/stable/literated/one_degree_simulation/), but not exactly the same.

Taken from #1501 to make that PR smaller.
This may be behavior changing for CMIP, but I don't see any changes visually in the buildkite plots.


experiments/ClimaEarth/components/ocean/oceananigans.jl
```diff
@@ -108,7 +108,24 @@ function OceananigansSimulation(
     end
 
     # Create ocean simulation
-    ocean = CO.ocean_simulation(grid; forcing)
+    free_surface = OC.SplitExplicitFreeSurface(grid; substeps = 70)
+    momentum_advection = OC.WENOVectorInvariant(order = 5)
+    tracer_advection = OC.WENO(order = 5)
+    eddy_closure = OC.TurbulenceClosures.IsopycnalSkewSymmetricDiffusivity(
+        κ_skew = 2e3,
+        κ_symmetric = 2e3,
```
simone-silvestri:
```suggestion
        κ_skew = 1e3,
        κ_symmetric = 1e3,
```

simone-silvestri (approved):

PR #1529, opened 2025-10-26
add integrated land nightly run

szy21:
## Purpose 

Adds a nightly run for integrated land. I didn't change the weekly amip (which has a higher resolution atmos) because it breaks at ~20 weeks now. Once the amip with integrated land is more stable and faster we can remove the bucket from nightly amip and change the weekly amip.

Also removes some extra diagnostics we don't need.

## Content

----

### DIFF
--- a/.buildkite/nightly/pipeline.yml
+++ b/.buildkite/nightly/pipeline.yml
@@ -62,9 +62,24 @@ steps:
           slurm_cpus_per_task: 4
           slurm_ntasks: 1
           slurm_mem: 30GB
+      
+      - label: "Coarse current AMIP: diagedmf + 0M + integrated land"
+        key: "amip_land"
+        command:
+          - echo "--- Run simulation"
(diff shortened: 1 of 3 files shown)


.buildkite/nightly/pipeline.yml
```diff
@@ -62,9 +62,24 @@ steps:
...
+          CLIMACOMMS_DEVICE: "CUDA"
+        agents:
+          slurm_gpus_per_task: 1
+          slurm_cpus_per_task: 4
+          slurm_ntasks: 1
+          slurm_mem: 30GB
 
       - label: "Coarse current AMIP: progedmf + 0M + bucket land"
-        key: "amip_random1"
+        key: "amip_progedmf"
```
szy21: do we need the key?

juliasloan25 (approved):

.buildkite/nightly/pipeline.yml
juliasloan25: I'm not sure. I think it's best to just include it for now

PR #1537, opened 2025-10-30
use binary ice fraction and fix the temperature in sea ice radiation

szy21:
Uses binary sea ice as it seems to be more stable in amip. Also fixes the surface temperature used in LW radiation of sea ice, and renames T_sfc in sea ice model to T_bulk for clarity.


juliasloan25 (approved):
looks good!

PR #1541, opened 2025-10-31
add a test to compare atmosphere and surface fluxes

### LINKED ISSUE #1536: Add a test for matching atmosphere/surface fluxes
We can catch more bugs earlier if we have a test case that does the following:
- set up a coupled simulation
- run for 1 step
- check that the fluxes in the atmosphere match those of each surface

This can be a separate experiments/ script that runs in buildkite, so it doesn't slow down the runtests and is run for each PR.

juliasloan25:
## Purpose 
closes #1536

## To-do
- [x] add a test comparing fluxes in atmosphere to bucket land, prescribed ocean, and prescribed sea ice
- [ ] add a test comparing fluxes in atmos to integrated land (separate PR?)

## Content

----

### DIFF
--- a/experiments/ClimaEarth/test/fluxes_test.jl
+++ b/experiments/ClimaEarth/test/fluxes_test.jl
@@ -1,0 +1,91 @@
+#
+# Flux consistency test (AMIP + bucket land)
+#
+# This test sets up an AMIP coupled simulation (atmosphere + bucket land + prescribed
+# ocean + prescribed sea ice), advances one coupling step, and verifies that the
+# surface radiative flux seen by the atmosphere matches what each surface model
+# computes/stores:
+# - Atmosphere: uses `sim.integrator.p.radiation.ᶠradiation_flux` (positive downward).
+# - Bucket land: compares against `sim.integrator.p.bucket.R_n` on land-dominant cells.
+# - Prescribed ocean: skipped — SST is prescribed so radiative fluxes are not computed
+#   in the same way.
+# - Prescribed sea ice: not stored directly; compute using cache fields and compare
+#   to the atmospheric flux on ice-dominant cells.
+
+import Test: @test, @testset
+
+# Use the AMIP setup helpers to construct a coupled simulation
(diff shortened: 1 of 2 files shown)


experiments/ClimaEarth/test/fluxes_test.jl
```diff
@@ -0,0 +1,96 @@
...
+    (; atmos_sim, land_sim, ocean_sim, ice_sim) = cs.model_sims
+
+    # Atmosphere: radiative flux on the surface interface
+    # Convention: positive downward to the surface
+    atmos_flux = CC.Spaces.level(
+        atmos_sim.integrator.p.radiation.ᶠradiation_flux.components.data.:1,
+        CC.Utilities.half,
+    )
+
+    # Integrated land: compare to net radiation stored in the bucket cache
```
juliasloan25:
```suggestion
    # Bucket land: compare to net radiation stored in the bucket cache
```

experiments/ClimaEarth/test/fluxes_test.jl
```diff
@@ -0,0 +1,96 @@
+#
+# Flux consistency test (AMIP + bucket land)
+#
+# This test sets up an AMIP coupled simulation (atmosphere + bucket land + prescribed
+# ocean + prescribed sea ice), advances one coupling step, and verifies that the
+# surface radiative flux seen by the atmosphere matches what each surface model
+# computes/stores:
+# - Atmosphere: uses `sim.integrator.p.radiation.ᶠradiation_flux` (positive downward).
+# - Bucket land: compares against `sim.integrator.p.bucket.R_n` with opposite sign
+#   convention (so atmos ≈ -R_n) on land-dominant cells.
```
szy21:
```suggestion
# - Bucket land: compares against `sim.integrator.p.bucket.R_n` on land-dominant cells.
```

szy21 (approved):
Thanks! Also, I realized that with our current way of computing radiation, there will be an offset of one timestep between atmosphere and land, so they won't be the same. We can think about what the best way to test it is.

Issue #1520, opened 2025-10-21
Don't hardcode constants


juliasloan25:
## Constants to get from ClimaParams
- Stefan-Boltzmann constant (sigma in LW_u computation for each surface model)
Right now we hardcode this in each surface model as we compute upwelling LW radiation. This change will require a new ClimaAtmos simulation constructor that can accept a TOML dict.

- Celsius/Kelvin converstion
Right now we hardcode `273.15` multiple times in the oceananigans.jl file - we should get it from a central parameter dict instead

See also #342, #70

Issue #1372, opened 2025-05-27
get boundary space radius from ClimaParams


juliasloan25:
When we construct a boundary space with `share_surface_space == false`, we should get the radius from ClimaParams instead of hardcoding it.

We should be able to do something like:
```
toml_dict = CP.create_toml_dict(FT, override_file=parameter_files[1])
(; radius) = CP.get_parameter_values(toml_dict, "planet_radius")
```

Issue #342, opened 2023-07-01
Unify parameter passing


LenkaNovak:
We need to ensure consistent passing of CLIMAParameters and other general parameter sets within the coupled system
This includes `CLIMAParameters`, `thermo_params` and `surface_params` which are now called separately in Atmos and Land.

PR #1542, opened 2025-10-31
use ClimaParams for shared parameters in coupled sims

juliasloan25:
## Purpose 
Get parameters from ClimaParams instead of hardcoding them in ClimaCoupler.

Note that any parameter overrides specified in `parameter_files` are now applied at the coupler level (in the driver), then passed to component models as a TOML dict. This might change interfaces for calibration and/or weather quest.

closes https://github.com/CliMA/ClimaCoupler.jl/issues/1520
closes https://github.com/CliMA/ClimaCoupler.jl/issues/1372

A step towards https://github.com/CliMA/ClimaCoupler.jl/issues/342 - see that issue for next steps

## Content
- parameters to get from CP
  - sigma (Stefan Boltzmann constant)
  - C_to_K
  - thermo_params
  - planet radius (note this was previously wrong when we constructed a boundary space independent of the atmosphere space, i.e. share_surface_space = false)
- add constructors for prescribed ice and slab ocean parameters that take in a toml dict
- for bucket and land constructors, take in a toml dict instead of a list of parameter files (parameters from these files are overriden at the coupler level now)
- confirmed that buildkite plots are unchanged as compared to main

## To Do
remaining:
atmos accepting a toml dict (maybe can just call AtmosConfig directly)
oceananigans/climaseaice accepting a toml dict of parameters
add tests to verify that components are using the same values

### DIFF
--- a/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
+++ b/experiments/ClimaEarth/components/atmosphere/climaatmos.jl
@@ -5,6 +5,7 @@ import LinearAlgebra
 import ClimaAtmos as CA
 import ClimaAtmos: set_surface_albedo!
 import ClimaAtmos.Parameters as CAP
+import ClimaParams as CP
 import ClimaCore as CC
 import ClimaCore.Geometry: ⊗
 import SurfaceFluxes as SF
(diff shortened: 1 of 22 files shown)


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -659,7 +670,25 @@ function get_atmos_config_dict(coupler_config::Dict, atmos_output_dir)
         atmos_config["restart_file"] =
             replace(atmos_config["restart_file"], "active" => "0000")
     end
-    return CA.AtmosConfig(atmos_config)
+
+    # Construct the AtmosConfig object
+    toml_dict = coupled_param_dict
```
nefrathenrici: I'm a bit confused by this toml_dict redefinition, do we not want to use the existing `toml_dict`?

nefrathenrici (approved):
Thanks! This looks good, though I'm a bit confused by the planet radius units.

experiments/ClimaEarth/components/atmosphere/climaatmos.jl
juliasloan25: Oops yes you are right

juliasloan25:
> It's nice that we don't hardcode the parameters now! It would be great if in the end we can have one toml file that controls everything.

Agreed! I updated this paramter-related issue to be more specific about what we want https://github.com/CliMA/ClimaCoupler.jl/issues/342

PR #1546, opened 2025-11-04
remove FluxCalculator.surface_inputs

juliasloan25:
## Purpose 
Since we now remap all quantities onto the boundary space when we compute fluxes, there's no need to access the underlying data layouts of ClimaCore Fields, which was previously being done via `maybe_fv`.

Once we remove `maybe_fv` from `FluxCalculator.surface_inputs`, that becomes a one line function that we can just call directly rather than wrapping it in a separate function.

Plots in buildkite look visually unchanged.

### DIFF
--- a/src/FluxCalculator.jl
+++ b/src/FluxCalculator.jl
@@ -72,53 +72,6 @@ function turbulent_fluxes!(csf, model_sims, thermo_params)
     return nothing
 end
 
-
-function surface_inputs(input_args::NamedTuple)
-    (;
-        thermo_state_sfc,
-        thermo_state_atmos,
-        uₕ_int,
-        z_int,
-        z_sfc,
-        scheme_properties,
-        boundary_space,
-    ) = input_args
-    FT = CC.Spaces.undertype(boundary_space)
-    (; z0b, z0m, beta, gustiness) = scheme_properties
-
-    # Extract the underlying data layouts of each field
-    # Note: this is a bit "dangerous" because it circumvents ClimaCore, but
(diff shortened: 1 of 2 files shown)


src/FluxCalculator.jl
```diff
@@ -324,20 +277,22 @@ function compute_surface_fluxes!(
...
+    inputs = @. SF.ValuesOnly(
+        SF.StateValues(csf.z_int, uₕ_int, thermo_state_atmos), # state_in
+        SF.StateValues(                                  # state_sfc
+            csf.z_sfc,
+            StaticArrays.SVector(FT(0), FT(0)),
+            thermo_state_sfc,
+        ),
+        z0m,
+        z0b,
+        gustiness,
```
akshaysridhar: `gustiness` and `beta` are `kwargs` within the existing constructors - so using the `;` separator should clarify this here.

src/FluxCalculator.jl
```diff
@@ -324,20 +277,22 @@ function compute_surface_fluxes!(
...
-        csf.z_int,
-        csf.z_sfc,
-        scheme_properties,
-        boundary_space,
-        surface_params,
+    # Set some scalars that we hardcode for now
+    gustiness = FT(1)
+
+    # Construct the SurfaceFluxes.jl container of inputs
+    inputs = @. SF.ValuesOnly(
```
akshaysridhar: Thanks for modifying this. The result of this PR is consistent with the expected outcome when running in `CPUMultiThreaded`; will test and report the GPU result shortly.

akshaysridhar (approved):

PR #1548, opened 2025-11-05
add an option `restart_cache`

### LINKED ISSUE #1544: Cannot restart a simulation with a different dt
See the error here: https://buildkite.com/clima/climacoupler-amip/builds/201#019a4ae1-47be-4fd3-9c59-4bdbff0f52e4/173-380

juliasloan25:
## Purpose 
Adds an option so that if restart files are available, we can choose to restart the state only (`restart_cache` false), or to restart both the state and the cache (`restart_cache` true). `restart_cache` is true by default, recovering the previous behavior.

If the cache is restarted (whether or not the state is), we perform an initial component model exchange and call `set_cache!` for each component model to make sure the caches all get fully set (e.g. for radiation).

As a next step we may want to add an option `checkpoint_cache` which we can set to false to avoid saving the very large cache JLD2 files.

closes #1544 

## Content
- [x] don't set dt in the atmos cache from the restart files
- [x] add an option `restart_cache`

cc @szy21

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -127,6 +127,15 @@ steps:
           slurm_gres: "gpu:1"
           slurm_mem: 32GB
 
+      - label: "GPU restarts"
+        command: "julia --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/test/restart_state_only.jl"
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+        agents:
+          slurm_ntasks: 1
+          slurm_gres: "gpu:1"
+          slurm_mem: 32GB
+
   - group: "Integration Tests"
     steps:
       # SLABPLANET EXPERIMENTS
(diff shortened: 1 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -127,6 +127,15 @@ steps:
           slurm_gres: "gpu:1"
           slurm_mem: 32GB
 
+      - label: "GPU restarts"
```
juliasloan25:
```suggestion
      - label: "GPU restarts (state only)"
```

experiments/ClimaEarth/cli_options.jl
```diff
@@ -100,6 +100,10 @@ function argparse_settings()
         help = "Time in seconds rounded to the nearest index to use at `t_start` for restarted simulation [nothing (default)]"
         arg_type = Int
         default = nothing
+        "--restart_cache"
+        help = "Boolean flag indicating whether to read the cache from the restart file if available [`true` (default), `false`]"
```
szy21:
```suggestion
        help = "Boolean flag indicating whether to read the cache from the restart file if available [`true` (default), `false`]". The simulation is not bit-wise reproducible when restarting if this is set to false.
```

szy21 (approved):
Thanks!

PR #1551, opened 2025-11-06
remove extra atmos diagnostics

szy21:
## Purpose 

We don't need these now. We can add them to ClimaAtmos later if we need them.

## Content

----


szy21:
I think we can remove `add_extra_diagnostics!` at some point too, but right now we are using it to get `diagnostics_dt`. Maybe we can remove it later.

juliasloan25 (approved):

PR #1556, opened 2025-11-07
fix sea ice conductive flux

szy21:
## Purpose 

Conductive flux should be k * (T_base - T_sfc) / h, instead of k * (T_base - T_bulk) / h.

## Content

----


juliasloan25 (approved):

juliasloan25:
Can you also update the equation in the docstring of `struct PrescribedIceSimulation`?

PR #1570, opened 2025-11-19
copy AMIP diagnostics to Azure VM for visualization

juliasloan25:
## Purpose 
Add a line to the weekly AMIP buildkite pipeline to send diagnostics to the Azure VM for visualization. Note that this only sends diagnostics, not checkpoints or artifacts (plots).

Next, we will find a way to automate setting up the dashboard and output the link to the user, so we can easily see a dashboard for each weekly AMIP run.

I'm skipping CI since this only modifies the AMIP pipeline, which isn't run as part of CI.

### DIFF
--- a/.buildkite/amip/pipeline.yml
+++ b/.buildkite/amip/pipeline.yml
@@ -58,6 +58,11 @@ steps:
 
     steps:
 
+      - label: "Send simulation output to Azure VM for visualization"
+        command:
+          - rsync -avz experiments/ClimaEarth/output/amip/output_active/* azure-arenchon:/home/arenchon/GitHub/ClimaViz.jl/data/climacoupler-amip/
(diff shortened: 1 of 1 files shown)


nefrathenrici (approved):

AlexisRenchon:
I realize this PR is not good. Sending the data is OK. It's a command from buildkite. 

The other command (sudo) is a command to run on the Azure VM, not on buikdkite. 

And I expect sudo on buildkite would trigger a warning... We should delete that command @nefrathenrici 

In this github action I run a command on the VM https://github.com/CliMA/ClimaViz.jl/blob/main/.github/workflows/hello-vm.yml.. I am not sure yet how to make a buildkite CI run a command on the VM. We can discuss next week

PR #1571, opened 2025-11-19
update for new slack-upload interface

juliasloan25:
## Purpose 
Since slack changed their file uploading API, we also had to change our `slack-upload` script (see changes [here](https://github.com/CliMA/slurm-buildkite/pull/94)). 

The two user-facing changes are
1. instead of providing the channel name, we provide the channel ID
2. the `-m` option is now unused

## Content
- [x] replace `#coupler-report` with channel ID `C04S6MBSDC4`
- [x] remove `-m` options

----


nefrathenrici (approved):
Thanks! The channel ID is confusing but I guess it can't be helped.

PR #1577, opened 2025-11-22
Add save_cache option

### LINKED ISSUE #1561: Add an option for not saving the cache in checkpoints
Now that we have an option `restart_cache` that sets whether to read the cache when restarting, we can have an option for not saving the cache in checkpoints to avoid saving large cache jld2 files. We can add this option to cli_options.jl and parse it in `get_coupler_args`, analogous to the changes in https://github.com/CliMA/ClimaCoupler.jl/pull/1548.

ph-kev:
closes #1561 - This PR adds `save_cache` as an option to save the cache or not. This is useful with `restart_cache = false`.

### DIFF
--- a/experiments/ClimaEarth/cli_options.jl
+++ b/experiments/ClimaEarth/cli_options.jl
@@ -104,6 +104,10 @@ function argparse_settings()
         help = "Boolean flag indicating whether to read the cache from the restart file if available [`true` (default), `false`]"
         arg_type = Bool
         default = true
+        "--save_cache"
+        help = "Boolean flag indicating whether to save the state and cache or only the state when checkpointing [`true` (default), `false`]"
+        arg_type = Bool
+        default = true
         # Diagnostics information
         "--use_coupler_diagnostics"
         help = "Boolean flag indicating whether to compute and output coupler diagnostics [`true` (default), `false`]"
--- a/src/Checkpointer.jl
+++ b/src/Checkpointer.jl
@@ -128,7 +128,8 @@ end
 """
     checkpoint_sims(cs::CoupledSimulation)
 
-This is a callback function that checkpoints all simulations defined in the current coupled simulation.
+This is a callback function that checkpoints all simulations defined in the
+current coupled simulation.
 """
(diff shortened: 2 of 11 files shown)


szy21 (approved):
Thanks!

szy21:
Should we merge this?

ph-kev:
I was hoping for @juliasloan25 to also take a look at it, but I'm okay with merging it in.

experiments/ClimaEarth/cli_options.jl
```diff
@@ -104,6 +104,10 @@ function argparse_settings()
         help = "Boolean flag indicating whether to read the cache from the restart file if available [`true` (default), `false`]"
         arg_type = Bool
         default = true
+        "--save_cache"
+        help = "Boolean flag indicating whether to save the cache or not when checkpointing [`true` (default), `false`]"
```
juliasloan25:
```suggestion
        help = "Boolean flag indicating whether to save the state and cache or only the state when checkpointing [`true` (default), `false`]"
```

src/Checkpointer.jl
```diff
@@ -126,11 +126,13 @@ function restore_cache!(sim::Interfacer.ComponentModelSimulation, new_cache)
...
-    checkpoint_sims(cs::CoupledSimulation)
+    checkpoint_sims(cs::CoupledSimulation, save_cache)
 
-This is a callback function that checkpoints all simulations defined in the current coupled simulation.
+This is a callback function that checkpoints all simulations defined in the
+current coupled simulation. The argument `save_cache` determines whether to save
+the cache or not.
 """
-function checkpoint_sims(cs::Interfacer.CoupledSimulation)
+function checkpoint_sims(cs::Interfacer.CoupledSimulation, save_cache)
```
juliasloan25: It would be nice to maintain the interface of `checkpoint_sims` taking in only `cs`. Could we store `save_cache` in there instead?

juliasloan25 (approved):
Looks good, thank you! Just a couple coments

PR #1581, opened 2025-11-25
Tr/walltime

### LINKED ISSUE #1582: WallTimeReport is calculated incorrectly
The WallTimeReport callback is created within ClimaAtmos, and is based on the atmos simulations steps. This caused the WallTimeReport to incorrectly compensate for compilation, resulting in SYPD that asymptomatically approaches the actual SYPD. The coupler should set `log_progress` to `false` in the atmos config, and use the WallTimeReport in the coupler callbacks

imreddyTeja:
## Purpose 
closes #1582
## To-do
- There is still SYPD fluctuation. I think this may be due to diagnostics

See buildkite runs [here](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/271#_)
There is still variation, but the final sypd matches up with the reported final, and it no longer has a sharp SYPD increase in the first few reports 

## Content
- Add a wall time callback to the coupler
- Add a new cli argument: `atmos_log_progress`, that can be used to enable the ClimaAtmos wall time callback and disable the coupler wall time callback. By default, the coupler wall time callback is used.
- disable NVTX gc annotations. This increases ttfx, and we are not using it

----

### DIFF
--- a/experiments/ClimaEarth/cli_options.jl
+++ b/experiments/ClimaEarth/cli_options.jl
@@ -141,6 +141,10 @@ function argparse_settings()
         help = "An optional YAML file used to overwrite the default model parameters."
         arg_type = String
         default = nothing
+        "--atmos_log_progress"
+        help = "Use the ClimaAtmos walltime logging callback instead of the default ClimaCoupler one [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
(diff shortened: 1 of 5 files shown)


juliasloan25 (approved):
Looks good! Thank you Teja

PR #1595, opened 2025-12-05
remove allocations in combine_surfaces

juliasloan25:
## Purpose 
Before we were calling the allocating `get_field` method. Now we remap and store the surface field in a temp coupler field, so we don't allocate. To me it's also more readable now which is nice.

## Content
- [x] remove allocations in combine_surfaces by remapping the surface field of each simulation onto a temp coupler field
- [x] update tests
- [ ] check buildkite results as compared to main
- [ ] check buildkite performance as compared to main

----

### DIFF
--- a/src/FieldExchanger.jl
+++ b/src/FieldExchanger.jl
@@ -156,10 +156,20 @@ from each surface model when surface fluxes are computed, in `compute_surface_fl
 - `model_sims`: [NamedTuple] containing `ComponentModelSimulation`s.
 """
 function import_combined_surface_fields!(csf, model_sims)
-    combine_surfaces!(csf, model_sims, Val(:emissivity))
-    combine_surfaces!(csf, model_sims, Val(:surface_temperature))
-    combine_surfaces!(csf, model_sims, Val(:surface_direct_albedo))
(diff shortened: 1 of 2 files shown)


ph-kev (approved):

juliasloan25:
This change actually slightly worsens performance. I think it's because of runtime dispatch on the `field_name` Val, so I'm restructuring this function a bit

juliasloan25:
Ok, with the current state of this PR it does speed up simulations. For example comparing [this build](https://buildkite.com/clima/climacoupler-ci/builds/7324) to [main](https://buildkite.com/clima/climacoupler-ci/builds/7310) we see the following SYPD changes:
- GPU slabplanet albedo from function: 2.174 SYPD --> 3.017 SYPD
- GPU AMIP diag. EDMF + bucket: 0.15 SYPD --> 0.274 SYPD

juliasloan25:
physical results
unchanged: slabplanet with albedo from function, CMIP with prescribed sea ice
changed (only MO length, precipitation): ED only + bucket, diag EDMF + bucket, AMIP ED only + integrated land

juliasloan25:
the allocations have been fixed in main

PR #1597, opened 2025-12-05
use 1m in the flagship amip

szy21:
## Purpose 

I'm not sure if this will be stable, but we will know tomorrow.

## Content

----


juliasloan25:
I don't think this will be stable. The [longrun with 1M](https://buildkite.com/clima/climacoupler-longruns/builds/977#019ad3c7-b78b-459c-b94c-a20b221e0ea2) fails around 3 days, but we can try if you want to

szy21:
> I don't think this will be stable. The [longrun with 1M](https://buildkite.com/clima/climacoupler-longruns/builds/977#019ad3c7-b78b-459c-b94c-a20b221e0ea2) fails around 3 days, but we can try if you want to

These only run for 20 secs though:)

petebachant (approved):
LGTM! Invalidates our performance dashboard, but better earlier than later 😄

PR #1599, opened 2025-12-06
use zero(FT) instead of zero(field) in broadcasts

juliasloan25:
## Purpose 
We commonly use the pattern `@. field = ifelse(area_fraction ≈ 0, zero(field), field)`. I was under the impression that the `zero(field)` would be equivalent to `zero(eltype(field))` when broadcasted, but this isn't the case. See the allocation and timing information below

## Content
- [x] replace `zero(field)` with 0 in all broadcasted `ifelse` calls
- [x] some other misc. cleanup of `ifelse` calls

## Timing comparison
Despite having the most allocations, `zero(FT)` is the fastest to run since it the type of the zero doesn't need to be promoted (as it does for e.g. 0).
```julia
julia> @btime @. field1 = ifelse(area_fraction == 0, zero(field1), field1)
  199.083 μs (3 allocations: 128 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, zero(eltype(field1)), field1)
  198.125 μs (4 allocations: 160 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, 0, field1);
  222.667 μs (2 allocations: 96 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, zero(FT), field1)
  174.708 μs (8 allocations: 288 bytes)
```

## Allocation comparison
```julia
julia> @allocated @. field1 = ifelse(area_fraction == 0, zero(field1), field1)
128

julia> @allocated @. field1 = ifelse(area_fraction == 0, zero(eltype(field1)), field1)
160

julia> @allocated @. field1 = ifelse(area_fraction == 0, 0, field1)
96
```

The allocations are even worse when the whole expression isn't broadcasted, which we were doing in a few places:
```julia
julia> @allocated field1 .= ifelse.(area_fraction .== 0, zero(field1), field1)
6368

julia> @allocated field1 .= ifelse.(area_fraction .== 0, zero(eltype(field1)), field1)
96

julia> @allocated field1 .= ifelse.(area_fraction .== 0, 0, field1)
96
```

----

### DIFF
--- a/experiments/ClimaEarth/components/land/climaland_integrated.jl
+++ b/experiments/ClimaEarth/components/land/climaland_integrated.jl
@@ -607,10 +607,8 @@ function FluxCalculator.compute_surface_fluxes!(
     )
 
     # Zero out the fluxes where the area fraction is zero
-    @. csf.scalar_temp1 =
-        ifelse(area_fraction == 0, zero(csf.scalar_temp1), csf.scalar_temp1)
-    @. csf.scalar_temp2 =
-        ifelse(area_fraction == 0, zero(csf.scalar_temp2), csf.scalar_temp2)
+    @. csf.scalar_temp1 = ifelse(area_fraction == 0, zero(FT), csf.scalar_temp1)
+    @. csf.scalar_temp2 = ifelse(area_fraction == 0, zero(FT), csf.scalar_temp2)
 
     # Update the coupler field in-place
     @. csf.F_lh += csf.scalar_temp1 * area_fraction
@@ -628,8 +626,7 @@ function FluxCalculator.compute_surface_fluxes!(
             p.snow.snow_cover_fraction .* snow_dest.vapor_flux
         ) .* ρ_liq,
     )
-    @. csf.scalar_temp1 =
-        ifelse(area_fraction == 0, zero(csf.scalar_temp1), csf.scalar_temp1)
+    @. csf.scalar_temp1 = ifelse(area_fraction == 0, zero(FT), csf.scalar_temp1)
     @. csf.F_turb_moisture += csf.scalar_temp1 * area_fraction
 
     # Combine turbulent momentum fluxes from each component of the land model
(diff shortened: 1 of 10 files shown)


juliasloan25:
cc @petebachant @szy21 @kmdeck 
We use this pattern a lot in the integrated land code, so I expect this will speed up that model. It was used in source too so the performance of everything should get better. We'll see by how much :)

experiments/ClimaEarth/components/land/climaland_integrated.jl
```diff
@@ -607,10 +607,8 @@ function FluxCalculator.compute_surface_fluxes!(
     )
 
     # Zero out the fluxes where the area fraction is zero
-    @. csf.scalar_temp1 =
-        ifelse(area_fraction == 0, zero(csf.scalar_temp1), csf.scalar_temp1)
-    @. csf.scalar_temp2 =
-        ifelse(area_fraction == 0, zero(csf.scalar_temp2), csf.scalar_temp2)
+    @. csf.scalar_temp1 = ifelse(area_fraction == 0, 0, csf.scalar_temp1)
+    @. csf.scalar_temp2 = ifelse(area_fraction == 0, 0, csf.scalar_temp2)
```
ph-kev: I am not sure if this matter, but do we need to do something like `zero(FT)` or something similar to that?

ph-kev (approved):

experiments/ClimaEarth/components/land/climaland_integrated.jl
juliasloan25: zero(FT) can be slightly more type-stable. I'll switch to that

juliasloan25: I also profiled the different options and saw that `zero(FT)` is faster even though it has fewer allocations, probably because when we use 0 the type has to be promoted.
```
julia> @btime @. field1 = ifelse(area_fraction == 0, zero(field1), field1)
  199.083 μs (3 allocations: 128 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, zero(eltype(field1)), field1)
  198.125 μs (4 allocations: 160 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, 0, field1);
  222.667 μs (2 allocations: 96 bytes)

julia> @btime @. field1 = ifelse(area_fraction == 0, zero(FT), field1)
  174.708 μs (8 allocations: 288 bytes)
```

juliasloan25:
I'm not sure if performance speedup at the cost of more allocations is worth it, so I'll close this PR for now. We can revisit this idea once we look into GPU performance more concretely

PR #1615, opened 2025-12-12
Reduce size of atmos cache when making checkpoint

ph-kev:
This PR enables compression, save only parts of the cache, and does not do any unnecessary memory copies to reduce the cache size from `57G` to `1.1G`.

Before this change, the size of the atmos cache from the AMIP simulation is `57G`. With this change and no compression, the size of the atmos cache is now `4.6G`. Then, with compression, the size of the atmos cache is `1.1G`. For compressing the data, it does not seems to take too long, but I am not sure about the decompression.

TODO
- [x] Add tests for iterating over the fields of a struct

### DIFF
--- a/docs/src/checkpointer.md
+++ b/docs/src/checkpointer.md
@@ -108,13 +108,12 @@ forward, but there are still several challenges that need to be solved:
 Point 3. adds significant amount of code and requires component models to
 specify how their cache has to be restored.
 
-If you are adding a component model, you have to extend the
+If you are adding a component model, you have to extend the methods.
 ```
 Checkpointer.get_model_prog_state
 Checkpointer.get_model_cache
 Checkpointer.restore_cache!
 ```
-methods.
 
 `ClimaCoupler` moves objects to the CPU with `Adapt(Array, x)`. `Adapt`
 traverses the object recursively, and proper `Adapt` methods have to be defined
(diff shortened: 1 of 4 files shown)


src/Checkpointer.jl
```diff
@@ -468,4 +509,187 @@ function restore!(
...
+Objects that share the same object ID are not copied, but a reference to the
+object is used instead.
+"""
+function make_atmos_cache(cache)
+    atmos_cache_itr = create_atmos_cache_itr(cache)
+    cache_vec = [] # Elements of the vector can be any type
+    obj_id_to_idx = Dict{UInt64, Int}()
+    for (i, obj) in enumerate(atmos_cache_itr)
+        obj_id = objectid(obj)
+        if obj_id in keys(obj_id_to_idx)
```
juliasloan25: I don't understand this part - does `obj_id_to_idx` check if we've already added this object to the cache vec? It would be good to have some comments to clarify

src/Checkpointer.jl
```diff
@@ -468,4 +509,187 @@ function restore!(
     return nothing
 end
 
+# Needed, because calling adapt on a StaticArray makes it into an array
+function restore!(v1::StaticArrays.StaticArray, v2::Array, comms_ctx; name, ignore)
+    v1 == v2 || error("$name is a immutable but it inconsistent ($(v1) != $(v2))")
+    return nothing
+end
+
+# include("checkpointer_helper.jl")
```
juliasloan25: remove?

juliasloan25 (commented):
Thanks Kevin, this is a great improvement! I left some comments, let me know if you have any questions

Also cc @szy21 - check out the saved cache size decrease!

szy21:
Ohhh great, thank you!

src/Checkpointer.jl
```diff
@@ -468,4 +509,187 @@ function restore!(
...
+    obj_id_to_idx = Dict{UInt64, Int}()
+    for (i, obj) in enumerate(atmos_cache_itr)
+        obj_id = objectid(obj)
+        if obj_id in keys(obj_id_to_idx)
+            # Reuse the same object
+            push!(cache_vec, cache_vec[obj_id_to_idx[obj_id]])
+        else
+            # Instead of calling adapt on the entire atmos cache, we call adapt
+            # on each individual object instead instead. Due to how the adapt
+            # function is implemented for the atmos cache, moving from GPU to
```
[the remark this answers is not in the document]
szy21: ClimaAtmos restart only uses the state (and as a result, may not be reproducible).

src/Checkpointer.jl
```diff
@@ -115,6 +115,50 @@ function checkpoint_model_cache(
...
+            # Reuse the same object if we encountered the same object ID while
+            # traversing the object
+            push!(cache_vec, cache_vec[obj_id_to_idx_map[obj_id]])
+        else
+            # Instead of calling adapt on the entire atmos cache, we call adapt
+            # on each individual object instead instead. Due to how the adapt
+            # function is implemented for the atmos cache, moving from GPU to
+            # CPU results in unnecessary duplicate memory.
+            # If obj is on GPU, then it is moved to CPU and otherwise, it does
+            # no operation and return obj
```
juliasloan25: Could you move this into the docstring instead since it's a little long?

src/Checkpointer.jl
```diff
@@ -498,4 +512,155 @@ function restore!(
...
+    CacheIterator(sim::Interfacer.ComponentModelSimulation)
+
+Create an iterator over the cache in `sim` of objects that should be
+saved.
+
+To use this constructor, the function `get_cache_ignore` must be implemented
+for `sim` which takes in `sim` and returns a set of symbols to ignore while
+iterating through the fields of the cache.
+"""
+function CacheIterator(sim::Interfacer.ComponentModelSimulation)
```
juliasloan25: I'm not sure, but maybe it's better to define this only for `AtmosModelSimulation` for now. I'm on the fence because it can technically be used for other components if `get_cache_ignore` is implemented, but that isn't the case right now and maybe dispatching on `ComponentModelSimulation` here will be misleading.

I'll leave it up to you if you want to restrict this to only atmos, or add a method of `get_cache_ignore` for `ComponentModelSimulation`, which errors and explains that the other checkpointing system should be used or the function needs to be extended.

juliasloan25 (approved):
Thank you! Just a few more comments.

Could you also add some documentation in docs/src/checkpointer.md about the CacheIterator, how it works, and why we want to use it for the atmos model?

docs/src/checkpointer.md
```diff
@@ -127,11 +127,38 @@ Types to watch for:
 - `TimeVaryingInputs` (because they contain `NCDatasets`, which contain pointers
   to files)
 
+!!! Adapt and references
```
juliasloan25: It looks like this doesn't render correctly

Image

we could do:

```suggestion
!!! warning "Adapt and references"
```

juliasloan25 (approved):
this looks great, thanks!

PR #1617, opened 2025-12-13
move non-plotting postprocessing to src (SimOutput)

juliasloan25:
## Purpose 
Moves diagnostics and benchmarks code into a new module in src called SimOutput. ~I'm not sure if this is the best place for the diagnostics code, but it's very short so I don't want to make a module just for that.~

I'm planning to move the rest of the postprocessing code into a ClimaCouplerMakie extension that gets loaded only when Makie is loaded, so that we can keep the ClimaCoupler environment a little smaller. I'll do that in a separate PR

## Content
- [x] moved coupler_diagnostics.jl, benchmarks.jl functions into a new SimOutput module
- [x] updated some benchmarks naming to remove `gpu_` prefix
- [x] added test
- [x] added docs

----


Project.toml
```diff
@@ -23,6 +24,7 @@ ArgParse = "1"
 ClimaAtmos = "0.27, 0.28, 0.29, 0.30, 0.31"
 ClimaComms = "0.6.2"
 ClimaCore = "0.14.25"
+ClimaDiagnostics = "0.2.6, 0.3"
```
ph-kev: Putting both 0.2.6 and 0.3 for the compat for ClimaDiagnostics could lead to correctness issues. There is a breaking release, because the times saved are now at the start of the reduction period instead of at the end. I would put 0.2.6 for now, unless you need this for nightly amip or something similar to that.

docs/src/postprocessor.md
```diff
@@ -0,0 +1,27 @@
+# Postprocessor
+
+The `Postprocessor` module provides utilities for postprocessing simulation results,
+including plotting and performance analysis.
+
+## Diagnostics Setup
+
+The diagnostics setup function configures default diagnostics for AMIP simulations.
+For more information about diagnostics in ClimaCoupler, see the [Diagnostics](@ref) documentation.
```
ph-kev: I find it weird that the diagnostics are part of this, because everything that the ClimaDiagnostics does is during the simulation. I guess it is postprocessing after every single timestep.

experiments/ClimaEarth/user_io/benchmarks.jl
```diff
@@ -144,42 +31,33 @@ else
...
-    "Coupled with progedmf + 1M (16 helem)",
-    run_info_coupled_progedmf_coarse...,
-)
-data = append_table_data(
-    data,
-    "Coupled with progedmf + 1M (30 helem)",
-    run_info_coupled_progedmf_fine...,
+# Set up a dictionary mapping run names to descriptions
+run_names_to_descriptions = Dict(
+    "coupled_progedmf_coarse" => "Coupled with progedmf + 1M (16 helem)",
```
ph-kev: Does this need to be a dictionary? Can this be a vector instead, so the order of the entries in the table is deterministic?

docs/src/postprocessor.md
juliasloan25: I agree it's a little weird. I wasn't sure where to put it, since I didn't want to have a diagnostics module that only has < 100 lines of code.

What if I rename the module from Postprocessor to Output? Then it would make a little more sense to have diagnostics, benchmark table, plotting all there

test/input_tests.jl
```diff
@@ -16,7 +16,7 @@ import YAML
 
     parsed = Input.parse_commandline(settings)
     @test parsed["config_file"] ==
-          joinpath(pkgdir(ClimaCoupler), "config", "ci_configs", "amip_default.yml")
+          joinpath(pkgdir(ClimaCoupler), "config/ci_configs/amip_default.yml")
```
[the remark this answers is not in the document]
juliasloan25: This was a mistake that got merged in at the end of last week because the windows CI wasn't required to merge. The path is hardcoded to use "/" in the config so it comes out like this

This fix was merged in separately so it won't be part of this PR

ph-kev (approved):

juliasloan25:
> Should the experiments/ClimaEarth manifests be updated?

We still import ClimaDiagnostics in experiments/ClimaEarth/ so I don't think we can remove it from the Manifests. I don't think there's any other change to make

PR #1621, opened 2025-12-16
add plotting extension

juliasloan25:
## Purpose 
Moves plotting code out of experiments/ and into a Makie extension and an OceananigansMakie extension.

## To-do
- [x] add plotting extension: ClimaCouplerMakieExt
- [x] add Oceananigans + plotting extension: ClimaCouplerOceananigansMakieExt
- [x] move data_sources.jl, test_rmses.jl into src SimOutput module (not ext)
- [x] add stub functions in src (new module Plotting)
- [x] update places these functions are called to use the extension
- [x] move some component model-specific code out of debug plots and into component model files
- [x] add docs
- [x] check plots are correctly generated in buildkite (debug plots, slabplanet conservation, calibration, CMIP ocean diagnostics)

## Content

----

### DIFF
--- a/docs/src/plotting.md
+++ b/docs/src/plotting.md
@@ -12,14 +12,14 @@ when the plotting packages and Oceananigans.jl are available.
 ## Plotting API
 
 ```@docs
-ClimaCoupler.Plotting.make_diagnostics_plots
-ClimaCoupler.Plotting.make_ocean_diagnostics_plots
-ClimaCoupler.Plotting.debug
-ClimaCoupler.Plotting.debug_plot_fields
-ClimaCoupler.Plotting.debug_plot!
-ClimaCoupler.Plotting.plot_global_conservation
-ClimaCoupler.Plotting.compute_leaderboard
-ClimaCoupler.Plotting.compute_pfull_leaderboard
+Plotting.make_diagnostics_plots
+Plotting.make_ocean_diagnostics_plots
+Plotting.debug
+Plotting.debug_plot_fields
+Plotting.debug_plot!
+Plotting.plot_global_conservation
+Plotting.compute_leaderboard
+Plotting.compute_pfull_leaderboard
 ```
 
 ## ClimaCouplerMakieExt Extension
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,17 @@
 using ClimaCoupler
 using Documenter, Literate
 using CairoMakie, ClimaCoreMakie, GeoMakie, Makie, Poppler_jll, Printf, Oceananigans
+using ClimaCoupler:
+    Checkpointer,
+    ConservationChecker,
+    FieldExchanger,
+    FluxCalculator,
+    Input,
+    Interfacer,
+    SimOutput,
+    TimeManager,
+    Utilities,
+    Plotting
 
 const COUPLER_DIR = joinpath(@__DIR__, "..")
 const EXPERIMENTS_DIR = joinpath(@__DIR__, "..", "experiments")
--- a/docs/src/checkpointer.md
+++ b/docs/src/checkpointer.md
@@ -156,11 +156,11 @@ This approach allows for a signficant reducation in the file size of the cache.
 ## Checkpointer API
 
 ```@docs
-    ClimaCoupler.Checkpointer.get_model_prog_state
-    ClimaCoupler.Checkpointer.get_model_cache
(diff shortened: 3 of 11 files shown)


ext/ClimaCouplerMakieExt/calibration_plots.jl
```diff
@@ -0,0 +1,37 @@
+import EnsembleKalmanProcesses as EKP
+import Statistics: var, mean
+
+function Plotting.scatter_plot(eki::EKP.EnsembleKalmanProcess)
```
juliasloan25: move these back to calibration

src/Plotting.jl
```diff
@@ -0,0 +1,173 @@
...
+    make_diagnostics_plots(
+        output_path::AbstractString,
+        plot_path::AbstractString;
+        output_prefix = "",
+    )
+
+Stub method for plotting diagnostic variables.
+See the ClimaCouplerMakieExt extension for the actual implementation.
+"""
+function make_diagnostics_plots(
```
juliasloan25: make stubs simpler and add _init_ with extension_fns like [this](https://github.com/CliMA/ClimaUtilities.jl/blob/2046cafc03354628a43f525bafffe29f69e0a006/src/Regridders.jl#L75)

imreddyTeja:
Why is Oceananigans being made into a weak dep? Would ClimaOceans and ClimaLand be weak or strong deps?

docs/src/plotting.md
```diff
@@ -0,0 +1,117 @@
...
+```@docs
+ClimaCoupler.Plotting.make_diagnostics_plots
+ClimaCoupler.Plotting.make_ocean_diagnostics_plots
+ClimaCoupler.Plotting.debug
+ClimaCoupler.Plotting.debug_plot_fields
+ClimaCoupler.Plotting.debug_plot!
+ClimaCoupler.Plotting.plot_global_conservation
+ClimaCoupler.Plotting.compute_leaderboard
+ClimaCoupler.Plotting.compute_pfull_leaderboard
+```
```
ph-kev: The docstrings doesn't show up. You might need to add something to the modules in `makedocs`. See https://github.com/CliMA/ClimaAnalysis.jl/blob/b1ab4c14498c1787877d4bf1396ad56c68def0b3/docs/make.jl#L16-L21 for an example.

Project.toml
```diff
@@ -19,18 +20,41 @@ SurfaceFluxes = "49b00bb7-8bd4-4f2b-b78c-51cd0450215f"
...
+GeoMakie = "db073c08-6b98-4ee5-b6a4-5efafb3259c6"
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
+Poppler_jll = "759029c8-65f6-5f13-b93f-2d7e0a4cb77c"
+Printf = "366bfd08-7218-5ad0-9036-d27ba1f5fe54"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[extensions]
+ClimaCouplerMakieExt = ["CairoMakie", "ClimaCoreMakie", "GeoMakie", "Makie", "Poppler_jll", "Printf"]
+ClimaCouplerOceananigansMakieExt = ["CairoMakie", "ClimaCoreMakie", "GeoMakie", "Makie", "Poppler_jll", "Oceananigans", "Printf"]
```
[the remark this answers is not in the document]
ph-kev: Also, this is something for the future, but it is strange that we only support CairoMakie and not any other Makie backends.

ph-kev (approved):

juliasloan25:
> Why is Oceananigans being made into a weak dep? Would ClimaOceans and ClimaLand be weak or strong deps?

I made Oceananigans a weak dep so we can still run simpler setups (slabplanet, AMIP) without requiring Oceananigans. I'm not yet sure about the other components, but one option would be to have an extension for each component, basically moving the files in experiments/ClimaEarth/components into different extensions. Then we can support running simple setups (the simplest being ClimaAtmos with a slab ocean, no land or ice) without loading extra packages.

For ClimaOcean at least, I would probably but Oceananigans and ClimaOcean together into the extension I added here.

I realized as I was working on this PR that I added ClimaAtmos as a strong dep. We could move it into an extension too, or since it's the only atmosphere we support right now we could just keep it as a strong dep.

PR #1630, opened 2025-12-19
Adding a new ocean setup

simone-silvestri:
This PR is adding a new ocean setup characterized by a 0.5 degree resolution ocean with 100 vertical levels.

The same setup has been evolved for 20 years coupled to a prescribed atmosphere after initialization from the ECCO4 climatology for January 1992.
At the moment the initial conditions are saved in a file loaded in dropbox.
I am not 100% sure we want to keep this setup as the latitude-logitude grid does not have the arctic, but if we want to keep it, I can open a PR in ClimaArtifacts to move the initialization there.


juliasloan25:
@simone-silvestri do we still want to merge this?

akshaysridhar:
cc @simone-silvestri Closing in favour of https://github.com/CliMA/ClimaCoupler.jl/pull/1971 - we can introduce any new setups in that rather than resolving merge issues with this PR.

PR #1632, opened 2025-12-22
Add MPI to env during benchmark pipeline init step

imreddyTeja:
Commit f7595753c6a2ee3d1f99a5204fc837d5d7530584 removed the step that adds CUDA and MPI to the benchmark env. The MPI add step is needed because it is not a direct dependency. If the ClimaAtmos compat does not have the latest downstream package versions as the lower bound for their compats, then this step may downgrade those packages.

## Content

----


juliasloan25:
> > Maybe we need to add back CUDA too? Or we could add them to the Manifests as direct dependencies
> 
> CUDA is already a direct dep, so it does not need to be added. Should I add MPI as well?

Sure - I think if we have it as a direct dep we don't have to have the `add` steps in the pipeline and can hopefully avoid the sneaky downgrade problem

PR #1634, opened 2026-01-04
fix some environment variable for the longrun

szy21:
## Content

----


szy21:
manually merging this in as it only changes longrun pipeline

PR #1643, opened 2026-01-06
add CMIP with integrated land to CI

juliasloan25:
## Purpose 
This makes sure we have CMIP runs in our regular CI that cover all combinations of sea ice (prescribed vs ClimaSeaIce) and land (bucket vs integrated) models. This will be helpful as we focus on CMIP runs more.


szy21 (approved):
I think maybe we don't need climaocean with prescribed sea ice, as I'm not sure if the ocean-sea ice flux is correct. I'll think about it.

PR #1645, opened 2026-01-07
don't use spun-up land ICs in CMIP

juliasloan25:
## Purpose 
In CMIP runs we put the land model at the poles because the ocean and sea ice run on capped latitude/longitude which only spans from 80°S to 80°N. As a result, we can't use previously-saved initial conditions for the land model in this setup, because they don't have reasonable values where the land wasn't run.

This PR switches all CMIP runs to use analytical ICs for the land instead, and documents this behavior.

With this change: [CMIP longrun](https://buildkite.com/clima/climacoupler-longruns/builds/1030) with integrated land fails at 14 weeks (same as bucket)
Without this change: [CMIP longrun](https://buildkite.com/clima/climacoupler-longruns/builds/1020) with integrated land fails at ~1 day

## To-do
- [x] update CMIP configs to not use spun up land ICs
- [x] update ClimaEarth readme to explain this behavior

## Content

----

### DIFF
--- a/config/ci_configs/cmip_oceananigans_climaseaice.yml
+++ b/config/ci_configs/cmip_oceananigans_climaseaice.yml
@@ -12,6 +12,7 @@ energy_check: false
 h_elem: 8
 ice_model: "clima_seaice"
 land_model: "integrated"
+land_spun_up_ic: false
 mode_name: "cmip"
 netcdf_output_at_levels: true
 output_default_diagnostics: true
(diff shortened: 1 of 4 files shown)


szy21 (approved):

kmdeck (approved):
looks fine to me! And I guess the code already exists elsewhere to ingest this flag?

One thought is that it would make more sense to put the prescribed ocean or sea ice at the north pole, instead of land.

Issue #1539, opened 2025-10-30
Add an option for binary area fraction


szy21:
Right now this is hardcoded for the land in [setup_run](https://github.com/CliMA/ClimaCoupler.jl/blob/796299298d600d3925b4ab4f5dad7d8964081a02/experiments/ClimaEarth/setup_run.jl#L276), and for sea ice/ocean [here](https://github.com/CliMA/ClimaCoupler.jl/blob/796299298d600d3925b4ab4f5dad7d8964081a02/experiments/ClimaEarth/components/ocean/prescr_seaice.jl#L180) and [here](https://github.com/CliMA/ClimaCoupler.jl/blob/796299298d600d3925b4ab4f5dad7d8964081a02/experiments/ClimaEarth/components/ocean/prescr_seaice.jl#L325). We could instead add an option in cli_options.jl, parse it in `get_coupler_args`, and use it to do this conditionally.

PR #1651, opened 2026-01-08
Make binary area treatment optional and support era5 land fraction

costachris:
Remove binary area fraction clippings and add support for era5 land fraction artifact.

### DIFF
--- a/experiments/ClimaEarth/components/ocean/prescr_seaice.jl
+++ b/experiments/ClimaEarth/components/ocean/prescr_seaice.jl
@@ -146,6 +146,7 @@ function PrescribedIceSimulation(
     land_fraction,
     stepper = CTS.RK4(),
     sic_path::Union{Nothing, String} = nothing,
+    binary_area_fraction::Bool = true,
 ) where {FT}
     # Set up prescribed sea ice concentration object
     sic_data =
@@ -176,8 +177,14 @@ function PrescribedIceSimulation(
     ice_fraction = CC.Fields.zeros(space)
     evaluate!(ice_fraction, SIC_timevaryinginput, tspan[1])
 
-    # Make ice fraction binary rather than fractional
-    ice_fraction = ifelse.(ice_fraction .> FT(0.5), FT(1), FT(0))
+    # Ensure ice fraction is finite and not NaN
+    @. ice_fraction = ifelse(isfinite(ice_fraction), ice_fraction, FT(0))
+    # binary ice fraction (threshold at 0.5)
+    if binary_area_fraction
+        @. ice_fraction = ifelse(ice_fraction > FT(0.5), FT(1), FT(0))
+    end
+    # max/min needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
+    @. ice_fraction = max(min(ice_fraction, FT(1) - land_fraction), FT(0))
 
     params = IceSlabParameters{FT}(coupled_param_dict)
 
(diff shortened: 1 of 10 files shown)


experiments/ClimaEarth/components/ocean/prescr_seaice.jl
```diff
@@ -176,8 +176,11 @@ function PrescribedIceSimulation(
     ice_fraction = CC.Fields.zeros(space)
     evaluate!(ice_fraction, SIC_timevaryinginput, tspan[1])
 
-    # Make ice fraction binary rather than fractional
-    ice_fraction = ifelse.(ice_fraction .> FT(0.5), FT(1), FT(0))
+    # Keep ice fraction *fractional* (no binary thresholding), but ensure it is finite
+    # and consistent with the static land fraction.
+    @. ice_fraction = ifelse(isfinite(ice_fraction), ice_fraction, FT(0))
+    # max/min needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
```
costachris: @juliasloan25 I was following the pattern in a couple other places re: the " Heisenbug on HPC" issue. Is that still relevant here?

costachris:
> If we use non-binary area fractions the fluxes will be wrong because we only store bulk fluxes per grid cell in the coupler (i.e. the area fraction-weighted sum across the present surface models). To enable fractional area fractions we need to store the flux between the atmosphere and each surface, then pass these individual fluxes to the surfaces.
> 
> I actually just fleshed out this issue with some details on this because it came up today: #1539
> 
> I don't think this is difficult to change so we can prioritize it if it will be helpful

Gotcha. In that case I can add back the 0.5 clipping for sea ice and leave the support for era5 artifact, but just binary clip it at 0.5? (it has the benefit of i.e. containing major inland seas and lakes, at least). That should preserve the current behavior since the landsea_mask_60arcseconds is already binary in the artifact. Then the changes needed for proper non-binary area fraction can be handled in a separate PR?

juliasloan25:
> > If we use non-binary area fractions the fluxes will be wrong because we only store bulk fluxes per grid cell in the coupler (i.e. the area fraction-weighted sum across the present surface models). To enable fractional area fractions we need to store the flux between the atmosphere and each surface, then pass these individual fluxes to the surfaces.
> > I actually just fleshed out this issue with some details on this because it came up today: #1539
> > I don't think this is difficult to change so we can prioritize it if it will be helpful
> 
> Gotcha. In that case I can add back the 0.5 clipping for sea ice and leave the support for era5 artifact, but just binary clip it at 0.5? (it has the benefit of i.e. containing major inland seas and lakes, at least). That should preserve the current behavior since the landsea_mask_60arcseconds is already binary in the artifact. Then the changes needed for proper non-binary area fraction can be handled in a separate PR?

that sounds good to me!

experiments/ClimaEarth/components/ocean/prescr_seaice.jl
```diff
@@ -176,8 +177,14 @@ function PrescribedIceSimulation(
     ice_fraction = CC.Fields.zeros(space)
     evaluate!(ice_fraction, SIC_timevaryinginput, tspan[1])
 
-    # Make ice fraction binary rather than fractional
-    ice_fraction = ifelse.(ice_fraction .> FT(0.5), FT(1), FT(0))
+    # Ensure ice fraction is finite
+    @. ice_fraction = ifelse(isfinite(ice_fraction), ice_fraction, FT(0))
```
juliasloan25: why do we need to check if ice_fraction is finite?

src/Input.jl
```diff
@@ -225,6 +229,14 @@ function argparse_settings()
         help = "Sea ice model to use. [`prescribed` (default), `clima_seaice`]"
         arg_type = String
         default = "prescribed"
+        "--land_fraction_source"
```
juliasloan25: can you add these to one of the tables in docs/src/Input.md too please?

juliasloan25 (approved):

PR #1652, opened 2026-01-08
convert momentum fluxes to Cartesian basis for Oceananigans

juliasloan25:
## Purpose 
In CMIP runs, we noticed grid imprinting on ocean velocities. This suggests that the momentum fluxes are not being correctly rotated when they're provided to the ocean (and sea ice). We have a function `_rotate_vector!` which calls `OC.Operators.intrinsic_vector`, but it doesn't seem like this would support the ClimaCore contravariant basis as an extrinsic basis ([docs](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Operators.intrinsic_vector-Tuple%7BAny,%20Any,%20Any,%20Oceananigans.Grids.AbstractGrid,%20Any,%20Any,%20Any%7D)).

`intrinsic_vector` does support rotating from Cartesian coordinates to the Oceananigans coordinates (which in the case of a LatitudeLongitudeGrid are the same, but will not be the same for TripolarGrid). This PR converts the momentum fluxes from the contravariant basis to a Cartesian basis before rotating them onto the Oceananigans grid.

### Before this PR
From [this build](https://buildkite.com/clima/climacoupler-ci/builds/7550#019b9eee-2eaf-49ee-b491-b6a29f4d0457): CMIP + integrated land + ClimaSeaIce
Screenshot 2026-01-09 at 11 07 38 AM

### After this PR
From [this build](https://buildkite.com/clima/climacoupler-ci/builds/7564#019ba353-2e3b-4cf9-aff4-4ba8c5ce82e8)
Screenshot 2026-01-09 at 11 19 43 AM

## Content
- [x] Add a function to convert from contravariant basis to UVVector (Cartesian basis)
- [x] Use this function when providing momentum fluxes to Oceananigans and ClimaSeaIce models
- [ ] Add temp fields to avoid allocations

----

### DIFF
--- a/experiments/ClimaEarth/components/ocean/climaocean_helpers.jl
+++ b/experiments/ClimaEarth/components/ocean/climaocean_helpers.jl
@@ -147,3 +147,30 @@ Fields to Face/Center and Center/Face coordinates, respectively.
         τy[i, j, 1] = OC.Operators.ℑyᵃᶠᵃ(i, j, 1, grid, τy_cc)
     end
 end
+
+"""
+    contravariant_to_cartesian!(ρτ_flux_uv, ρτxz, ρτyz)
+
+Convert the covariant vector components `ρτxz` and `ρτyz` from the
+contravariant basis (as they are output by the surface flux calculation)
+to the Cartesian basis. These are now in an extrinsic coordinate system
+that can be rotated onto the ocean/sea ice grid by `_rotate_vector!`.
+"""
+function contravariant_to_cartesian!(ρτ_flux_uv, ρτxz, ρτyz)
+    # Get the local geometry of the boundary space
+    local_geometry = CC.Fields.local_geometry_field(ρτxz)
+
+    # Get the vector components in the CT1 and CT2 directions
+    xz = @. CA.CT12(
+        CA.CT1(CA.unit_basis_vector_data(CA.CT1, local_geometry)),
+        local_geometry,
+    )
+    yz = @. CA.CT12(
+        CA.CT2(CA.unit_basis_vector_data(CA.CT2, local_geometry)),
+        local_geometry,
+    )
+
+    # Convert the vector components to a UVVector on the Cartesian basis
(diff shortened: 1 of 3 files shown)


experiments/ClimaEarth/components/ocean/climaocean_helpers.jl
```diff
@@ -147,3 +147,35 @@ Fields to Face/Center and Center/Face coordinates, respectively.
...
+    contravariant_to_cartesian(ρτxz, ρτyz)
+
+Convert the contravariant tensor components `ρτxz` and `ρτyz`, which are
+output by the surface flux calculation, to UVVector components.
+These are now in a Cartesian coordinate system, which is an extrinsic coordinate system
+that can be rotated onto the ocean/sea ice grid by `_rotate_vector!`.
+"""
+function contravariant_to_cartesian(ρτxz, ρτyz)
+    # Get the local geometry of the boundary space
+    # TODO broadcasting fails in tensor_from_components so we need to make local_geometry a Field
```
juliasloan25: @szy21 @dennisYatunin this seemed suspicious to me but if I just took `local_geometry = boundary_space.grid.local_geometry` I got an error in line 164 because `broadcastable` isn't defined for the `LocalGeometry` object. Have you encountered that before?

simone-silvestri:
much better!

experiments/ClimaEarth/components/ocean/climaocean_helpers.jl
```diff
@@ -147,3 +147,30 @@ Fields to Face/Center and Center/Face coordinates, respectively.
         τy[i, j, 1] = OC.Operators.ℑyᵃᶠᵃ(i, j, 1, grid, τy_cc)
     end
 end
+
+"""
+    contravariant_to_cartesian!(ρτ_flux_uv, ρτxz, ρτyz)
+
+Convert the contravariant tensor components `ρτxz` and `ρτyz`, which are
```
szy21: I had a previous comment about this here: https://github.com/CliMA/ClimaCoupler.jl/pull/1669#discussion_r2700270516, but I'm fine with the naming / docstring if it makes more sense to you.

PR #1658, opened 2026-01-10
wxquest v3

costachris:
Update configurations for weatherquest runs. Add functionality for running Coupler through ERA5 historical period.

Includes options for: 
1.  running bucket with prescribed ERA5 albedo
2. running integrated land with MODIS climatology (to run coupler before 2010)
3. initializing sea ice with ERA5 surface ice temperatures

### DIFF
--- a/config/subseasonal_configs/wxquest_progedmf.yml
+++ b/config/subseasonal_configs/wxquest_progedmf.yml
@@ -2,12 +2,14 @@ FLOAT_TYPE: "Float32"
 albedo_model: "CouplerAlbedo"
 atmos_config_file: "config/atmos_configs/climaatmos_wx_progedmf.yml"
 coupler_toml: ["toml/amip_progedmf.toml"]
+strict_params: false
 mode_name: "subseasonal"
 era5_initial_condition_dir: "/net/sampo/data1/wxquest_data/initial_conditions"
 initial_condition: "WeatherModel"
 checkpoint_dt: "7days"
 dt: "10secs"
 dt_cpl: "10secs"
+dt_seaice: "10secs"
 energy_check: false
 h_elem: 32
 land_fraction_source: "era5"
@@ -15,23 +17,25 @@ binary_area_fraction: false
 
 ### integrated land ###
 # land_model: "integrated"
+# lai_source: "modis_monthly_climatology"
 
 ### bucket land ###
 land_model: "bucket"
 bucket_albedo_type: "map_temporal"
-bucket_initial_condition: "/net/sampo/data1/wxquest_data/initial_conditions/era5_bucket_processed_20250907_0000.nc"
+bucket_initial_condition: "/net/sampo/data1/wxquest_data/initial_conditions/era5_bucket_processed_20250831_0000.nc"
 
 land_spun_up_ic: false
 land_temperature_anomaly: "nothing"
 use_land_diagnostics: true
 radiation_reset_rng_seed: true
-start_date: "20250907"
+start_date: "20250831"
 surface_setup: "PrescribedSurface"
 topo_smoothing: true
 topography: "Earth"
 t_end: "7days"
 netcdf_output_at_levels: true
+use_coupler_diagnostics: true
 extra_atmos_diagnostics:
-  - short_name: [tas, mslp, pr, ua, va, rhoa, pfull, hur, hus, clw, cli, clivi, clwvi, cl, arup, tke]
-    period: 1hours
+  - short_name: [zg, orog, tas, ta, mslp, pr, ua, va, ts, rhoa, pfull, hur, hus, clw, cli, clivi, clwvi, cl, arup, tke, hfss, hfls, evspsbl, hussfc, rsds, rlds, rlus, rsus, rld, wa, waup]
+    period: 3hours
     reduction_time: average
--- a/experiments/ClimaEarth/README.md
+++ b/experiments/ClimaEarth/README.md
@@ -160,11 +160,13 @@ Generates 3-4 week forecasts from ERA5-derived initial conditions. Similar to `a
 - `era5_initial_condition_dir`: directory that contains files named like:
   - `sst_processed_YYYYMMDD_0000.nc`
   - `sic_processed_YYYYMMDD_0000.nc`
-  - `era5_land_processed_YYYYMMDD_0000.nc`
-  - `era5_bucket_processed_YYYYMMDD_0000.nc`
+  - `era5_land_processed_YYYYMMDD_0000.nc` (for integrated land model)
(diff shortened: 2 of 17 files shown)


config/subseasonal_configs/wxquest_progedmf.yml
```diff
@@ -19,19 +21,20 @@ binary_area_fraction: false
...
 topography: "Earth"
 t_end: "7days"
 netcdf_output_at_levels: true
+use_coupler_diagnostics: true
 extra_atmos_diagnostics:
-  - short_name: [tas, mslp, pr, ua, va, rhoa, pfull, hur, hus, clw, cli, clivi, clwvi, cl, arup, tke]
-    period: 1hours
+  - short_name: [zg, orog, tas, ta, mslp, pr, ua, va, ts, rhoa, pfull, hur, hus, clw, cli, clivi, clwvi, cl, arup, tke, hfss, hfls, evspsbl, hussfc, rsds, rlds, rlus, rsus, rld, wa, waup]
+    period: 3hours
     reduction_time: average
```
szy21: add a line?

src/Models/prescr_seaice.jl
```diff
@@ -345,19 +452,26 @@ function ice_rhs!(dY, Y, p, t)
     #  max needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
     @. p.area_fraction = max(min(p.area_fraction, FT(1) - p.land_fraction), FT(0))
 
-    # Calculate the conductive flux, and set it to zero if the area fraction is zero
-    F_conductive = @. k_ice / (h) * (T_base - ice_surface_temperature(Y.T_bulk, T_base)) # fluxes are defined to be positive when upward
+    # Calculate the surface temperature
+    T_sfc = @. clamp(ice_surface_temperature(Y.T_bulk, T_base), T_sfc_min, T_freeze)
```
szy21: maybe just clamp in the ice_surface_temperature function?

experiments/ClimaEarth/README.md
```diff
@@ -160,11 +160,13 @@ Generates 3-4 week forecasts from ERA5-derived initial conditions. Similar to `a
 - `era5_initial_condition_dir`: directory that contains files named like:
```
juliasloan25: If these files don't change too often, it would be better to put them in ClimaArtifacts rather than requiring that users set up a folder with this specific structure. But if they're still changing a lot right now it's ok

experiments/ClimaEarth/README.md
```diff
@@ -195,6 +197,12 @@ Generates 3-4 week forecasts from ERA5-derived initial conditions. Similar to `a
       - `skt` (K): skin temperature, dims `(lat, lon)`
     - era5 land level midpoints: 0.035, 0.175, 0.64, 1.945 (m)
 
+  - `albedo_processed_YYYYMMDD_0000.nc` (optional, for ERA5-derived albedo):
```
juliasloan25: is this defined over the whole surface and just used for the bucket, or was a land mask used to generate it? it would be nice to clarify a little

experiments/ClimaEarth/setup_run.jl
```diff
@@ -284,11 +286,23 @@ function CoupledSimulation(config_dict::AbstractDict)
...
     # Determine whether to use a shared surface space
     shared_surface_space = share_surface_space ? boundary_space : nothing
+    # Use subseasonal bucket IC if bucket_initial_condition is not specified
+    bucket_ic = if isempty(bucket_initial_condition) && !isnothing(subseasonal_bucket_ic)
+        subseasonal_bucket_ic
+    else
+        bucket_initial_condition
+    end
+    # Use subseasonal albedo file if available
+    era5_albedo_file = isnothing(subseasonal_albedo) ? "" : subseasonal_albedo
```
juliasloan25: Since this is land model-specific, can we just pass `subseasonal_albedo` to the constructor and do this conversion internally to there?

src/Models/prescr_seaice.jl
```diff
@@ -105,6 +111,63 @@ function slab_ice_space_init(::Type{FT}, space, params) where {FT}
...
+    space,
+    params,
+    sic_data,
+    start_date,
+    t_start,
+    ice_fraction,
+) where {FT}
+    T_fallback = params.T_freeze - FT(10.0)
+    T_base = params.T_base
+    T_freeze = params.T_freeze
```
juliasloan25:
```suggestion
    (; T_base, T_freeze) = params
    T_fallback = T_freeze - FT(10.0)
```

src/Models/prescr_seaice.jl
```diff
@@ -198,7 +261,21 @@ function PrescribedIceSimulation(
...
+            boundary_space,
+            params,
+            sic_data,
+            start_date,
+            tspan[1],
+            ice_fraction,
+        )
+    else
+        slab_ice_space_init(FT, boundary_space, params)
+    end
```
juliasloan25: What if we just call like this:
```suggestion
    Y = slab_ice_space_init(
            FT,
            boundary_space,
            params,
            sic_data,
            start_date,
            tspan[1],
            ice_fraction,
        )
```
and the `slab_ice_space_init_from_file` function you added can instead be a method of `slab_ice_space_init`. We can dispatch on `sic_data::Nothing` for the existing behavior

src/Models/prescr_seaice.jl
```diff
@@ -345,13 +434,17 @@ function ice_rhs!(dY, Y, p, t)
     #  max needed to avoid Float32 errors (see issue #271; Heisenbug on HPC)
     @. p.area_fraction = max(min(p.area_fraction, FT(1) - p.land_fraction), FT(0))
 
-    # Calculate the conductive flux, and set it to zero if the area fraction is zero
-    F_conductive = @. k_ice / (h) * (T_base - ice_surface_temperature(Y.T_bulk, T_base)) # fluxes are defined to be positive when upward
+    # Calculate the surface temperature (clamped to [T_sfc_min, T_freeze] inside ice_surface_temperature)
+    (; T_sfc_min) = p.params
+    T_sfc = @. ice_surface_temperature(Y.T_bulk, T_base, T_sfc_min, T_freeze)
```
juliasloan25: This is allocating. This model isn't super performance-critical, but still we should try to avoid allocations when we can.

We could either:
- Allocate space for T_sfc in the sea ice cache
- Make this a lazy broadcast (and add `using LazyBroadcast: lazy`):
```suggestion
    T_sfc = @. lazy(ice_surface_temperature(Y.T_bulk, T_base, T_sfc_min, T_freeze))
```

This applies to LW_emission too - maybe we can just add that back into the `rhs = @. ...` expression?

experiments/ClimaEarth/README.md
costachris: It's still changing a bit and we may modify some of these more (e.g. remove or use different ERA fields), but when that's settled we can put them in artifacts.

PR #1669, opened 2026-01-16
fix ocean momentum flux units, sea ice radiative emission

juliasloan25:
## Purpose 
This PR makes 1 fix to ocean momentum fluxes:
- divide momentum fluxes by rho0

and 1 fix to sea ice radiative emission:
- add `ConcentrationMaskedRadiativeEmission` so that we don't have radiative emission from sea ice where there is no sea ice (this was causing sea ice formation everywhere before)

## To-do
- [x] momentum flux fixes x1
- [x] sea ice radiative emission fix x1
- [x] add more Oceananigans diagnostics

### DIFF
--- a/experiments/ClimaEarth/components/ocean/clima_seaice.jl
+++ b/experiments/ClimaEarth/components/ocean/clima_seaice.jl
@@ -39,6 +39,50 @@ struct ClimaSeaIceSimulation{SIM, A, REMAP, NT, IP} <: Interfacer.SeaIceModelSim
 end
 
 """
+    ConcentrationMaskedRadiativeEmission
+
+A heat boundary condition that emits radiation only where the sea ice
+concentration is greater than zero. This is needed to prevent radiative
+emission where we have no sea ice.
+"""
+struct ConcentrationMaskedRadiativeEmission{FT}
+    emissivity::FT
+    stefan_boltzmann_constant::FT
+    reference_temperature::FT
+end
+
+function ConcentrationMaskedRadiativeEmission(
+    FT = Float64;
+    emissivity = 1,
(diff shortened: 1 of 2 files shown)


experiments/ClimaEarth/components/ocean/clima_seaice.jl
```diff
@@ -38,6 +38,44 @@ struct ClimaSeaIceSimulation{SIM, A, REMAP, NT, IP} <: Interfacer.SeaIceModelSim
...
+    stefan_boltzmann_constant::FT
+    reference_temperature::FT
+end
+
+
+function ConcentrationMaskedRadiativeEmission(
+    FT = Float64;
+    emissivity = 1,
+    stefan_boltzmann_constant = 5.67e-8,
+    reference_temperature = 273.15,
```
szy21: I guess we don't have access to ClimaParams here?

experiments/ClimaEarth/components/ocean/climaocean_helpers.jl
```diff
@@ -147,3 +147,35 @@ Fields to Face/Center and Center/Face coordinates, respectively.
...
+
+"""
+    contravariant_to_cartesian(ρτxz, ρτyz)
+
+Convert the contravariant tensor components `ρτxz` and `ρτyz`, which are
+output by the surface flux calculation, to UVVector components.
+These are now in a Cartesian coordinate system, which is an extrinsic coordinate system
+that can be rotated onto the ocean/sea ice grid by `_rotate_vector!`.
+"""
+function contravariant_to_cartesian(ρτxz, ρτyz)
```
juliasloan25: make this operate in-place

PR #1675, opened 2026-01-20
pin GPUCompiler < 1.7.6

juliasloan25:
## Purpose 
Temporary fix to prevent [this error](https://buildkite.com/clima/climacoupler-ci/builds/7640#019bdc9a-93d2-407a-a3e0-3f46d41732e6) that comes up when using GPUCompiler 1.7.6 or 1.8.0. It would be better to fix this in ClimaCore, but until that PR is merged we can use this as a temporary fix https://github.com/CliMA/ClimaCore.jl/pull/2430


szy21 (approved):
Thanks! Do you want to make Manifest.toml consistent with Manifest-v1.11.toml?

PR #1676, opened 2026-01-21
add SimCoordinator module

### LINKED ISSUE #1610: Move functions from `experiments/` to `src/`
We should minimize the code that's in `experiments/` and instead have the bulk of our package in `src/`. For now we'll skip the component models, since that will be a large change on its own.

#### Checkpointer module: #1612
- [x] experiments/ClimaEarth/components/shared/restore.jl --> Checkpointer

#### Input module: #1613
- [x] experiments/ClimaEarth/user_io/arg_parsing.jl --> Input
- [x] experiments/ClimaEarth/cli_options.jl -> Input

#### SimOutput module #1617 
- [x] experiments/ClimaEarth/user_io/coupler_diagnostics.jl --> SimOutput
- [x] experiments/ClimaEarth/user_io/benchmarks.jl --> SimOutput

#### Plotting extension #1621
- [x] experiments/ClimaEarth/user_io/diagnostics_plots.jl
- [x] experiments/ClimaEarth/user_io/debug_plots.jl
- [x] experiments/ClimaEarth/leaderboard/ (data_sources.jl mostly dependent on component model-specific diagnostic names, but I think this is okay)
- [x] ~~experiments/calibration/run_calibration.jl: scatter_plot, param_versus_iter_plot (these depend on EKP too)~~
- [x] experiments/ClimaEarth/leaderboard/: test_rmses.jl (and functions to get sim, obs dicts) --> SimOutput

#### SimCoordinator module; move postprocess to Plotting module #1676
- [x] experiments/ClimaEarth/setup_run.jl: run!(cs), step!(cs) --> SimCoordinator
- [x] experiments/ClimaEarth/user_io/postprocessing.jl
- [x] experiments/ClimaEarth/setup_run.jl: postprocess(cs)
  - some non-plotting postprocessing is done, but we need access to Plotting module and I think it's similar enough
  - can then move benchmarks.jl out of user_io folder

#### Add extensions for each component model
- So that source remains independent of heavy packages (e.g. Oceananigans/ClimaOcean), but we don't have everything in experiments
- Simple models that are defined in ClimaCoupler (slab ocean, prescribed ocean, prescribed sea ice) can go into source without adding dependencies

#### Move `CoupledSimulation` constructors from setup_run.jl to SimCoordinator
- Right now these depend on component model-specific constructors (e.g. BucketSimulation). We should define stubs (e.g. `LandSimulation`) which get extended by the respective component model extensions

#### Move `setup_and_run.jl` from setup_run.jl to SimCoordinator
- Depends on `CoupledSimulation` constructors

juliasloan25:
## Purpose 
Add a module to contain the `step!` and `run!` functions. I'm choosing not to add unit tests because setting up a CoupledSimulation object takes a long time and these functions are thoroughly tested in the integration tests.

Also moves `postprocess(cs)` into the Plotting module. This function doesn't entirely do plotting but it mostly does so I think this is fine.

closes #1610

## Content
- [x] add SimCoordinator module
- [x] add SimCoordinator docs page
- [x] update ClimaCoupler module exports for usability
- [x] merge `postprocess_sim`, `postprocess`, and move `postprocess` into Plotting module
- [x] remove `experiments/ClimaEarth/user_io/` folder
- [x] confirm buildkite results are identical to main

## Next steps
Now very few functions remain in setup_run.jl, and they all depend on code defined in the component model files within `experiments/ClimaEarth/components/`. The next step will be to move the component model code into extensions, and then move the last setup_run.jl functions into src.


src/SimCoordinator.jl
```diff
@@ -0,0 +1,162 @@
...
+Keyword arguments
+==================
+
+`precompile`: If `true`, run the coupled simulations for two steps, so that most functions
+              are precompiled and subsequent timing will be more accurate.
+"""
+function run!(
+    cs::Interfacer.CoupledSimulation;
+    precompile = (cs.tspan[end] > 2 * cs.Δt_cpl + cs.tspan[begin]),
+)
```
ph-kev:
```suggestion
    precompile::Bool = cs.tspan[end] > 2 * cs.Δt_cpl + cs.tspan[begin],
)
```

Issue #1682, opened 2026-01-22
Move component model code into extensions


juliasloan25:
Right now, the experiments/ClimaEarth/components/ folder contains code that is specific to component models. We want to reduce code in experiments/, but also avoid having component model-specific code in src/. E.g. if a user wants to run AMIP, they shouldn't have to load Oceananigans (which is only needed for CMIP). This idea applies to other component model types as we want to be able to support a range of model complexities for each component.

For each component model that comes from an external package, we should make an extension. These extensions should extend the interface functions defined in ClimaCoupler src (these should already be set up/extended correctly).

ClimaCoupler also implements some simple surface component models. These can just be moved into src since they won't bring in any extra dependencies.

Part of SDI #1609

### Make generic interface 
- Define stubs for component model constructors (e.g. `LandSimulation`, `OceanSimulation`) which will be extended in extensions. This should already be done for most other functions we need (e.g. `get_field`, `update_sim!`, etc).
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1683

### Move simple models into src/ 
- slab ocean, prescribed ocean, prescribed sea ice
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1692

### Move external models into ext/
- ClimaLand extension (for both bucket and integrated land) [Kevin]
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1706
- ClimaAtmos extension [Kevin]
  - Right now we use ClimaAtmos in src when we get the coupler config file. We should disentangle these so we don't need to rely on ClimaAtmos in src.
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1721
- CMIP extension (will require Oceananigans, ClimaOcean, ClimaSeaIce, KernelAbstractions) [Julia] https://github.com/CliMA/ClimaCoupler.jl/pull/1704

### Update documentation
- explain that ClimaCoupler provides simple models for e.g. aquaplanet; extensions must be loaded for more complex setups (AMIP, CMIP, etc)
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1760

### Move setup_run.jl into src/
- After these changes, `setup_run.jl` is now mostly two `CoupledSimulation` constructors and two `setup_and_run` methods. Once we move the component models into src/ext we should be able to also move these into src/ which will be much cleaner
  - https://github.com/CliMA/ClimaCoupler.jl/pull/1759

[...]

PR #1692, opened 2026-01-29
move simple models to extension

juliasloan25:
## Purpose 
Now that all component models strictly extend the interface defined by ClimaCoupler.jl, we can start moving them into extensions. First I'm moving the 3 simple, coupler-defined models into src. The set of packages they bring in is fairly random so we decided to do this over putting them in an extension.

Part of #1682

## Content
- [x] move prescribed ocean, slab ocean, prescribed sea ice to new src/Models/
- [x] moved tests from experiments/ClimaEarth/test/ to test/models/ and added them to test/runtests.jl
- [x] added documentation page docs/src/models.md
- [x] moved historical_sst_sic artifact to top-level ClimaCoupler.jl Artifacts.toml
- [x] confirm that buildkite results are unchanged

----


docs/src/models.md
```diff
@@ -0,0 +1,127 @@
...
+temperature evolution based on the net energy flux into the ocean. The ocean
+surface temperature is used to compute upwelling longwave radiation and surface
+fluxes.
+
+**Usage**: The slab ocean model is used in:
+- `SlabplanetMode` simulations
+- `SlabplanetAquaMode` simulations
+
+The slab ocean model should not be used with a sea ice model, as it does not
+account for ice-ocean interactions.
```
ph-kev: I am just wondering, but is this enforced in code or is the user supposed to know to not do this?

PR #1693, opened 2026-01-30
update to ClimaAtmos v0.35 and ClimaLand v1.5

szy21:
## Purpose 

- Update to the latest ClimaAtmos and ClimaLand
- Update to the latest Insolation
- Update to the latest Thermodynamics, remove thermo state

## To-do

- [ ] Update Manifest
- [x] Remove thermo state in test

## Content

----


experiments/ClimaEarth/components/atmosphere/climaatmos.jl
```diff
@@ -373,18 +373,8 @@ function Interfacer.update_field!(
...
-        sim.integrator.p.precomputed.sfc_conditions.ts .=
-            TD.PhaseNonEquil_ρTq.(
-                thermo_params,
-                ρ_sfc_atmos,
-                T_sfc_atmos,
-                TD.PhasePartition.(q_sfc_atmos),
-            )
-    end
+    sim.integrator.p.precomputed.sfc_conditions.T_sfc .= T_sfc_atmos
+    sim.integrator.p.precomputed.sfc_conditions.q_vap_sfc .= q_sfc_atmos
```
juliasloan25: I thought that now that we aren't using thermo states, this method can just become:
```
Interfacer.update_field!(sim::ClimaAtmosSimulation, ::Val{:surface_temperature}, T_sfc)
    Interfacer.remap!(sim.integrator.p.precomputed.sfc_conditions.T_sfc, T_sfc_atmos)
end
```
(note we can just pass in one field T_sfc instead of all the coupler fields)

I guess the surface humidity is more complicated because we need to compute it rather than just remapping, but that can be a separate method I think.

juliasloan25 (commented):
Looks good so far! I left some initial comments

PR #1699, opened 2026-02-02
Use tagged version of ClimaAtmos and ClimaLand

imreddyTeja:
I forgot to do this in the last PR
## Content

----


szy21 (approved):
Thanks! We are going to release ClimaLand 1.5.2 soon and it has a fix to 1.5.1, so you can wait for that too.

PR #1706, opened 2026-02-04
Add ClimaCouplerClimaLandExt extension

ph-kev:
This PR makes an extension for ClimaLand which involves moving all the files in `experiments/ClimaEarth/components/land` to `ext`.

TODO

- [x] See which tests need the ClimaLand extension loaded and add `using` statement if needed


Project.toml
```diff
@@ -30,16 +30,19 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
...
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 Poppler_jll = "759029c8-65f6-5f13-b93f-2d7e0a4cb77c"
 Printf = "366bfd08-7218-5ad0-9036-d27ba1f5fe54"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [extensions]
 ClimaCouplerCMIPExt = ["ClimaOcean", "ClimaSeaIce", "KernelAbstractions", "Oceananigans"]
+ClimaCouplerClimaLandExt = ["ClimaLand", "NCDatasets", "Statistics"]
```
juliasloan25: It would be nice if this extension was loaded by including only ClimaLand (we can probably improve this for the other extensions too, especially the Makie extensions).

It looks like Statistics is loaded in climaland_bucket.jl but not used. Could you remove it in that file and here? NCDatasets is needed only for the bucket IC reading. It's ok to keep it for now, but let's see if we can work towards removing that in a separate PR.

PR #1711, opened 2026-02-05
Fix bug with missing imports

ph-kev:
Nightly AMIP is currently failing because `ClimaDiagnostics` is not imported when using `pkgversion` to check the version of ClimaDiagnostics. This PR adds the necessary imports and rename `ClimaDiagnostics` to `CD` in some cases.


ext/ClimaCouplerMakieExt/leaderboard/data_sources.jl
```diff
@@ -1,5 +1,6 @@
 import ClimaAnalysis
 import ClimaUtilities.ClimaArtifacts: @clima_artifact
+import ClimaDiagnostics
```
juliasloan25:
```suggestion
import ClimaDiagnostics as CD
```
for consistency?

PR #1721, opened 2026-02-10
Add ClimaCouplerClimaAtmosExt extension

ph-kev:
This PR makes an extension for ClimaAtmos which involves moving all the files in `experiments/ClimaEarth/components/atmosphere` to ext.

### DIFF
--- a/experiments/ClimaEarth/setup_run.jl
+++ b/experiments/ClimaEarth/setup_run.jl
@@ -60,19 +60,11 @@ using Makie, GeoMakie, CairoMakie, ClimaCoreMakie, NCDatasets, Poppler_jll
 # Note we only need these if running CMIP, but for now we share one environment for all experiments
 import Oceananigans, ClimaOcean, ClimaSeaIce, KernelAbstractions
 
-#  Trigger ClimaCouplerClimaLandExt extension
+# Trigger ClimaCouplerClimaLandExt extension
 import ClimaLand, NCDatasets
 
-pkg_dir = pkgdir(ClimaCoupler)
-
-#=
-### Helper Functions
-These will be eventually moved to their respective component model and utility packages, and so they should not
-contain any internals of the ClimaCoupler source code, except extensions to the Interfacer functions.
-=#
-
-## helpers for component models
-include("components/atmosphere/climaatmos.jl")
+# Trigger ClimaCouplerClimaAtmosExt
+import ClimaAtmos
 
 #=
 ### Configuration Dictionaries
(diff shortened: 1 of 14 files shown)


experiments/ClimaEarth/setup_run.jl
```diff
@@ -180,6 +159,7 @@ function CoupledSimulation(config_dict::AbstractDict)
         isnothing(restart_dir) && (restart_dir = dir_paths.checkpoints_dir)
     end
     should_restart = !isnothing(restart_t) && !isnothing(restart_dir)
+    t_start_int = nothing
     if should_restart
```
ph-kev: Should this part be moved into `get_coupler_config_dict`?

juliasloan25: Yeah that would be better. We can also put it into a helper function that's called from `get_coupler_config_dict` if you think that would help keep things organized too

juliasloan25: If we put this in `get_coupler_config_dict` we can also then just update `t_start` in the config file directly so we don't need to pass `t_start_int` to the atmos constructor. It looks like how the code is now, `t_start_int` is always nothing and never overwritten

experiments/ClimaEarth/test/restart.jl
```diff
@@ -62,6 +62,7 @@ four_steps_reading["detect_restart_files"] = true
 four_steps_reading["restart_dir"] = cs_four_steps.dir_paths.checkpoints_dir
 four_steps_reading["restart_t"] = 720
 four_steps_reading["job_id"] = "four_steps_reading"
+Input.update_t_start!(four_steps_reading)
```
ph-kev: I wasn't sure about this. The alternative is making a config file at runtime and passing it to `get_coupler_config_dict` so that `t_start` is updated.

experiments/ClimaEarth/setup_run.jl
```diff
@@ -471,6 +439,12 @@ function CoupledSimulation(config_dict::AbstractDict)
     When the caches are not read from the restart file, we have to perform the initial component
     model exchange so that `set_caches!` can be called to initialize the caches.
     =#
+    if detect_restart_files
+        isnothing(restart_t) &&
+            (restart_t = Checkpointer.t_start_from_checkpoint(dir_paths.checkpoints_dir))
+        isnothing(restart_dir) && (restart_dir = dir_paths.checkpoints_dir)
+    end
+    should_restart = !isnothing(restart_t) && !isnothing(restart_dir)
```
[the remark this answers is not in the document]
ph-kev: I realize that I can't do this because `restart_t` and `restart_dir` are updated in this function and `dir_paths` doesn't exist before the calls to `get_coupler_config_dict` and `get_coupler_args`.

src/Input.jl
```diff
@@ -380,7 +416,10 @@ function get_coupler_args(config_dict::Dict)
     if use_itime
         t_end = ITime(t_end, epoch = start_date)
         t_start = ITime(t_start, epoch = start_date)
-        Δt_cpl = ITime(Δt_cpl, epoch = start_date)
+        # Restarts can silently fail here because ITime automatically choose the
+        # appropriate period based on time value. The result ITime(0) gives a
+        # period of 1 second, but ITime(120) gives a period of 1 minute.
```
juliasloan25: Is it possible to handle the case that would silently fail?

juliasloan25 (approved):

PR #1723, opened 2026-02-10
Concurrent and overlapped component model stepping

### LINKED ISSUE #1700: support concurrent coupling step
Right now, we only support sequential component model stepping. We may see improved performance if we switch to concurrent/parallel component stepping.

Note that because of the land/atmosphere ordering restriction, and the comparative costs of each component, this will probably only result in speedup of CMIP, not AMIP.

## Restrictions
Currently, the integrated land model computes fluxes implicitly as part of its step (whereas for other surface models, surface fluxes are computed explicitly before each step). This imposes an ordering in the `step_model_sims!` function and prevents parallelization of the atmosphere and land steps. This is conveyed by labeling the `ClimaLandSimulation` as a subtype of the `AbstractImplicitFluxSimulation` abstract type.

We will eventually have the same restriction for sea ice but don't yet.

## Approach
- Link atmosphere and land in a sub-stepped model, since their timesteps are similar and they can't be stepped in parallel
- Instead of the for loop in `step_model_sims!`, we should spawn 3 threads: `@spawn atmos/land step`, `@spawn ocean step`, and `@spawn ice step`
- We should add a flag to the config options so we can choose which timestepping scheme we want to use - e.g. `coupler_timestepping` with options `explicit` or `concurrent`
- Test out performance gains by comparing CMIP SYPD between the two timestepping options, given the same compute resources

daverumph:
## Purpose 
Step the component models concurrently rather than one after another, so the coupled wall clock is not the sum of the components. Closes #1700.

## Content
Three configuration options, **all default `false`**, each building on the last.

**`step_concurrently`** advances the components as two groups that run at the
same time in separate threads: land then atmosphere in one task, sea ice then ocean in another. The groupings are forced, not chosen. Land and atmosphere are implicitly coupled. Sea ice and ocean turn out to be coupled too, less visibly: the sea ice model is
constructed holding `view`s into the ocean's surface velocity and salinity (`ocean_surface_velocities`, `ocean_surface_salinity`), so stepping them in parallel lets the ice read state the ocean is writing. Membership is decided by type, via `Interfacer.is_overlapped`, so a new ocean or ice model joins without bookkeeping.

**`overlap_slow_surfaces`** spans one ocean/sea ice step across several coupling steps. Because those components typically take a timestep several coupling steps long, `step_concurrently` alone can only hide one coupling step of atmosphere and land work behind them. This launches their step where its forcing is fully assembled and joins it several steps later. While it is in flight the coupler must not touch ice or ocean state, so `step_model_sims!`, `update_surface_fractions!`, `exchange!`, `update_model_sims!`, the flux accumulator push and `ocean_seaice_fluxes!` all stand down for those components. The atmosphere still needs a blended surface state every step, so
`combine_surfaces!` sums the fast and slow surfaces separately and reuses the slow sum for the window — land keeps contributing live values.

**`prime_slow_surfaces`** advances the slow group one step during initialization, so each overlapped step integrates the window about to happen rather than the one that just did and lands in time to be used on schedule.

## Testing
`test/overlap_schedule_tests.jl` is new and runs in CI: 28 tests, about two seconds, no GPU. It drives the real scheduling functions (not a copy of their ordering) with stub ocean and ice sims, covering `k` = 1, 2 and 5, an overlap-off control, both `Float64` and `ITime` time, the lag comparison across baseline/overlap/priming, and two restart scenarios including a mid-window checkpoint.

## Note for reviewers: what needs a judgement call rather than a code read
- **Priming moves a lag rather than removing it.** The atmosphere sees the ocean at the right time again, but the ocean integrates each window using the previous window's accumulated fluxes. The argument for accepting that is a timestep-ratio one: a component whose step spans several coupling steps already integrates under forcing held constant across a window, so shifting it by one window is the same order as an approximation already accepted. It weakens as `dt_ocean` approaches `dt_cpl`. The coupler team has discussed and agreed this is the better trade; recording it here for the record.
- **A startup transient.** During the first window a primed run shows the atmosphere an ocean slightly ahead of coupler time.
- **Diagnostic timestamps.** Under priming the slow components' own diagnostics are written on their own clocks and lead coupler time by up to one slow step. Left as-is deliberately — that is the honest label, and it is what those components' developers want to see. Coupler diagnostics stay on coupler time. Documented in `simcoordinator.md` and cross-referenced from `checkpointer.md`.
- I have **not** audited whether downstream post-processing assumes a shared time axis across components. Priming would break that assumption silently.

----


docs/src/simcoordinator.md
```diff
@@ -17,6 +17,54 @@ These functions orchestrate the interaction between component models (atmosphere
...
+By default `step!` advances the component models one after another. Two
+configuration options change that.
+
+`step_concurrently` splits them into two groups that advance at the same time:
+land then atmosphere in one task, sea ice then ocean in another. The groupings
+are not arbitrary. Land and atmosphere are implicitly coupled, so they have to
+run in sequence. Sea ice and ocean have to run in sequence for a different
+reason: the sea ice model is constructed holding views into the ocean's surface
+velocity and salinity fields, so stepping them in parallel would let the ice
+read state the ocean is writing. Group membership is decided by type, through
```
hgpeterson:
```suggestion
are not arbitrary. Land and atmosphere are implicitly coupled, and ocean and
sea ice must step together to properly pass the frazil heat flux. 
Group membership is decided by type, through
```
see #2041 and #2111

docs/src/simcoordinator.md
```diff
@@ -17,6 +17,54 @@ These functions orchestrate the interaction between component models (atmosphere
...
+
+The cost is that the atmosphere sees an ocean one slow step older than it
+otherwise would. `prime_slow_surfaces` removes that by advancing the slow group
+one step during initialization, so each overlapped step integrates the window
+that is about to happen rather than the one that just did and lands in time to
+be used on schedule. The lag does not vanish; it moves to the ocean's forcing,
+which is integrated from the previous window's accumulated fluxes. That is the
+side better able to absorb it, since a component whose timestep spans several
+coupling steps already integrates under forcing held constant across a whole
+window.
```
hgpeterson: An ASCII graphic here showing the coupling cadence would be helpful here.

src/FieldExchanger.jl
```diff
@@ -317,36 +409,269 @@ Iterates `step!` over all component model simulations saved in `cs.model_sims`.
...
+
+        # Group 2: sea ice and ocean. These must NOT be separate tasks. The sea
+        # ice model is built holding views into the ocean's live surface fields
+        # -- `ocean_surface_velocities` and `ocean_surface_salinity` return
+        # `view`s of `ocean.model.velocities.u/v` and `ocean.model.tracers.S`,
+        # which are captured in the ice model's `SemiImplicitStress` and
+        # `IceWaterThermalEquilibrium`. Stepping them in parallel lets the ice
+        # read ocean velocity and salinity while the ocean is writing them.
+        # Step the ice first, matching the ordering of the sequential branch
+        # below, so the ice sees the ocean state from before the ocean steps.
```
hgpeterson:
```suggestion
        # Group 2: sea ice and ocean.
```
Lots of long, repetitive comments and docstrings from claude in this PR... Can we try to cut some of them back?

src/FieldExchanger.jl
```diff
@@ -317,36 +409,269 @@ Iterates `step!` over all component model simulations saved in `cs.model_sims`.
...
+        ice_ocean_group = () -> step_slow_sims!(model_sims, t)
+
+        if skip_slow
+            # An asynchronous ice/ocean step is already in flight (or is being
+            # launched separately); advance only the fast group here.
+            land_atmos_group()
+            return nothing
+        end
+
+        if get(ENV, "COUPLER_SEQUENTIAL_GROUPS", "0") in ("1", "true", "TRUE", "yes")
```
hgpeterson: Why is this an environment variable?

src/FieldExchanger.jl
```diff
@@ -317,36 +409,269 @@ Iterates `step!` over all component model simulations saved in `cs.model_sims`.
...
+
+"""
+    step_slow_sims!(model_sims, t)
+
+Advance the overlapped group (sea ice, then ocean) to time `t`.
+
+This is the body handed to the asynchronous task when `overlap_slow_surfaces` is
+set. The ordering matches the group in `step_model_sims!`: the sea ice holds
+views into the ocean's surface velocity and salinity, so it must step before the
+ocean rather than beside it.
```
hgpeterson:
```suggestion
set. The ordering matches the group in `step_model_sims!`.
```

src/FieldExchanger.jl
```diff
@@ -317,36 +409,269 @@ Iterates `step!` over all component model simulations saved in `cs.model_sims`.
...
-                coupler_fields,
-                sim,
-                model_sims.atmos_sim,
-                thermo_params,
-            )
+        sim isa Interfacer.AbstractSeaIceSimulation && Interfacer.step!(sim, t)
     end
+    for sim in model_sims
+        sim isa Interfacer.AbstractOceanSimulation && Interfacer.step!(sim, t)
+    end
```
hgpeterson, line 502: @simone-silvestri will this mess up #2041/#2111 because if forces sea-ice then ocean?

src/FluxCalculator.jl
```diff
@@ -27,9 +27,24 @@ export turbulent_fluxes!,
     push_ready_accumulators!,
     reset!
 
-function turbulent_fluxes!(cs::Interfacer.CoupledSimulation)
+function turbulent_fluxes!(
+    cs::Interfacer.CoupledSimulation;
+    slow_frozen::Bool = false,
+    force_slow_push::Bool = false,
+)
     turbulent_fluxes!(cs.fields, cs.model_sims, cs.thermo_params, cs.flux_accumulators)
```
hgpeterson: Doesn't this also need to be guarded because `compute_surface_fluxes!` will read the live state of the slow models?

src/Interfacer.jl
```diff
@@ -215,6 +253,24 @@ abstract type AbstractSeaIceSimulation <: AbstractSurfaceSimulation end
...
+coupling timestep, and the sea ice holds views into the ocean's surface velocity
+and salinity, so the two must advance as a unit rather than in parallel.
+
+Membership is decided by type rather than by the key a model happens to be stored
+under in `model_sims`, so adding an ocean or sea ice model needs no bookkeeping
+and a mistyped key cannot silently exclude a model from the group.
+"""
+is_overlapped(::AbstractComponentSimulation) = false
+is_overlapped(::AbstractOceanSimulation) = true
+is_overlapped(::AbstractSeaIceSimulation) = true
```
hgpeterson: These will also apply for `SlabOceanSimulation` and `PrescribedSeaIceSimulation`, which I don't think we want.

src/FieldExchanger.jl
```diff
@@ -384,21 +695,42 @@ function combine_surfaces!(csf, sims, field_name_val::Val{field_name}) where {fi
...
             # Zero out the contribution from this surface if the area fraction is zero.
             # Note that multiplying by `area_fraction` is not sufficient in the case of NaNs
-            combined_field .+=
+            contribution =
                 area_fraction .* ifelse.(area_fraction .≈ 0, zero(FT), surface_field)
+            if !isnothing(slow_sum) && Interfacer.is_overlapped(sim)
+                slow_sum .+= contribution
+            else
+                combined_field .+= contribution
+            end
```
hgpeterson: This is allocating. Replace with something like:
```
target = (!isnothing(slow_sum) && Interfacer.is_overlapped(sim)) ? slow_sum : combined_field
target .+= area_fraction .* ifelse.(area_fraction .≈ 0, zero(FT), surface_field)
```

src/Checkpointer.jl
```diff
@@ -231,6 +231,13 @@ This is a callback function that checkpoints all simulations defined in the
 current coupled simulation.
 """
 function checkpoint_sims(cs::Interfacer.CoupledSimulation)
+    # Checkpointing reads every component's state, so an overlapped ice/ocean
+    # step must be joined first. No-op unless one is in flight.
+    inflight = cs.slow_task[]
+    if !isnothing(inflight)
+        wait(inflight.task)
+        cs.slow_task[] = nothing
+    end
```
hgpeterson: Use `FieldExchanger.wait_slow_sims!(cs)`?

PR #1725, opened 2026-02-11
add callback

kmdeck:
## Purpose 
When we added the pmodel version of ClimaLand to the coupled run, I forgot to add the model callbacks. Because of this, VCmax25 was always zero, and there was no photosynthesis (and then no ET because the stomata were not open).

This adds the callbacks.

## Content

----

### DIFF
--- a/ext/climaland/climaland_integrated.jl
+++ b/ext/climaland/climaland_integrated.jl
@@ -338,7 +338,7 @@ function ClimaLandSimulation(
     updatefunc = CL.make_update_drivers(CL.get_drivers(model))
     driver_cb = CL.DriverUpdateCallback(updatefunc, update_dt, tspan[1])
     dt_itime = dt isa ITime ? dt : ITime(Int(dt), Second(1), start_date)
-    t0_itime = tspan[1] isa ITime ? tspan[1] : ITime(Int(t0), Second(1), start_date)
+    t0_itime = tspan[1] isa ITime ? tspan[1] : ITime(Int(tspan[1]), Second(1), start_date)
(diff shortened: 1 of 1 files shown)


ext/climaland/climaland_integrated.jl
```diff
@@ -336,6 +336,10 @@ function ClimaLandSimulation(
     update_dt = dt isa ITime ? ITime(3600) : 3600
     updatefunc = CL.make_update_drivers(CL.get_drivers(model))
     driver_cb = CL.DriverUpdateCallback(updatefunc, update_dt, tspan[1])
+    dt_itime = dt isa ITime ? dt : ITime(Int(dt))
```
juliasloan25: maybe we need to pass `epoch = start_date` to the ITime constructors too?

juliasloan25: oh, it's because we call `date(t0)` but we never call `date(update_dt)`. So even though they're both ITimes it's okay for `update_dt` to not have an epoch.

From the tests stacktrace in [this build](https://buildkite.com/clima/climacoupler-ci/builds/7951#019c4eae-0fb3-4da3-a3fa-99bc37cfa1f2), the call to `date` is what fails:
```
  Got exception outside of a @test
  Time does not have epoch information
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] date(t::ClimaUtilities.TimeManager.ITime{Int64, Dates.Second, Nothing})
      @ ClimaUtilities.TimeManager /resnick/scratch/esm/slurm-buildkite/climacoupler-ci/depot/cpu/packages/ClimaUtilities/5ClV5/src/ITime.jl:129
```

Issue #1691, opened 2026-01-28
ClimaCalibrate Extension


nefrathenrici:
Following feedback from @juliasloan25 , after #1583 is merged we should add a ClimaCalibrate extension to standardize and test the `CalibrateConfig` API and related boilerplate.

PR #1727, opened 2026-02-12
Add CalibrateTools and refactor generating observations

ph-kev:
As the first step towards https://github.com/CliMA/ClimaCoupler.jl/issues/1691, this PR adds `CalibrateTools` as a module in ClimaCoupler and implements a `ERA5DataLoader` struct for loading preprocessed ERA5 data as `OutputVar`s. 

The module is not an extension, because the dependencies for this module are already strong dependencies for ClimaCoupler.

### DIFF
--- a/experiments/calibration/subseasonal/generate_observations.jl
+++ b/experiments/calibration/subseasonal/generate_observations.jl
@@ -2,92 +2,25 @@ import ClimaAnalysis
 import ClimaAnalysis: OutputVar
 import ClimaCalibrate
 import ClimaCoupler
-import Dates
+import ClimaCoupler: CalibrationTools
 import EnsembleKalmanProcesses as EKP
 import JLD2
 import ClimaDiagnostics
 import ClimaCore
-import ClimaUtilities.ClimaArtifacts.@clima_artifact
-import Pkg.Artifacts
 # Access CalibrateConfig
 include(joinpath(@__DIR__, "run_calibration.jl"))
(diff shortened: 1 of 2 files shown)


src/CalibrateTools.jl
```diff
@@ -0,0 +1,122 @@
...
+
+"""
+    ERA5DataLoader(; era5_to_clima_names = ERA5_TO_CLIMA_NAMES)
+
+Construct a data loader which you can load preprocessed ERA5 data in
+`OutputVar`, where
+- the short name, sign of the data, and units match CliMA conventions
+- the latitudes are from -180 to 180 degrees,
+- the times are at the start of the time period (e.g. the time average of
+  Janauary is on the first of Janauary instead of January 15th),
```
nefrathenrici:
```suggestion
  January is on the first of January instead of January 15th),
```

src/CalibrateTools.jl
```diff
@@ -0,0 +1,122 @@
...
+    )
+    return _preprocess(var, Val(Symbol(short_name)))
+end
+
+"""
+    _preprocess(var, ::Val{varname}) where {varname}
+
+Preprocess `var` with short name `varname`.
+"""
+function _preprocess(_, ::Val{varname}) where {varname}
```
nefrathenrici: Nitpick, but should this function just be called `preprocess` if it is intended to be extended? The error message below also refers to `preprocess` instead of `_preprocess`

PR #1729, opened 2026-02-13
start using LandSimulation constructor

juliasloan25:
## Purpose 
We can use ClimaLand's LandSimulation constructor to simplify the setup of our bucket and integrated land components. We hadn't used this previously because it initialized the cache in the constructor, but we need to initialize the land cache after a first exchange in coupled runs. ClimaLand has been changed so the cache isn't initialized in the coupled case.

This PR also uses functions from ClimaLand to simplify how we set ICs of the land models.

AMIP runs with the bucket look unchanged in buildkite. AMIP + integrated land has slight changes in momentum fluxes, and CMIP has slight changes in multiple fluxes

## Content
- [x] use LandSimulation constructor for bucket
- [x] use LandSimulation constructor for integrated land
- [x] use new IC function for bucket
- [x] use new IC functions x2 for integrated land

## Content that is less-related to the LandSimulation change
- [x] improve fluxes tests for bucket case
- [x] set everything needed to compute radiation (T, albedo, emissivity) in bucket/land model initialization. Now we should be able to remove the extra method of `set_surface_albedo!` for the coupled case in ClimaAtmos

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -109,7 +109,7 @@ steps:
           slurm_ntasks: 2
           slurm_mem: 32GB
 
-      - label: "GPU restarts"
+      - label: "GPU restarts (state and cache)"
         command: "julia -O0 --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/test/restart.jl"
         env:
           CLIMACOMMS_DEVICE: "CUDA"
@@ -118,7 +118,7 @@ steps:
           slurm_gres: "gpu:1"
           slurm_mem: 32GB
 
-      - label: "GPU restarts"
+      - label: "GPU restarts (state only)"
         command: "julia -O0 --color=yes --project=experiments/ClimaEarth/ experiments/ClimaEarth/test/restart_state_only.jl"
         env:
           CLIMACOMMS_DEVICE: "CUDA"
--- a/experiments/ClimaEarth/test/fluxes_test.jl
+++ b/experiments/ClimaEarth/test/fluxes_test.jl
@@ -27,11 +27,55 @@ include(joinpath("..", "code_loading.jl"))
 
     # Construct coupled simulation and run one coupling step
     cs = CoupledSimulation(config_dict)
-    step!(cs)
     boundary_space = Interfacer.boundary_space(cs)
 
     # Unpack component models
     (; atmos_sim, land_sim, ocean_sim, ice_sim) = cs.model_sims
+    land_fraction = Interfacer.get_field(land_sim, Val(:area_fraction))
+
+    # Check SWD, LWD, albedo, temp, emissivity between atmos and land before flux calculation
+    atmos_swd = Interfacer.get_field(boundary_space, atmos_sim, Val(:SW_d))
+    land_swd = Interfacer.remap(boundary_space, land_sim.integrator.p.drivers.SW_d)
+    err_swd = @. atmos_swd - land_swd
+    err_swd = @. ifelse(land_fraction ≈ 0, zero(err_swd), err_swd)
+    @test maximum(abs.(err_swd)) < 1e-10
+
+    atmos_lwd = Interfacer.get_field(boundary_space, atmos_sim, Val(:LW_d))
+    land_lwd = Interfacer.remap(boundary_space, land_sim.integrator.p.drivers.LW_d)
+    err_lwd = @. atmos_lwd - land_lwd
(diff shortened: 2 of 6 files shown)


ext/climaland/climaland_integrated.jl
```diff
@@ -117,10 +114,11 @@ function ClimaLandSimulation(
     # Interpolate atmosphere height field to surface space of land model,
     #  since that's where we compute fluxes for this land model
     atmos_h = Interfacer.remap(surface_space, atmos_h)
+    gustiness = FT(1)
```
kmdeck: We might pass this in via a keyword argument, because the atmosphere should be the one setting this parameter (right now it is 1 in any case)

ext/climaland/climaland_helpers.jl
```diff
@@ -98,3 +98,44 @@ function make_land_domain(
...
+    p.bucket.T_sfc .= CL.Domains.top_center_to_surface(Y.bucket.T)
+end
+function _coupler_set_ic!(Y, p, t, model::CL.LandModel, atmos_T, set_ic!)
+    p.drivers.T .= atmos_T
+    set_ic!(Y, p, t, model)
+
+    # Set albedo, T_sfc, and emissivity so that the atmosphere can compute radiation.
+    # Note that we normally use SWD, SWU, LWU to compute these, so we just set them to
+    # some reasonable values here. These will be overwritten during the first land step.
+    p.α_sfc .= 0.3
```
kmdeck: will this work if the FT of the sim is Float32?

ext/climaland/climaland_bucket.jl
```diff
@@ -374,8 +315,12 @@ function Interfacer.update_field!(
     Interfacer.remap!(sim.integrator.p.bucket.turbulent_fluxes.vapor_flux, field ./ ρ_liq) # TODO: account for sublimation
 end
 
-Interfacer.step!(sim::BucketSimulation, t) =
-    Interfacer.step!(sim.integrator, t - sim.integrator.t, true)
+function Interfacer.step!(sim::BucketSimulation, t)
```
kmdeck: is this mirroring what we do in integrated land?

ext/climaland/climaland_bucket.jl
```diff
@@ -156,127 +149,75 @@ function BucketSimulation(
...
-        saveat = saveat,
-        adaptive = false,
-        callback = SciMLBase.CallbackSet(diag_cb),
-    )
-
-    # TODO move these to ClimaLand.jl as part of CoupledRadiativeFluxes
-    radiative_fluxes =
-        (; SW_d = CC.Fields.zeros(surface_space), LW_d = CC.Fields.zeros(surface_space))
+    # Convert start_date and stop_date to ITime if using ITime
+    stop_date = start_date + Dates.Second(float(tspan[2]))
```
kmdeck: same comment as in integrated.jl for lines 200-203

ext/climaland/climaland_bucket.jl
```diff
@@ -156,127 +149,75 @@ function BucketSimulation(
...
 
-    return BucketSimulation(
-        model,
-        integrator,
-        area_fraction,
-        output_writer,
-        radiative_fluxes,
+    # Choose the timestepping algorithm
+    timestepper = CTS.ExplicitAlgorithm(CTS.RK4())
+    simulation = CL.Simulations.LandSimulation(
```
kmdeck: small future improvement for CL: we could have a default timestepper per model and so handle the choice of TS instead the simulations constructor for both Bucket + others

ext/climaland/climaland_bucket.jl
juliasloan25: Yeah exactly - what we had before didn't work with ITime so I updated it to do the same thing as integrated land

ext/climaland/climaland_helpers.jl
juliasloan25: It should work!
```
julia> field = ones(space)
Float32-valued Field:
  Float32[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]

julia> field .= 0.5
Float32-valued Field:
  Float32[0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5  …  0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5]
  ```

PR #1730, opened 2026-02-13
Iterative sea ice skin temperature

### LINKED ISSUE #1649: Use skin temperature for ClimaSeaIce flux calculation
Once we update to SurfaceFluxes v1 (in #1646), we'll be able to provide a function for the surface model's temperature which will get solved for iteratively with the fluxes. We should take advantage of this for the ClimaSeaIce model.

When we make this change, we should also remove the extra `RadiativeEmission` entry in the sea ice model's `top_heat_flux`.

akshaysridhar:
Solve flux balance at ice-surface to diagnose surface temperature. Passes an `update_T_sfc` callback function to the surface fluxes function (positional argument). 

Closes #1649

### DIFF
--- a/ext/ClimaCouplerCMIPExt/clima_seaice.jl
+++ b/ext/ClimaCouplerCMIPExt/clima_seaice.jl
@@ -1,8 +1,11 @@
 using ClimaSeaIce.SeaIceThermodynamics.HeatBoundaryConditions:
-    IceWaterThermalEquilibrium, MeltingConstrainedFluxBalance, get_tracer, RadiativeEmission
+    IceWaterThermalEquilibrium, PrescribedTemperature, get_tracer
 import ClimaComms
-import Thermodynamics as TD
 import ClimaOcean.EN4: download_dataset
+import SurfaceFluxes as SF
+import SurfaceFluxes.Parameters as SFP
+import Thermodynamics as TD
+using StaticArrays
 
 # Rename ECCO password env variable to match ClimaOcean.jl
 haskey(ENV, "ECCO_PASSWORD") && (ENV["ECCO_WEBDAV_PASSWORD"] = ENV["ECCO_PASSWORD"])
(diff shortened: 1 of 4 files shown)


ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -276,8 +279,12 @@ Interfacer.step!(sim::ClimaSeaIceSimulation, t) =
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:area_fraction}) = sim.area_fraction
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:ice_concentration}) =
     sim.ice.model.ice_concentration
+Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:ice_thickness}) =
+    Interfacer.remap(axes(sim.area_fraction), sim.ice.model.ice_thickness)
+Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:internal_temperature}) =
+    Interfacer.remap(axes(sim.area_fraction), sim.ocean_ice_interface.temperature)
```
juliasloan25: The convention we usually follow is to return the field on the component model's native grid from `get_field`. That way we can avoid allocating a new field if we instead call `remap!` wherever we access the field from. Or if we do need a new field we can call `Interfacer.get_field(boundary_space, sim, Val(:__))` to get the quantity and remap it

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -246,7 +250,6 @@ function sea_ice_simulation(
 
     bottom_heat_flux = OC.Field{OC.Center, OC.Center, Nothing}(grid)
     top_heat_flux = OC.Field{OC.Center, OC.Center, Nothing}(grid)
-    top_heat_flux = (top_heat_flux, ConcentrationMaskedRadiativeEmission(FT))
```
juliasloan25: we can delete this struct and the 2 functions defined for it in lines ~41-82 of this file

PR #1734, opened 2026-02-17
Move CalibrateConfig to CalibrationTools

ph-kev:
This PR moves `CalibrateConfig` to `CalibrationTools` and removes `api.jl` from the calibration directory.

This PR also add checks for whether the G ensemble matrix is completed and for sequential indices when filling out the G ensemble matrix. Without the first check, the G ensemble matrix could be filled with uninitialized values. Without the second one, incorrect data would be used to fill out the G ensemble matrix if the period of reductions are not the same (see `Checker.check` for `SequentialIndicesChecker` in [ClimaCalibrate docs](https://clima.github.io/ClimaCalibrate.jl/dev/api/#ClimaCalibrate.Checker.check)). 

This is another step towards https://github.com/CliMA/ClimaCoupler.jl/issues/1691.


experiments/calibration/subseasonal/observation_map.jl
```diff
@@ -58,7 +60,13 @@ function process_member_data!(g_ens_builder, diagnostics_folder_path, col_idx, i
         var = get_var(short_name, simdir)
         var = preprocess_var(var, sample_date_ranges)
 
-        EnsembleBuilder.fill_g_ens_col!(g_ens_builder, col_idx, var; verbose = true)
+        EnsembleBuilder.fill_g_ens_col!(
+            g_ens_builder,
+            col_idx,
+            var;
+            checkers = (EnsembleBuilder.SequentialIndicesChecker(),),
```
ph-kev: I didn't add `SignChecker` here, since it seems that some of the variables (e.g. hfss) doesn't have a consistent sign (e.g. all positive or all negative).

nefrathenrici (approved):
Thanks!

PR #1735, opened 2026-02-18
Remove sea ice warnings, add monthly calibration config, up deps 

costachris:
Improve logic for reading sea ice quantities from file (removes warnings introduced recently). Output only diagnostics needed for subseasonal calibration in model interface.

### DIFF
--- a/experiments/calibration/subseasonal/model_interface.jl
+++ b/experiments/calibration/subseasonal/model_interface.jl
@@ -26,11 +26,8 @@ function ClimaCalibrate.forward_model(iter, member)
     sim_length = Second(end_date - start_date)
 
     config_dict["start_date"] = start_date_str
-    config_dict["bucket_initial_condition"] = "/net/sampo/data1/wxquest_data/initial_conditions/era5_bucket_processed_$(start_date_str)_0000.nc"
     config_dict["t_end"] = "$(sim_length.value)secs"
     config_dict["checkpoint_dt"] = "900days"
-    config_dict["dt"] = "90secs"
-    config_dict["dt_cpl"] = "90secs"
 
     # Set member parameter file
     sampled_parameter_file = ClimaCalibrate.parameter_path(output_dir_root, iter, member)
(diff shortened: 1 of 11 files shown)


nefrathenrici (approved):
This looks good to me.

experiments/calibration/subseasonal/model_interface.jl
```diff
@@ -32,6 +32,15 @@ function ClimaCalibrate.forward_model(iter, member)
...
     config_dict["dt_cpl"] = "90secs"
 
+    # Output monthly averages to match the observation preprocessing
+    config_dict["extra_atmos_diagnostics"] = [
+        Dict(
+            "short_name" => CALIBRATE_CONFIG.short_names,
+            "period" => "1months",
+            "reduction_time" => "average",
+        ),
+    ]
```
ph-kev: This seems problematic if a value exists in `config_dict` for `extra_atmos_diagnostics`.

juliasloan25 (commented):
I agree with Kevin's comment about error handling but other than that this looks good to me!

Also, I don't think this fixes the IC error we got in the calibration longrun this week. Will that be done in a separate PR?

experiments/calibration/subseasonal/model_interface.jl
costachris: The point here was just to decouple the diagnostics in the top-level wxquest config (which the calibration is currently referencing) from diagnostics actually needed for calibration. The top-level .yml has diagnostics needed for the weekly submission, which are a different frequency (daily) and set of variables than what's used in the monthly calibrations. The idea was to write only the variables needed for the loss function, which is why it overwrites what's in the original yml. The drawback to just appending diagnostics is that the calibration would be outputting numerous diagnostics at daily frequencies that aren't used in the calibration, which might give an unnecessary I/O performance hit. 


Perhaps a better solution is just to have a separate config for calibration (vs the weekly operational runs) with diagnostics that match what's being calibration.

PR #1739, opened 2026-02-19
modify nightly and weekly runs

szy21:
## Purpose 

Updates some configs and parameters, mostly in the nightly and weekly runs. Some of them only makes sense when it is updated to the latest ClimaAtmos, and I commented them out for now.

Also modifies the nightly and weekly configs, as I am trying to reduce the number of total runs, so we can still have some gpus for testing during the weekend. I mostly only keep integrated land in the longrun. I kept one bucket run just to make sure we can still run it.

Closes #1741 #1740 

## Content

----


.buildkite/longruns/pipeline.yml
szy21: @nefrathenrici The climate sensitivity will be the difference in global mean `rlut - rsut` / 2 between the two simulations in this file: `amip_diagedmf_land_gpu` and `amip_diagedmf_land_p2k_gpu`. Is there a way to automate this?

juliasloan25 (approved):

PR #1747, opened 2026-02-20
WeatherQuest Calibration pipeline with weekly/monthly support + CERES, ERA5 Pressure Levels

costachris:
## Content

----

### DIFF
--- a/experiments/calibration/amip/generate_observations.jl
+++ b/experiments/calibration/amip/generate_observations.jl
@@ -137,8 +137,8 @@ if abspath(PROGRAM_FILE) == @__FILE__
     vars = select_pressure_levels.(vars, Ref(PRESSURE_LEVELS))
     lonlat_regridder = get_lonlat_regridder(config_file)
     vars = lonlat_regridder.(vars)
-    lat_left = -60
-    lat_right = 60
+    lat_left = -90
+    lat_right = 90
     vars = apply_lat_window.(vars, lat_left, lat_right)
 
     # Normalize data
--- a/experiments/calibration/amip/observation_map.jl
+++ b/experiments/calibration/amip/observation_map.jl
@@ -46,8 +46,8 @@ function preprocess_sim_vars(vars)
     vars = select_pressure_levels.(vars, Ref(PRESSURE_LEVELS))
     # We do not resample since the simulation variables are already on the
     # simulation grid
-    lat_left = -60
-    lat_right = 60
+    lat_left = -90
+    lat_right = 90
     vars = apply_lat_window.(vars, lat_left, lat_right)
 
     if isfile(NORMALIZATION_STATS_FP)
--- a/experiments/calibration/amip/preprocessing.jl
+++ b/experiments/calibration/amip/preprocessing.jl
@@ -30,8 +30,8 @@ Apply latitude window by constraining the longitudes to be in the range
 """
 function apply_lat_window(var, lat_left, lat_right)
     lats = ClimaAnalysis.latitudes(var)
-    first_lat_idx = findfirst(lon -> lon >= lat_left, lats)
-    last_lat_idx = findlast(lon -> lon <= lat_right, lats)
+    first_lat_idx = findfirst(lat -> lat >= lat_left, lats)
+    last_lat_idx = findlast(lat -> lat <= lat_right, lats)
     var = ClimaAnalysis.window(
         var,
         "latitude",


experiments/calibration/subseasonal_weekly/generate_observations.jl
```diff
@@ -0,0 +1,629 @@
...
+We pass start_date for both start and end to ObservationRecipe since the data
+represents the entire week as a single snapshot.
+"""
+function make_observation_vector(vars_by_date, sample_date_ranges, short_names)
+    include(joinpath(@__DIR__, "calibration_setup.jl"))
+
+    covar_estimator = ClimaCalibrate.ObservationRecipe.ScalarCovariance(;
+        scalar = Float64(CALIBRATION_NOISE_SCALAR),
+        use_latitude_weights = true,
+    )
```
ph-kev: I realized this function doesn't work too well since `sample_date_ranges` is overloaded for both calibration and what samples to choose for the observation. For example, the samples might be over 100 years, but you are only running a calibration for a couple of years.

experiments/calibration/subseasonal/observation_utils.jl
```diff
@@ -3,17 +3,33 @@ using Statistics
 import Dates
 using ClimaAnalysis
 
-include(joinpath(pkgdir(ClimaCoupler), "experiments/ClimaEarth/setup_run.jl"))
+# Shared variable units (used by both subseasonal and subseasonal_weekly)
 
 var_units = Dict(
+    # Surface temperature/pressure
     "pr" => "kg m^-2 s^-1",
```
ph-kev: This should be simplified once we put all of this in a data loader.

experiments/calibration/subseasonal/observation_utils.jl
```diff
@@ -26,3 +42,40 @@ function remove_global_mean(var)
...
+end
+
+"""
+    preprocess_var(var::ClimaAnalysis.OutputVar, sample_date_range)
+
+Preprocess simulation output `var` before flattening for the G ensemble matrix.
+
+Shifts the time axis by the largest period in the date range, sets units, and
+windows to the date range. Calls `largest_period` (defined per-pipeline) to
+determine the shift.
```
ph-kev: There shouldn't be a need for shifting the times.

experiments/calibration/subseasonal_weekly/generate_observations.jl
```diff
@@ -0,0 +1,629 @@
...
+    
+    var.attributes["short_name"] = short_name
+    var.attributes["start_date"] = start_date
+    
+    if !issorted(ClimaAnalysis.latitudes(var))
+        var = ClimaAnalysis.reverse_dim(var, ClimaAnalysis.latitude_name(var))
+    end
+
+    # Note: We do NOT add time dimension here - it's added in preprocess_vars after resampling
+    # This avoids issues with ClimaAnalysis.resampled_as which doesn't handle 3D data
```
ph-kev: `resampled_as` can handle any number of dimensions. It's just can't handle the time dimension too well since the distinction between dates and times is difficult to deal with. You can also call `resampled_as(var; lon = [0.0, 1.0], lat = [0.0, 1.0, 2.0])` if that is easier.

experiments/calibration/subseasonal_weekly/generate_observations.jl
```diff
@@ -0,0 +1,629 @@
...
+    if length(times) == 0
+        error("No CERES data found for $short_name in month of $start_date")
+    end
+    
+    @info "Loaded CERES $short_name for $(Dates.monthname(start_date)) $(Dates.year(start_date))"
+    
+    # Extract 2D data (average over time if multiple points, though should be 1)
+    if length(times) > 1
+        var = ClimaAnalysis.average_time(var)
+    end
```
ph-kev: I don't like this since it is introducing a fallback that shouldn't exist.

experiments/calibration/subseasonal_weekly/generate_observations.jl
```diff
@@ -0,0 +1,629 @@
...
+    
+    # Load the full time series
+    var = Base.get(loader, short_name)
+    
+    # Window to the month containing start_date
+    month_start = Dates.firstdayofmonth(start_date)
+    month_end = Dates.lastdayofmonth(start_date)
+    
+    # CERES dates are at start of month, so window to get the correct month
+    var = ClimaAnalysis.window(var, "time"; left = month_start, right = month_end)
```
ph-kev: Instead of this, this should use `select`with `MatchValue()`.

nefrathenrici:
For pre-rebase state with all 31 commits, see branch `ne/calibration_before_rebase`

PR #1753, opened 2026-02-23
Update configs and leaderboard

ph-kev:
This PR updates the leaderboard for diagnostics in pressure coordinates by throwing a warning if no diagnostics in pressure coordinates are found. Furthermore, this PR adds diagnostics in pressure coordinates for all configs in `amip_configs`, `longrun_configs`, and `nightly_configs`.

### DIFF
--- a/config/amip_configs/amip.yml
+++ b/config/amip_configs/amip.yml
@@ -7,6 +7,12 @@ coupler_toml: ["toml/amip_diagedmf.toml"]
 dt: "120secs"
 dt_cpl: "120secs"
 energy_check: false
+extra_atmos_diagnostics:
+  - short_name: [hur, hus, ta]
+    period: 1months
+    reduction_time: average
+    pressure_coordinates: true
+    compute_every: 6hours
 mode_name: "amip"
 netcdf_output_at_levels: true
 output_default_diagnostics: true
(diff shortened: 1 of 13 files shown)


ph-kev:
> > > Thanks! Would it be easier if we make them default atmos diagnostics?
> > 
> > 
> > I am not too sure if they should be part of the default since outputting diagnostics in pressure coordinates is computationally expensive.
> 
> Sounds good. How expensive is it?

Each accumulation / compute step of the diagnostics involves a kernel call to interpolation data from z coordinates to pressure coordinates. This isn't too expensive since `compute_every` is 6 hours, but it can be expensive if compute step is called more often. I haven't benchmark this, so it might be faster than what I think it is. It is probably okay to include them as part of the default atmos diagnostics.

szy21:
> This looks fine to me but I am a little worried about the diagnostics duplicated in so many files leading to problems later on e.g. if we don't update all of them. If we don't want to put them in the default atmosphere config, maybe we can add one like `clima_diagedmf.yml` but including the diagnostics - `clima_diagedmf_longrun.yml` or something like that

Oh, we can just add them in `config/atmos_configs/`, if it doesn't make ci slower? (not sure if `period: 1months` and `compute_every: 6hours` will cause a problem for short ci runs though)

PR #1755, opened 2026-02-24
Fix make_diagnostics_plot when there are pressure diagnostics

ph-kev:
This PR makes `make_diagnostics_plot`  compatible with pressure diagnostics. This is done by preferring diagnostics with z coordinates when making the diagnostics plots.

### DIFF
--- a/ext/ClimaCouplerMakieExt/diagnostics_plots.jl
+++ b/ext/ClimaCouplerMakieExt/diagnostics_plots.jl
@@ -27,7 +27,9 @@ short_name(var) = var.attributes["short_name"]
     )
 Create plots for diagnostics. The plots are saved to `plot_path`.
 This function will plot all variables that have been saved in `output_path`.
-The `reduction` keyword argument should be consistent with the reduction used to save the diagnostics.
+
+When plotting diagnostics, diagnostics that are daily averages in z coordinates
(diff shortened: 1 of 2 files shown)


ext/ClimaCouplerMakieExt/diagnostics_plots.jl
```diff
@@ -45,18 +45,20 @@ function Plotting.make_diagnostics_plots(
...
         # Use "average" if available, otherwise use the first reduction
         reductions = CAN.available_reductions(simdir; short_name)
-        "average" in reductions ? (reduction = "average") : (reduction = first(reductions))
+        reduction = "average" in reductions ? "average" : first(reductions)
         periods = CAN.available_periods(simdir; short_name, reduction)
-        "1d" in periods ? (period = "1d") : (period = first(periods))
-        vars[i] = get(simdir; short_name, reduction, period)
+        period = "1d" in periods ? "1d" : first(periods)
+        coord_types = CAN.available_coord_types(simdir; short_name, reduction, period)
+        coord_type = nothing in coord_types ? nothing : first(coord_types)
```
juliasloan25: Should the check instead be `isnothing(coord_types)`? Also, do we want to take a specific coordinate type rather than just the one that appears first?

PR #1763, opened 2026-02-26
Fix runaway eta (free sfc displacement) at high latitudes

akshaysridhar:
Current main branch simulations fail with runaway eta values at high latitudes - e.g. (Longrun; 16 horizontal elements in atmos cubed-sphere mesh) 

free_sfc_displacement_2657

This PR introduces an ocean flux high-lat mask which zeros all ocean surface flux updates where |lat| ≥ 78° (user defined polar mask latitude), so the ocean is not forced in that band. This fixes an issue with inconsistent flux balances in the overlap area and stabilizes simulations for up to 2 weeks. 

Screenshot 2026-02-26 at 11 45 01 AM

debug_coupler

### DIFF
--- a/ext/ClimaCouplerCMIPExt/clima_seaice.jl
+++ b/ext/ClimaCouplerCMIPExt/clima_seaice.jl
@@ -340,8 +340,8 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaSeaIceSimulation, fie
     OC.set!(sim.remapping.scratch_cc2, sim.remapping.scratch_arr2) # meridional momentum flux
 
     # Rename for clarity; these are now Center, Center Oceananigans fields
-    F_turb_ρτxz_cc = sim.remapping.scratch_cc1
-    F_turb_ρτyz_cc = sim.remapping.scratch_cc2
+    F_turb_ρτxz_cell = sim.remapping.scratch_cc1
+    F_turb_ρτyz_cell = sim.remapping.scratch_cc2
 
     # Set the momentum flux BCs at the correct locations using the remapped scratch fields
     # Note that this requires the sea ice model to always be run with dynamics turned on
@@ -350,8 +350,8 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaSeaIceSimulation, fie
     set_from_extrinsic_vector!(
         (; u = si_flux_u, v = si_flux_v),
         grid,
-        F_turb_ρτxz_cc,
-        F_turb_ρτyz_cc,
+        F_turb_ρτxz_cell,
+        F_turb_ρτyz_cell,
     )
 
     # Remap the latent and sensible heat fluxes using scratch arrays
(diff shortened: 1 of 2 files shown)


juliasloan25 (commented):
I had extended the land fraction down to 78 degrees to improve stability, but I wonder if it just delayed the bug that you're fixing here. Maybe we can keep

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -495,14 +499,21 @@ function FluxCalculator.ocean_seaice_fluxes!(
...
+    flux_v = OC.interior(oc_flux_v, :, :, 1)
+    flux_u .= ifelse.(polar_excl_u .≈ 0, zero(flux_u), flux_u)
+    flux_v .= ifelse.(polar_excl_v .≈ 0, zero(flux_v), flux_v)
 
     oc_flux_T = surface_flux(ocean_sim.ocean.model.tracers.T)
+    # polar-exclusion mask
     OC.interior(oc_flux_T, :, :, 1) .+=
-        OC.interior(ice_concentration, :, :, 1) .* OC.interior(Qi, :, :, 1) .* ρₒ⁻¹ ./ cₒ
-
+        polar_excl_centers .* OC.interior(ice_concentration, :, :, 1) .*
```
juliasloan25: I wonder if we might run into NaNs for the interface fluxes (Qi and salt) here too without the `ifelse`. But we can try leaving it like this and see how it goes

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -493,12 +543,12 @@ function FieldExchanger.update_sim!(sim::OceananigansSimulation, csf)
 
     # Update only the part due to radiative fluxes. For the full update, the component due
     # to latent and sensible heat is missing and will be updated in update_turbulent_fluxes.
-    oc_flux_T = surface_flux(sim.ocean.model.tracers.T)
     (; σ, C_to_K) = sim.ocean_properties
     α = Interfacer.get_field(sim, Val(:surface_direct_albedo)) # scalar
     ϵ = Interfacer.get_field(sim, Val(:emissivity)) # scalar
+    # polar-exclusion mask
```
juliasloan25: we could use ifelse here and for the salt flux below

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -348,6 +360,26 @@ Interfacer.get_field(sim::OceananigansSimulation, ::Val{:surface_diffuse_albedo}
 Interfacer.get_field(sim::OceananigansSimulation, ::Val{:surface_temperature}) =
     sim.ocean.model.tracers.T + sim.ocean_properties.C_to_K # convert from Celsius to Kelvin
 
+"""
+    ocean_flux_highlat_mask(grid; location)
+
+Build the ocean flux high latitude mask once at setup.
+Currently we define ocean between 80degS to 80degN with 2 degree overlap in the coupler mask.
```
juliasloan25:
```suggestion
Currently we define ocean between 80degS to 80degN.
```

PR #1768, opened 2026-02-27
store fields as NamedTuple of Fields

### LINKED ISSUE #1767: restructure exchange fields as a NamedTuple of Fields
Right now, the coupler exchange fields `cs.fields` stored in the CoupledSimulation object are stored as a Field of NamedTuples. Since we typically access one field at a time, this is likely worse for performance than storing a NamedTuple of Fields due to the underlying memory structure.

juliasloan25:
## Purpose 
Let's see if this is faster at all

Closes #1767

## Content

----


juliasloan25:
This didn't result in a meaningful speedup, so I'll close it for now. We can revisit in the future if desired

PR #1778, opened 2026-03-05
Added and enabled GW flags in AMIP run files

ray-chew:
CI of this PR depends on the following PRs to pass:
https://github.com/CliMA/ClimaArtifacts/pull/150
https://github.com/CliMA/ClimaAtmos.jl/pull/4346

## Purpose and Content

As the gravity wave schemes have been added to ClimaAtmos, we now enable gravity wave parameterization in all AMIP runs with non-trivial orographic.

Note that `nogw_dc` was increased from the default 0.8 to reduce the number of spectral components from 250 to 50 for GPU compatibility. This is necessary to avoid memory issues on the P100s.

----

### DIFF
--- a/config/atmos_configs/climaatmos_diagedmf.yml
+++ b/config/atmos_configs/climaatmos_diagedmf.yml
@@ -30,3 +30,5 @@ viscous_sponge: true
 z_elem: 63
 z_max: 60000.0
 microphysics_model: "0M"
+non_orographic_gravity_wave: true
+orographic_gravity_wave: "raw_topo"
--- a/config/atmos_configs/climaatmos_edonly.yml
+++ b/config/atmos_configs/climaatmos_edonly.yml
@@ -23,3 +23,5 @@ viscous_sponge: true
 z_elem: 63
 z_max: 60000.0
 microphysics_model: "0M"
+non_orographic_gravity_wave: true
+orographic_gravity_wave: "raw_topo"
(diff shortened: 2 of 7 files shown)


szy21:
Do you know if there is a significant difference between dc = 0.8 and dc = 4? (I ran it with the default dc and had no issue on A100 before - do you mean memory issues on P100, rather than H100?)

akshaysridhar (approved):

szy21:
> > > > Do you know if there is a significant difference between dc = 0.8 and dc = 4?
> > > 
> > > 
> > > It reduces the number of spectral components used in the non-orographic GW scheme from 250 to 50. But idk what the physics looks like. If you like, I can do a sensitivity study.
> > 
> > 
> > That's ok, but I think it would be good to only change the regular ci job to use dc = 4, so longrun and amip can still run with the default dc = 0.8 (they run on A100 so there will be no issue). And we can test the sensitivity when there is more time.
> 
> Ok, I am asking around how best to do that, since AMIP / longrun uses the same toml files as regular CI jobs.

You can add separate tomls for regular ci jobs. (Or, I'm also fine with not adding GWs to regular ci - we have the same memory issue for 1M, so we only add it to longrun, we just need to keep an eye on it every week.) The reason I want to change this is that I don't want to use a not-so-good parameter just because of P100, as we don't use P100 for our production run anyway.

szy21:
What's the status of this? @akshaysridhar @ray-chew Could you get this in?

akshaysridhar:
> What's the status of this? @akshaysridhar @ray-chew Could you get this in?

It was stuck in a CI library / CUDA failure cycle; cc @ray-chew please rebase this PR; we can try again today.

PR #1793, opened 2026-03-10
Add infrastructure for testing calibration

### LINKED ISSUE #1750: Add a test calibration pipeline
For this test calibration pipeline, it should exercise all the functionality of a full calibration pipeline except actually running the simulation. It should be sufficient to just generate dummy data to use instead. One way to do this is to call only the diagnostic callback. Furthermore, to make this faster, it should be possible to increase the time step as long as it is not longer than the period of outputting the diagnostics.

One problem with generating dummy data is that some checks might fail such as the sign check if it is enabled. Furthermore, there might be NaNs in the data which will cause issues as well. If this happen, `N` real steps can be taken to populate the ClimaCore fields.

2/24/26 - I remembered that `GEnsembleBuilder` can handle `NaN`s in the data so I don't think it will be an issue for that part. The iteration can still fail because of the number of `NaN`s in the G ensemble matrix.

ph-kev:
closes [#1570](https://github.com/CliMA/ClimaCoupler.jl/issues/1750)

This PR adds infrastructure for testing whether a calibration can run or not. A calibration involves

1. Generating observations
2. Running an ensemble of simulations to produce diagnostics
3. Computing the observation map (data from diagnostics to G ensemble matrix)
4. Run post iteration analysis (plotting of parameters, bias, and RMSE)

The main computational cost is running the ensemble of simulations and the amount of mistakes/typos that can happen in this step is minimal compared to the other steps. This PR adds the function `setup_and_emulate_diagnostics` which repeatedly call the diagnostics handler of the atmos simulation with a default time step of one day. This forces all the diagnostics of the atmos simulation to be produced with data that is probably not correct. Beside that, the diagnostics are exactly what would be produced from the atmos simulation.

When running a test calibration on `clima` with precompilation finished with a simulation lasting for a month, the first iteration takes ~20 minutes (most of the time is spent creating the simulation).

### DIFF
--- a/experiments/calibration/amip/model_interface.jl
+++ b/experiments/calibration/amip/model_interface.jl
@@ -1,5 +1,3 @@
-ENV["CLIMACOMMS_DEVICE"] = "CUDA"
-ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
 import ClimaCoupler
 import ClimaCalibrate
 using Pkg
@@ -8,6 +6,11 @@ using Pkg
 # This allows other pipelines to include their own run_calibration.jl first
 if !@isdefined(CALIBRATE_CONFIG)
     include(joinpath(@__DIR__, "run_calibration.jl"))
+end
+
+if !TEST_CALIBRATION
+    ENV["CLIMACOMMS_DEVICE"] = "CUDA"
+    ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
 end
 
 function ClimaCalibrate.forward_model(iter, member)
@@ -31,7 +34,12 @@ function ClimaCalibrate.forward_model(iter, member)
 
     @info "Simulation dates" start_date end_date
 
-    ClimaCoupler.SimCoordinator.setup_and_run(config_dict)
+    if !TEST_CALIBRATION
+        ClimaCoupler.SimCoordinator.setup_and_run(config_dict)
+    else
+        @info "Emulating diagnostics for test calibration"
+        CalibrationTools.setup_and_emulate_diagnostics(config_dict)
+    end
     @info "Completed member $member"
     return nothing
 end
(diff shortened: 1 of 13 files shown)


experiments/calibration/amip/model_interface.jl
```diff
@@ -1,5 +1,3 @@
-ENV["CLIMACOMMS_DEVICE"] = "CUDA"
-ENV["CLIMACOMMS_CONTEXT"] = "SINGLETON"
```
ph-kev: I am not sure if it is okay to remove these lines.

src/CalibrationTools.jl
```diff
@@ -627,4 +630,59 @@ function add_parameter_filepath!(config_dict, parameter_filepath)
...
+
+Set up and emulate producing the diagnostics of the coupled model simulation
+specified by the `config_dict`. Returns the `CoupledSimulation` after simulating
+the diagnostics.
+
+This is analogous to [`SimCoordinator.setup_and_run`](@ref) and is meant a
+drop-in replacement for starting a simulation that will produce only
+diagnostics.
+
+By default, the size of the time step will be set to one day.
```
juliasloan25: why does it matter what the timestep is if we don't step?

juliasloan25 (commented):
If this doesn't take too long to run, could we add a calibration test to the regular CI? I think that would help catch typos etc

src/CalibrationTools.jl
```diff
@@ -627,4 +630,50 @@ function add_parameter_filepath!(config_dict, parameter_filepath)
     return nothing
 end
 
+"""
+    setup_and_simulate_diagnostics(config_dict::AbstractDict; dt = "86400secs")
```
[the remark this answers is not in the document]
juliasloan25: I meant the `TEST_DIAGNOSTICS` env var that's used to trigger this new function. Is this something we want to run regularly in CI or is it expected to be run manually by people who are running calibrations? I think we'll be able to catch typos if we run this in CI

nefrathenrici (approved):
Thanks, this will be very useful for quickly testing modifications to the pipeline.

nefrathenrici (approved):

juliasloan25 (approved):
This looks good to merge. The IC artifact should be available now right? Is there anything else blocking this PR?

PR #1794, opened 2026-03-10
Update coupler stepping logic so that fluxes are exchanged as often as possible

### LINKED ISSUE #1756: Step component models such that fluxes are exchanged as often as possible
Currently, the coupler steps component models using the step size of the largest model, e.g. ocean, only exchanging fluxes at that quantum.  Instead, the coupler should step each model using the shortest surface model step size, so that it and atmos exchange fluxes as often as possible.

This requires that each component model correctly handles being asked to step to a time that requires no work, because it has not yet reached that model's step increment.  There are also assertions that need to be updated.

daverumph:
## Purpose 
Update coupler stepping logic so that fluxes are exchanged as often as possible
Closes #1756

## Content
Before this PR, the coupler time step (dt) was constrained to be the same as the dt of
the largest component model. Consequently, fluxes were exchanged less often than
optimal when one model (e.g. ocean) stepped much more slowly than others.  With
this change, the coupler can, and should, step at the rate of the fastest surface model.


ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -335,9 +350,28 @@ end
...
+    if isapprox(Δt, sim.model_Δt) || Δt > sim.model_Δt
+        @info "Ocean: stepping"
+        OC.time_step!(sim.ocean, Δt)
+        @info "Ocean: reached time t = $(sim.ocean.model.clock.time)"
+    end
+end
+
+function Interfacer.step!(sim::OceananigansSimulation, t::ITime)
+    Δt_msec = Dates.DateTime(t) - sim.ocean.model.clock.time
+    model_Δt_msec = Dates.DateTime(sim.model_Δt) - sim.model_Δt.epoch
```
juliasloan25: does this `model_\Deltat` have an epoch?

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -270,8 +276,12 @@ end
 ###############################################################################
 
 # Timestep the simulation forward to time `t`
-Interfacer.step!(sim::ClimaSeaIceSimulation, t) =
-    OC.time_step!(sim.ice, float(t) - sim.ice.model.clock.time)
+function Interfacer.step!(sim::ClimaSeaIceSimulation, t)
+    Δt = float(t) - sim.ice.model.clock.time
```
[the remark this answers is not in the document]
daverumph: Like our locally built models, sim.ice.model.clock.time can be a float, or it can be a Dates.DateTime.

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -335,9 +337,13 @@ end
...
-# Timestep the simulation forward to time `t`
-Interfacer.step!(sim::OceananigansSimulation, t) =
-    OC.time_step!(sim.ocean, float(t) - sim.ocean.model.clock.time)
+# Timestep the simulation forward to time `t`. This may not actually do anything.
+function Interfacer.step!(sim::OceananigansSimulation, t)
+    Δt = float(t) - sim.ocean.model.clock.time
+    if isapprox(Δt, sim.model_Δt) || Δt > sim.model_Δt
+        OC.time_step!(sim.ocean, Δt)
+    end
+end
```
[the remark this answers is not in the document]
daverumph: I (finally) addressed these requests.

src/Input.jl
```diff
@@ -678,7 +678,23 @@ function parse_component_dts!(config_dict)
...
+                @assert assertion "Coupler time step and atmosphere time step must be integer multiples of each other\n dt_cpl = $Δt_cpl\n $key = $component_dt"
+            else
+                # all other (surface) model dts must be divisible by the coupler dt,
+                # or the coupler time step must be divisible by the surface model dt,
+                # to ensure consistent coupling time steps and compability with legacy configs
+                assertion =
+                    isapprox(Δt_cpl % component_dt, 0.0) ||
+                    isapprox(component_dt % Δt_cpl, 0.0)
+                @assert assertion "All surface component dts and coupler dtmust be integer multiples of each other\n $key = $component_dt\n dt_cpl = $Δt_cpl"
+            end
```
juliasloan25: since we do the same check for all components we can drop the atmos check and the if/else here

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -513,13 +513,26 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
...
-        Interfacer.step!(sim.integrator)
+    # Don't step until we've reached a step boundary
+    # (This can happen if the coupler dt is less than this model's)
+    Δt = t - sim.integrator.t
+    if Δt >= sim.integrator.dt
+        while sim.integrator.t < t
+            Interfacer.step!(sim.integrator)
+        end
     end
-    return nothing
```
juliasloan25, line 522: Kevin's comment still stands - could you return `nothing` from both of the `step!` methods here for consistency?

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -173,8 +186,22 @@ end
...
+    end
+    return nothing
+end
+
+function Interfacer.step!(sim::ClimaSeaIceSimulation, t::ITime)
+    Δt_msec = Dates.DateTime(t) - sim.ice.model.clock.time
+    model_Δt_msec = Dates.DateTime(sim.model_Δt) - sim.model_Δt.epoch
+    if Δt_msec >= model_Δt_msec
+        OC.time_step!(sim.ice, float(sim.model_Δt))
+    end
```
juliasloan25: @ph-kev do you know if there's a more straightforward way to do this with ITimes?

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -311,12 +311,30 @@ function Interfacer.update_field!(
...
 end
 
-function Interfacer.step!(sim::BucketSimulation, t)
-    while float(sim.integrator.t) < float(t)
-        Interfacer.step!(sim.integrator)
+# Don't step if we haven't reached a step boundary
+# (This can happen if the coupler dt is less than this model's)
+function Interfacer.step!(sim::BucketSimulation, t::Float64)
+    model_t = Float64(sim.integrator.t)
+    Δt = t - model_t
```
juliasloan25, line 318: I'm a little concerned about the Float subtraction here. Instead of computing `Δt` as a float, could we instead compute the number of steps as an integer and use that to take the correct number of steps? And same for the other methods too

```
function Interfacer.step!(sim::ClimaLandSimulation, t::Float64)
    model_t = Float64(sim.integrator.t)
    model_dt = Float64(sim.integrator.dt)
    n_steps = round(Int, (t - model_t) / model_dt)
    for _ in 1:n_steps
        Interfacer.step!(sim.integrator)
    end
end
```
@ph-kev do you agree this should avoid the catastrophic cancellation?

ext/ClimaCouplerCMIPExt/clima_seaice.jl
ph-kev: I think a more idiomatic way of doing this is

```suggestion
    Δt_msec = date(t) - sim.ice.model.clock.time
    model_Δt_msec = counter(sim.model_Δt) * period(sim.model_Δt)
    if Δt_msec >= model_Δt_msec
        OC.time_step!(sim.ice, float(sim.model_Δt))
    end
```

I am not sure if this is more straightforward though.

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -156,10 +158,22 @@ function OceananigansSimulation(
         closure = (horizontal_viscosity, vertical_mixing)
     end
 
-    Δt = isnothing(dt) ? CO.OceanSimulations.estimate_maximum_Δt(grid) : float(dt)
+    if tspan[1] isa ITime
+        # create a model clock that uses DateTime, for compatibility with ITime.
+        model_clock = OC.TimeSteppers.Clock(time = start_date)
+        stop_time = Dates.DateTime(tspan[2])
```
ph-kev, line 164: Using `date` is more generic than `Dates.DateTime` if we ever want to support other calendar systems.

```suggestion
        stop_time = date(tspan[2])
```

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
ph-kev: I could be missing something, but aren't `t` and `model_t` relatively the same scale? I still think you would get catastrophic cancellation in this case. I haven't looked too deeply at what this function is doing, but `Δt` is only used for comparing against other values. For example, `isapprox(Δt, model_dt)` is the same as `isapprox(t - model_t, model_dt)` which is the same as `isapprox(t, 2 * model_dt)` which avoids the subtraction.

ext/ClimaCouplerCMIPExt/oceananigans.jl
daverumph: Times in ITime all have some epoch, which may be defaulted.  When expressing a time difference, as opposed to a point in time, it seems correct to use the difference between the current time and the epoch.  When you subtract one Dates.DateTime from another, you get either a Dates.Millisecond or a Dates.Days, depending on the values.

This code is admittedly kludgy, but I think is what is required to use ITime with DateTime.

ext/ClimaCouplerClimaAtmosExt.jl
daverumph: I had already made this change, probably on the next commit.

ext/ClimaCouplerCMIPExt/oceananigans.jl
ph-kev: It is not true that all `ITime`s have an epoch associated with them. 

```julia
julia> ClimaUtilities.TimeManager.ITime(1)
1.0 second [counter = 1, period = 1 second]
```

ext/ClimaCouplerCMIPExt/clima_seaice.jl
daverumph: I think new line 199 produces seconds, not milliseconds, right?  It's annoying that DateTime subtraction produces Millisecond results, and nothing else uses that.

ext/ClimaCouplerCMIPExt/oceananigans.jl
juliasloan25: It looks like in this case it will have epoch because it's constructed from the model dt, which should get the epoch when we apply promote [here](https://github.com/CliMA/ClimaCoupler.jl/blob/dac5ca3dba7ce632b2890cfbd3ecf1133f43b0e3/src/Input.jl#L441-L452). So this code is probably okay

ext/ClimaCouplerClimaAtmosExt.jl
juliasloan25: I don't see it in the diff on the PR, which includes changes from all commits

ext/ClimaCouplerCMIPExt/clima_seaice.jl
juliasloan25: Could we do something similar here to the proposed change for Float time if we use `counter` to tell how many steps we need to take, rather than working with dts?

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
juliasloan25: The rounding should counteract the catastrophic cancellation - e.g. if `(t - model_d) / model_dt` gives 0.999997 instead of 1.0 (or even 0.9), it will get rounded to 1. 

I think it's better to use concrete number of steps here rather than any `isapprox` (though I agree what you're suggesting is better), so that we reduce the uncertainty in how many steps will be taken.

ext/ClimaCouplerCMIPExt/oceananigans.jl
daverumph: I didn't change it because in my repl date() was undefined.  Where does it come from?

ph-kev: See https://clima.github.io/ClimaUtilities.jl/dev/timemanager/#ClimaUtilities.TimeManager.date

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: To avoid the catastrophic situation, use ITime/DateTime.  That's why we're putting in support for it.
Is there a consensus on what we should do here?

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
```diff
@@ -378,12 +378,29 @@ function Interfacer.update_field!(
...
+function Interfacer.step!(sim::ClimaLandSimulation, t::ITime)
+    Δt = t - sim.integrator.t
+    if Δt >= sim.integrator.dt
+        while sim.integrator.t < t
+            Interfacer.step!(sim.integrator)
+        end
     end
     return nothing
 end
+function Interfacer.step!(sim::ClimaLandSimulation, t::Float64)
```
ph-kev, line 392: Was there a reason why we can't do this?

```suggestion
function Interfacer.step!(sim::ClimaLandSimulation, t::AbstractFloat)
```

ext/ClimaCouplerClimaAtmosExt.jl
daverumph: I made the change everywhere else, but miss this one.  It's in the next commit.

ext/ClimaCouplerCMIPExt/oceananigans.jl
daverumph: I think it's best to leave it as a DateTime because that's what the Oceananigan clock is, if it's not a Float64.  Using ITime.date() didn't work for me.

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
daverumph: We could, which would be more general and support Float32, but that's much worse at representing times than Float64, and I thought it would be better to not support it rather than produce wrong results or crash in a subtle way later.  But maybe a check for t being a Float32 should be done explicitly elsewhere.

ext/ClimaCouplerCMIPExt/oceananigans.jl
juliasloan25: I do think it's better to use `date` for the reasons Kevin mentioned. You can see an example of it called in ClimaCoupler [here](https://github.com/CliMA/ClimaCoupler.jl/blob/dac5ca3dba7ce632b2890cfbd3ecf1133f43b0e3/src/Interfacer.jl#L107), and it's imported at the top of that file [here](https://github.com/CliMA/ClimaCoupler.jl/blob/dac5ca3dba7ce632b2890cfbd3ecf1133f43b0e3/src/Interfacer.jl#L14).

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
juliasloan25: I think it's ok to not support Float32 time type

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: I don't see a viable way around this without changing the call signature.  The problem is we're handed t, which is exactly what is getting larger and larger, and needs to be compared to the model_t to determine if it's time to step the model.  Alternatively, we could have the coupler figure out when to step each model strictly on the basis of the ratio of dt's, but that would require an interface change I think.  I suppose step! could ignore the t argument and just keep internal to the model and use something like model_dt_since_last_step, which would keep resetting when the model actually stepped.  Each model would need to know dt_cpl though to know how much to advance model_dt_since_last_step.

I continue to think that using Float64 clocks is inherently flawed for long runs and we already have the solution of using ITime/DateTime for that case, so we shouldn't spend more time trying to make the Float64 case better.  We should just document that it should not be used for long simulations.

ph-kev: I thought Julia's approach was reasonable by computing the number of steps to take to help with catastrophic cancellation. Is there any blockers with that approach? I agreed with for long runs, you would want to use `ITime`, but I don't think the `Float64` case be neglected.

daverumph: I don't think Julia's method works correctly.  The land and coupler dts are expected to be the same, so the problem doesn't surface, but I'm assuming we want to make the same change for other component models, where t increments are smaller than the model_dt.  Take the ocean model as an example, with model_dt == 1800.0, and dt_cpl == 360.0.
With model_t == 0.0, i.e. just getting started, step! will be called sequentially with t == 360.0, 720.0, 1080.0, 1440.0, and finally 1800.0 as the coupler runs the faster-stepping land and atmos models.  But the round() will cause the ocean model to step at t == 1080.0, because it starts rounding up at the halfway point, 900.0, not at t == 1800.0 as we want.

config/ci_configs/cmip_oceananigans_climaseaice.yml
```diff
@@ -3,7 +3,7 @@ albedo_model: "CouplerAlbedo"
 atmos_config_file: "config/atmos_configs/climaatmos_edonly.yml"
 coupler_toml: ["toml/amip_edonly.toml"]
 dt_atmos: "120secs" # 2 minutes
-dt_cpl: "1800secs" # 30 minutes
+dt_cpl: "360secs" # 6 minutes
```
ph-kev, line 6: Could you revert this to keep the previous behavior?

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
```diff
@@ -378,12 +378,29 @@ function Interfacer.update_field!(
...
+    model_t = Float64(sim.integrator.t)
+    Δt = t - model_t
+    model_dt = Float64(sim.integrator.dt)
+    if isapprox(Δt, model_dt, atol = 0.125) || Δt > model_dt
```
ph-kev, line 396: I don't think you should be using absolute tolerance here. I am also confused on where this number comes from.

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: Based on some testing, Float64 time does not have a risk of catastrophic cancellation anywhere near the time ranges we're interested in.  The concern is very real for Float32 time, however.

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -311,12 +311,30 @@ function Interfacer.update_field!(
...
+    model_dt = Float64(sim.integrator.dt)
+    if isapprox(Δt, model_dt, atol = 0.125) || Δt > model_dt
```
imreddyTeja, line 320: What is the reasoning behind an `atol` value of 0.125?

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -311,12 +311,30 @@ function Interfacer.update_field!(
...
+        while Float64(sim.integrator.t) < t
```
imreddyTeja, line 321: Would it be better to use a `for` loop instead of a `while` loop here?
something like:
```
if isapprox(Δt, model_dt, atol = 0.125) || Δt > model_dt
    n_steps = round(Int, Δt  / model_dt)
    for _ in 1:n_steps
          Interfacer.step!(sim.integrator)
```
I think that would avoid an infinite loop in case the condition where `sim.integrator.t  + model_dt == sim.integrator.t ` because `sim.integrator.t ` is very large

imreddyTeja (approved):
I agree that it would be good to revert the `dt_cpl` changes in the config ymls before merging this.

config/ci_configs/cmip_oceananigans_climaseaice.yml
daverumph: The point of this PR is to improve the behavior, not keep it.  The previous coupler time step is less accurate because land/atmos flux exchange only occurs every 30 minutes.

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: I want something that's bigger than the default, but not close to actual step sizes.  I arbitrarily picked 1/8 second, but it's a power of 2 so that it's precisely represented as a Float.

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: Rounding won't work correctly if Δt is not an even multiple of model_dt, but the coupler does enforce that condition.  I'll make the change.

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
daverumph: As I replied to Teja, it's intended to be larger than the default, which could run into false negatives due to subtracting large floating point numbers, but much smaller than our actual step sizes.  The compiler has no way top know what our actual expected float ranges are, so I'm telling it explicitly.  0.125 represents 1/8 second, where our step sizes are typically minutes, and it's a power of 2 (2^-3), so its floating point representation is exact.

PR #1796, opened 2026-03-11
enable checkpointing for CMIP

### LINKED ISSUE #1383: add support for checkpointing OceananigansSimulation
Right now we can't save and restart simulations when using `OceananigansSimulation`. We need to extend our checkpointing infrastructure for this simulation type so that we can restart. This will involve extending functions from the `Checkpointer` module, including `get_model_prog_state`, `restart_model_state!`, and probably more. We may also need to revisit our checkpointing structure to cleanly accommodate non-ClimaCore objects.

juliasloan25:
## Purpose 
Now that CMIP simulations are stable for longer (!) it will be useful to have checkpointing + restarts for this setup.

Closes #1383 

## Content
- [x] extend ClimaCoupler's checkpointing/restarting interface for Oceananigans-based simulations (OceananigansSimulation and ClimaSeaIce)
- [x] add a test that runs a CMIP sim for 2 steps, saves, restarts, and runs for 2 more steps - then compare this to a sim that runs for 4 steps straight through (as we do in the existing `restart_state_only.jl` test)
- [x] generalize the logic in src/Checkpointer.jl to handle both the ClimaCore and Oceananigans cases cleanly
- [x] update docs


juliasloan25:
replaced by #1909

PR #1803, opened 2026-03-12
Support coupled single-column runs

juliasloan25:
## Purpose 
We want ClimaCoupler to be a general coupling infrastructure. Right now it has the limitation of only supporting global runs. Since each component model can run in single column mode, there isn't really anything holding us back from running coupled SCM runs.

This PR adds a couple of hacky things to get SCMs working:
- `_column_boundary_space` - needed because we can't easily construct a PointSpace that includes lat/lon using ClimaCore functions. Land model needs lat/lon for spatial inputs, and we pass the lat/lon to the atmosphere so that's in sync too
  - alternative solution: construct spaces without lat/lon, and for land use land domain functions to add in lat/lon
- dss functions (`Utilities.init_dss_buffer`, `Utilities.apply_dss!`) - dss doesn't make sense for a space without horizontal direction, so it should do nothing. However ClimaCore doesn't handle this case - instead of doing nothing, calling dss on a PointSpace or Column errors.
  - I'll fix this in ClimaCore and update ClimaCoupler after

## Content
- [x] add new input options specific to the SCM mode `domain_type`, `column_latlon`, `scm_surface_type`
- [x] add a function `create_boundary_space` which constructs the correct space based on `domain_type`
- [x] extend `remap!` for PointSpaces (no regridding)
- [x] write wrappers for dss since dss doesn't handle the PointSpace case (this should be changed in ClimaCore - we just want to do nothing in this case)
- [x] update docs (new options, example for running SCM)
- [x] add SCM tests to CI
- [x] add plots to SCM cases

## Plotting
The SCM case requires new plotting infrastructure. For depth-resolved (column) variables, line plots of the value vs depth are made (for whatever time the variable was saved at/reduced over. For 1D (scalar) variables, a plot axis is made but is left empty. The extrema are put in the plot title, which for the scalar case displays the variable's value.

For example, atmosphere debug plots:
Screenshot 2026-03-17 at 1 05 16 PM

coupler debug plots (intentionally blank, with extrema in the titles)
Screenshot 2026-03-17 at 1 05 35 PM

## To add in future PRs
- [ ] support Oceananigans + ClimaSeaIce in SCM runs https://github.com/CliMA/ClimaCoupler.jl/issues/1860
- [ ] update ClimaCore to handle dss of PointSpace fields - https://github.com/CliMA/ClimaCoupler.jl/issues/1859
- [ ] update plotting: make timeseries of 1D fields (scalars) if we have multiple diagnostics saved https://github.com/CliMA/ClimaCoupler.jl/issues/1862
- [ ] add a PointSpace constructor in ClimaCore that accepts lat/lon so we don't have to do gymnastics in constructing one - ?

----

### DIFF
--- a/docs/src/input.md
+++ b/docs/src/input.md
@@ -185,6 +185,14 @@ specific timesteps should be specified, rather than only `dt`.
 |----------|------|---------|---------------|-------------|
 | `--evolving_ocean` | Bool | `true` | `true`, `false` | Whether to use a dynamic slab ocean model, as opposed to constant surface temperatures |
 
+#### Single-column model (SCM) settings
+
+| Argument | Type | Default | Valid Options | Description |
+|----------|------|---------|---------------|-------------|
+| `--domain_type` | String | `"global"` | `global`, `column` | Domain type for the simulation. Use `column` for single-column model runs |
+| `--column_latlon` | Vector{Float64} | `[0.0, 0.0]` | `[lat, lon]` in degrees | Latitude within [-90, 90] and longitude within [-180, 180] for the SCM column location |
+| `--scm_surface_type` | String | `nothing` | `land`, `ocean`, `sea_ice` (required for `column`) | Select SCM surface model: enables exactly one of land, ocean, or sea ice. Ignored for global runs. |
+
 ## Input API
 
 ```@docs
--- a/ext/ClimaCouplerClimaAtmosExt.jl
+++ b/ext/ClimaCouplerClimaAtmosExt.jl
@@ -633,6 +633,9 @@ function get_atmos_config_dict(
         atmos_config["toml"] = toml
     end
 
+    # If running in SCM mode, set the atmosphere space type and lat/lon
+    coupler_config["domain_type"] == "column" && (atmos_config["config"] = "column")
+
     # Override atmos parameters with coupled parameters
     FT = atmos_config["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
     override_file = CP.merge_toml_files(atmos_config["toml"], override = true)
@@ -692,11 +695,9 @@ one here.
 """
 function dss_state!(sim::ClimaAtmosSimulation)
     Y = sim.integrator.u
-    for key in propertynames(Y)
-        field = getproperty(Y, key)
-        buffer = CC.Spaces.create_dss_buffer(field)
-        CC.Spaces.weighted_dss!(field, buffer)
-    end
+    buffer = Utilities.init_dss_buffer(Y)
+    Utilities.apply_dss!(Y, buffer)
+    return nothing
 end
 
 """
--- a/src/Input.jl
+++ b/src/Input.jl
@@ -263,6 +263,19 @@ function argparse_settings()
         help = "Boolean flag indicating whether to use binary (thresholded) area fractions for land and ice [`true` (default), `false`]. When true, land fraction > eps becomes 1, and ice fraction > 0.5 becomes 1."
         arg_type = Bool
         default = true
+        # Single-column model (SCM) settings
+        "--domain_type"
+        help = "Domain type for the simulation. [`global` (default), `column`]"
+        arg_type = String
+        default = "global"
+        "--column_latlon"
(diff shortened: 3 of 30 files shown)


docs/src/input.md
```diff
@@ -183,6 +183,14 @@ specific timesteps should be specified, rather than only `dt`.
 |----------|------|---------|---------------|-------------|
 | `--evolving_ocean` | Bool | `true` | `true`, `false` | Whether to use a dynamic slab ocean model, as opposed to constant surface temperatures |
 
+#### Single-column model (SCM) settings
+
+| Argument | Type | Default | Valid Options | Description |
+|----------|------|---------|---------------|-------------|
+| `--domain_type` | String | `"global"` | `global`, `column` | Domain type for the simulation. Use `column` for single-column model runs |
+| `--column_latlon` | Vector{Float64} | `[0.0, 0.0]` | `[lat, lon]` in degrees | Latitude and longitude for the SCM column location |
```
ph-kev: Can you specify the bounds for the latitudes and longitudes, especially the longitudes since it is either 0 to 360 or -180 to 180 degrees?

src/Input.jl
```diff
@@ -263,6 +263,19 @@ function argparse_settings()
         help = "Boolean flag indicating whether to use binary (thresholded) area fractions for land and ice [`true` (default), `false`]. When true, land fraction > eps becomes 1, and ice fraction > 0.5 becomes 1."
         arg_type = Bool
         default = true
+        # Single-column model (SCM) settings
+        "--domain_type"
+        help = "Domain type for the simulation. [`global` (default), `column`]"
+        arg_type = String
+        default = "global"
+        "--column_latlon"
+        help = "Latitude and longitude (degrees) for SCM column as [lat, lon]. [[0.0, 0.0] (default)]"
```
ph-kev: Same comment from before for specifying the bounds on the latitudes and longitudes.

nefrathenrici:
> Note that the single column right now is not correct for the atmosphere, so we are using a box: [CliMA/ClimaAtmos.jl#4209](https://github.com/CliMA/ClimaAtmos.jl/pull/4209). @nefrathenrici knows more (maybe we should open an issue).

Yeah, ClimaAtmos column configuration runs using an `ExtrudedFiniteDifferenceGrid` instead of a pure 1D geometry because the conversions are broken in ClimaCore Geometry. This is intended to be fixed in the Geometry module [refactor](https://github.com/CliMA/ClimaCore.jl/pull/2457) that @dennisYatunin started.

szy21:
> > > Note that the single column right now is not correct for the atmosphere, so we are using a box: [CliMA/ClimaAtmos.jl#4209](https://github.com/CliMA/ClimaAtmos.jl/pull/4209). @nefrathenrici knows more (maybe we should open an issue).
> > 
> > 
> > Yeah, ClimaAtmos column configuration runs using an `ExtrudedFiniteDifferenceGrid` instead of a pure 1D geometry because the conversions are broken in ClimaCore Geometry. This is intended to be fixed in the Geometry module [refactor](https://github.com/CliMA/ClimaCore.jl/pull/2457) that @dennisYatunin started.
> 
> Yeah - this PR uses the same setup for each component that they use for their offline column runs. So the land uses a column and the atmosphere uses its box-column. The coupler uses a point space for coupler fields and flux calculations

Oh, that's probably ok then, as long as the coupler flux calculations have the correct conversions.

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -633,6 +633,13 @@ function get_atmos_config_dict(
         atmos_config["toml"] = toml
     end
 
+    # If running in SCM mode, set the atmosphere space type and lat/lon
+    if coupler_config["domain_type"] == "column"
+        atmos_config["config"] = "column"
+        atmos_config["site_latitude"] = first(coupler_config["column_latlon"])
+        atmos_config["site_longitude"] = last(coupler_config["column_latlon"])
```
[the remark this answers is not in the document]
juliasloan25: it looks like these are only used for ERA5 forcing so I'll remove them here

PR #1804, opened 2026-03-12
Config updates

trontrytel:
- For this weekend scheduled longruns I want to swap ED only run with prognostic EDMF 1M run. And see how stable it actually is, after all the fixes from the last two weeks.
- I also want to use ARS343 when running with diagnostic EDMF. It seems to be more stable

### DIFF
--- a/config/longrun_configs/amip_diagedmf_1m_land.yml
+++ b/config/longrun_configs/amip_diagedmf_1m_land.yml
@@ -20,7 +20,7 @@ land_model: "integrated"
 mode_name: "amip"
 netcdf_output_at_levels: true
 microphysics_model: "1M"
-ode_algo: "ARS232"
+ode_algo: "ARS343"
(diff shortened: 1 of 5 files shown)


szy21 (commented):
You can leave the EDonly one in the pipeline. That's the only longrun with bucket land now. I think we have enough gpus to run them.

juliasloan25:
> Any ideas why the downgrade test is flagged here?

I'm not sure why it just started happening (and across all coupler PRs), but our downgrade test used to use ArgParse v1.2.0 and now they're using v1.0.0. [This commit](https://github.com/CliMA/ClimaCoupler.jl/pull/1808/changes/62f1ddec245e210dbae8402510177e48bba1c9b2) should fix it, or if it doesn't we can increase the ArgParse compat to "1.2" (but I prefer to keep the less strict compat if the fix does work)

PR #1816, opened 2026-03-16
compute fluxes for the bucket in the coupler

juliasloan25:
## Purpose 
Instead of computing fluxes within the bucket's step. This way the coupler handles flux calculation of each `ExplicitFluxSimulation`, and we can accumulate fluxes for each of them.

Requires https://github.com/CliMA/ClimaLand.jl/pull/1677 and a ClimaLand release

Maybe we want to go back to what we had in https://github.com/CliMA/ClimaCoupler.jl/commit/288987875b84413fc01e5de6a8826f11abab9261#diff-a2880400c2145db122968b23c23a980c061601611b3bfe9e2cce55c10f147c73 ?
## Content

----


ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -379,78 +376,62 @@ function FluxCalculator.compute_surface_fluxes!(
...
-    # When L_MO is infinite, avoid multiplication by zero to prevent NaN
-    @. csf.L_MO +=
-        ifelse(isinf(csf.scalar_temp1), csf.scalar_temp1, csf.scalar_temp1 * area_fraction)
 
+    # Get atmosphere fields
+    uv_int = StaticArrays.SVector.(csf.u_int, csf.v_int)
+
+    # Get surface fields we need to compute surface fluxes
+    # compute surface humidity from the surface temperature, surface density, and phase
+    Interfacer.get_field!(csf.scalar_temp1, sim, Val(:surface_temperature))
```
juliasloan25: copy this from src/shared_utilities.drivers.jl turbulent_fluxes! in land

juliasloan25:
see #1865 instead

PR #1827, opened 2026-03-23
feat: expose DateTime clock parameter for CMIP simulations (#1819)

### LINKED ISSUE #1819: use DateTimes in the Clock for CMIP
Right now we use the default clock constructed in `ClimaOcean.ocean_simulation`, which uses Float64. We should expose this in the `ocean_simulation` constructor, then construct one that uses DateTimes so we have better precision.

We also need to pass it to the ClimaSeaIce model [here](https://github.com/CliMA/ClimaCoupler.jl/blob/8614f1e40ff90cfcfea17ca51f054d6063161efe/ext/ClimaCouplerCMIPExt/clima_seaice.jl#L206). This is already available so we don't need to change the constructor.

KavyaKapoor420:
## Purpose 
Replace the default Float64-based clock with a DateTime-based clock in ocean and sea ice simulations to improve temporal precision for CMIP runs.  
Closes #1819

## To-do
- [x] Add DateTime-based clock support to ocean and sea ice simulations  
- [x] Update time-stepping logic to use DateTime arithmetic  
- [x] Update docstrings  
- [x] Run tests and formatting checks  

## Content
- Added `start_date::DateTime` to:
  - `OceananigansSimulation`
  - `ClimaSeaIceSimulation`

- Updated constructors to initialize clocks using:
  `OC.Clock{DateTime}(time = start_date)`

- Replaced Float64-based time stepping with DateTime-based Δt computation in:
  - `Interfacer.step!` for ocean
  - `Interfacer.step!` for sea ice

- Ensured consistent use of DateTime clocks across both components

## Testing
- Ran test suite using:
  `julia --project -e 'using Pkg; Pkg.test()'`

- Verified:
  - Simulations initialize with `OC.Clock{DateTime}`
  - Time stepping behaves correctly

- Code formatted using JuliaFormatter

----

Snapshots of Test passed 
---

Screenshot 2026-03-24 at 1 27 34 AM

### DIFF
--- a/ext/ClimaCouplerCMIPExt/clima_seaice.jl
+++ b/ext/ClimaCouplerCMIPExt/clima_seaice.jl
@@ -32,6 +32,7 @@ struct ClimaSeaIceSimulation{SIM, A, REMAP, NT, IP} <: Interfacer.AbstractSeaIce
     remapping::REMAP
     ocean_ice_interface::NT
     ice_properties::IP
+    start_date::DateTime
 end
 
 """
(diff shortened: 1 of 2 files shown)


juliasloan25:
done in main

PR #1829, opened 2026-03-24
Enable GPU outputs for OC > 0.96.22

simone-silvestri:
If we have OC > 0.96.22 we can use NetCDFWriter also from the GPU.
Also, sea ice should use the same writer as the ocean

cc @jm-c


akshaysridhar:
Superseded by https://github.com/CliMA/ClimaCoupler.jl/pull/1898

PR #1834, opened 2026-03-26
Move AMIP longrun to central

nefrathenrici:
## Content

----


nefrathenrici:
It will choose a P100 by default. We can set it to use the L40S instead with slurm_gres: gpu:l40s:1 .

szy21:
> It will choose a P100 by default. We can set it to use the L40S instead with slurm_gres: gpu:l40s:1 .

So that will probably be slow, right? Maybe we should set it to L40S?

PR #1850, opened 2026-03-30
Fix climacommon mismatch for calibration

ph-kev:
The subseasonal calibration pipeline was using `climacommon/2026_02_18` but the pipeline was using `climacommon/2025_05_15`.

### DIFF
--- a/experiments/calibration/subseasonal/run_calibration.jl
+++ b/experiments/calibration/subseasonal/run_calibration.jl
@@ -24,9 +24,18 @@ model_interface = joinpath(
     "model_interface.jl",
 )
 
+"""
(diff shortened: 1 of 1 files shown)


ph-kev:
> The fix looks good, but this seems pretty prone to error. The modules should be selected where we select resources/set up the run - why do we need a method to do it in ClimaCalibrate?

See this [issue](https://github.com/CliMA/ClimaCalibrate.jl/issues/287) in ClimaCalibrate. The method only exists since the backend submits slurm/PBS job scripts so the modules also also need to be specified there. I am not sure why a keyword argument for specifying the `climacommon` isn't exposed for this.

PR #1865, opened 2026-04-07
compute bucket fluxes in compute_surface_fluxes

juliasloan25:
## Purpose 
As opposed to in the bucket model's step (which is how it is done now). This version continues to use the `CL.turbulent_fluxes!` method, as opposed to #1816 which copies the internals of that function into the coupler and as a result is much more complicated.

Also adds some functions to simplify the flux calculation code structure.

I find it odd that we were calling `CL.turbulent_fluxes!` from the bucket's `compute_surface_fluxes!` method, even though the fluxes were already computed in the bucket's step. I expect this was just extra computation but didn't change the values.

Requires https://github.com/CliMA/ClimaLand.jl/pull/1677 + land release

## Content
- [x] Add `get_roughness_params` to reduce code duplication
- [x] Add `update_flux_fields!` to split work into smaller functions with clear roles (and reduce code duplication)
- [x] Use new land branch, where fluxes aren't computed in `step!`

## Dependency updates
```
    Updating `~/clima/ClimaCoupler.jl/experiments/CMIP/Project.toml`
  [b2c96348] ↑ ClimaAtmos v0.38.0 ⇒ v0.38.1
  [1ecacbb8] ↑ ClimaDiagnostics v0.3.3 ⇒ v0.3.4
  [08f4d4ce] ↑ ClimaLand v1.6.5 ⇒ v1.7.0
  [0376089a] ↑ ClimaOcean v0.9.4 ⇒ v0.9.5
  [85f8d34a] ↑ NCDatasets v0.14.14 ⇒ v0.14.15
```
I accidentally switched the `Manifest.toml` files from julia 1.10 to 1.12 earlier this week, so this PR switches them back to 1.10.

----


ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -376,18 +376,72 @@ function FluxCalculator.compute_surface_fluxes!(
...
+    return nothing
+end
+
+"""
+    FluxCalculator.update_flux_fields!(csf, sim::BucketSimulation, fluxes)
 
-    # Get area fraction of the land model (min = 0, max = 1)
+Update the surface simulation `sim` with the values stored in `fluxes,
+without any area-weighting. Then, update the coupler fields `csf` with
+the area-weighted fluxes.
```
kmdeck:
```suggestion
the area-weighted fluxes after remapping to the coupler space.
```

src/FluxCalculator.jl
```diff
@@ -306,21 +290,38 @@ function compute_surface_fluxes!(
             csf.q_liq_atmos,
             csf.q_ice_atmos,
             csf.ρ_atmos,
-            csf.height_int,
-            uv_int .* FT(0),
+            csf.height_int, # h_int
+            uv_int .* FT(0), # uv_sfc
```
kmdeck, line 297: Should this be hardcoded (will surface components ever have nonzero wind speed at the surface)?

src/FluxCalculator.jl
```diff
@@ -306,21 +290,38 @@ function compute_surface_fluxes!(
...
             T_sfc,
             q_sfc,
-            csf.height_sfc,
-            FT(0),
+            csf.height_sfc, # h_sfc
+            FT(0), # d
```
kmdeck, line 301: Should this be hardcoded?

src/FluxCalculator.jl
```diff
@@ -357,6 +358,49 @@ function compute_surface_fluxes!(
     return nothing
 end
 
+get_fluxes(sim::Interfacer.AbstractSurfaceSimulation, fluxes) = fluxes
```
kmdeck: this function seems a little odd!

src/FluxCalculator.jl
```diff
@@ -357,6 +358,49 @@ function compute_surface_fluxes!(
...
+
+Load momentum and buoyancy roughness from the simulation into csf scratch fields
+and return element-wise `ConstantRoughnessParams`.
+"""
+function get_roughness_params(csf, sim, ::Val{:constant})
+    Interfacer.get_field!(csf.scalar_temp3, sim, Val(:roughness_momentum))
+    z0m = csf.scalar_temp3
+    Interfacer.get_field!(csf.scalar_temp4, sim, Val(:roughness_buoyancy))
+    z0b = csf.scalar_temp4
+    return SF.ConstantRoughnessParams.(z0m, z0b)
```
kmdeck, line 393: I think this will allocate a new field each step. is it possible to store a field of type ConstantRoughnessParams instead?

kmdeck: or maybe this can be lazy

src/FluxCalculator.jl
```diff
@@ -277,24 +277,8 @@ function compute_surface_fluxes!(
...
-        z0m = csf.scalar_temp3
-        Interfacer.get_field!(csf.scalar_temp4, sim, Val(:roughness_buoyancy))
-        z0b = csf.scalar_temp4
-        SF.ConstantRoughnessParams.(z0m, z0b)
-    else
-        error("Unknown roughness_model: $roughness_model. Must be :coare3 or :constant")
-    end
-
+    roughness_params = get_roughness_params(csf, sim)
     config = SF.SurfaceFluxConfig.(roughness_params, SF.ConstantGustinessSpec.(gustiness))
```
kmdeck, line 282: this will allocate also. Should this be lazy?

src/FluxCalculator.jl
kmdeck: If it always will be zero, maybe it is better to not create/allocate a field here and instead pass in something like `zero(eltype(uv_int))`

experiments/CMIP/Project.toml
```diff
@@ -49,7 +49,7 @@ ClimaComms = "0.6.2"
 ClimaCore = "0.14.46"
 ClimaCoreMakie = "0.4"
 ClimaDiagnostics = "0.2.6, 0.3"
-ClimaLand = "1.6"
+ClimaLand = "1.6, 1.7"
```
kmdeck: Will this still work with 1.6 if we compute the bucket fluxes inside land in that? I guess it's how it is running now..

kmdeck: but Im wondering if the compat should be >=1.7

src/FluxCalculator.jl
```diff
@@ -277,24 +277,8 @@ function compute_surface_fluxes!(
     # Set some scalars that we hardcode for now
     gustiness = ones(boundary_space)
```
kmdeck, line 278: should this be gotten from the atmosphere model parameters?

kmdeck (approved):
Looks good, left some comments that you can resolve as makes sense to you!

experiments/CMIP/Project.toml
juliasloan25: Good catch - I think it should be 1.7 only. Maybe it would technically work with 1.6 and we would just recompute the fluxes, but I think it's more clear to restrict to 1.7

src/FluxCalculator.jl
juliasloan25: ocean may have non-zero surface velocity, but land won't. I set a temp field to 0 and passed it in here so we don't allocate

src/FluxCalculator.jl
juliasloan25: I think it's okay because this function isn't used for land. It probably would be better to not hardcode it, but I think it's ok for now

src/FluxCalculator.jl
juliasloan25: Oops yes I meant to remove this. Thank you!

src/FluxCalculator.jl
juliasloan25: Good points. I'll open an issue about removing these allocations since they weren't introduced in this PR, just moved - see https://github.com/CliMA/ClimaCoupler.jl/issues/1868

src/FluxCalculator.jl
juliasloan25: The atmosphere doesn't provide gustiness currently so we just hardcode it. Eventually we'll want to change this

src/FluxCalculator.jl
juliasloan25: Actually my change errored because we only have scalar temp fields, and `uv` is a StaticVector. I updated the issue related to the other allocations you pointed out in this PR to include this too - https://github.com/CliMA/ClimaCoupler.jl/issues/1868

PR #1867, opened 2026-04-08
Fix ocean polar mask bug

### LINKED ISSUE #1866: Bug in polar mask
While working on #1800, I think I've found a bug in `ocean_flux_highlat_mask` [here](https://github.com/CliMA/ClimaCoupler.jl/blob/dac5ca3dba7ce632b2890cfbd3ecf1133f43b0e3/ext/ClimaCouplerCMIPExt/oceananigans.jl#L372-L381). In the `experiments/CMIP` environment, I've run
```Julia
julia> include("experiments/CMIP/code_loading.jl")
julia> config_file = "config/ci_configs/cmip_oceananigans_climaseaice.yml"
julia> cs = CoupledSimulation(config_file)
```
and after the setup I find that the masks are just zero, e.g.:
```julia
julia> cs.model_sims.ocean_sim.remapping.polar_exclusion_flux_mask_centers
1×1 CUDA.CuArray{Float64, 2, CUDA.DeviceMemory}:
 0.0
```
I think this is because (1) `OC.φnodes` is already in degrees, so applying `rad2deg` on it makes everything much smaller than 78 and (2) it is just a `StepRangeLen` so all of the reshaping lines just reduce it to a single number.

I believe this means that the ocean fluxes have been *completely* masked from the simulation, not just at the poles! I'll work on fixing both this and #1800, but since we're trying to move away from this masking anyways it may not be all that necessary.

hgpeterson:
Closes #1866

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,26 +19,26 @@ steps:
   - label: "init environment :computer:"
     key: "init_cpu_env"
     command:
-      # - echo "--- Instantiate package env"
-      # - "julia -O0 --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      # - "julia -O0 --project -e 'using Pkg; Pkg.precompile()'"
-      # - "julia -O0 --project -e 'using Pkg; Pkg.status()'"
+      - echo "--- Instantiate package env"
+      - "julia -O0 --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia -O0 --project -e 'using Pkg; Pkg.precompile()'"
+      - "julia -O0 --project -e 'using Pkg; Pkg.status()'"
 
-      # - echo "--- Instantiate AMIP experiments env"
-      # - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      # - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.precompile()'"
-      # - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.status()'"
+      - echo "--- Instantiate AMIP experiments env"
+      - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.precompile()'"
+      - "julia -O0 --project=experiments/AMIP/ -e 'using Pkg; Pkg.status()'"
 
       - echo "--- Instantiate CMIP experiments env"
       - "julia -O0 --project=experiments/CMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
       - "julia -O0 --project=experiments/CMIP/ -e 'using Pkg; Pkg.precompile()'"
       - "julia -O0 --project=experiments/CMIP/ -e 'using Pkg; Pkg.status()'"
 
-      # - echo "--- Instantiate test env"
-      # - "julia -O0 --project=test/ -e 'using Pkg; Pkg.develop(path=\".\")'"
-      # - "julia -O0 --project=test/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
akshaysridhar, line 1: Please uncomment these pipeline entries so we can run them in regular CI. Once approved, we also tend to squash commits prior to merging. Thanks @hgpeterson!

hgpeterson: Woops, should be fixed now!

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -371,11 +371,9 @@ unlike the polar mask which is defined on the boundary_space)
...
 function ocean_flux_highlat_mask(grid; location = (OC.Center(), OC.Center(), OC.Center()))
     polar_flux_lat_deg = 78.0  # zero fluxes where |lat| ≥ this (same band as polar_mask)
-    φ = OC.φnodes(grid, location[1], location[2], location[3])
-    φ_2D = Array(φ[:, :, 1])
-    lat_deg = abs.(rad2deg.(φ_2D))
-    mask = ifelse.(lat_deg .< polar_flux_lat_deg, 1.0, 0.0)
-    mask = reshape(mask[1, :], 1, :)
+    φ = OC.φnodes(grid, location[1], location[2], location[3])  # in degrees
+    mask = ifelse.(abs.(φ) .< polar_flux_lat_deg, 1.0, 0.0)
+    mask = repeat(mask', grid.Nx, 1) # Nx × Ny
```
akshaysridhar: cc @juliasloan25 . This LGTM. Thanks @hgpeterson - the coarse run from this morning's CMIP meeting runs at least for 18 weeks (simple-ocean, ClimaAtmos ED-only, bucket land configuration)

akshaysridhar (approved):

akshaysridhar:
Thanks @hgpeterson ; squash + rebase onto 1 commit and we can merge this in!

ext/ClimaCouplerCMIPExt/oceananigans.jl
juliasloan25: could you add a comment here explaining the dimensions of `φ` and `mask`, and why we use `repeat` here?

juliasloan25 (approved):
This looks good to me!

I'm thinking about what kind of simple test we can implement that may have caught this before. We have unit tests in `test/` which test only the coupler source code, and we also have some slightly-larger-scale tests in `experiments/test/` which test the AMIP components and some flux calculations. We could do something similar for CMIP - e.g. a test in `experiments/CMIP/test_masks.jl` that makes an ocean grid and the polar mask for it, and checks the dimensions of the mask match the dims of the lat/lon coords of the grid.

Since we'll drop the polar mask soon we don't need to add that here if it sounds like too much work, but we should keep in mind adding more unit tests for CMIP.

akshaysridhar:
> This looks good to me!
> 
> I'm thinking about what kind of simple test we can implement that may have caught this before. We have unit tests in `test/` which test only the coupler source code, and we also have some slightly-larger-scale tests in `experiments/test/` which test the AMIP components and some flux calculations. We could do something similar for CMIP - e.g. a test in `experiments/CMIP/test_masks.jl` that makes an ocean grid and the polar mask for it, and checks the dimensions of the mask match the dims of the lat/lon coords of the grid.
> 
> Since we'll drop the polar mask soon we don't need to add that here if it sounds like too much work, but we should keep in mind adding more unit tests for CMIP.

Agreed - In general it would be worth reviewing which uncovered blocks from our [codecov report](https://app.codecov.io/gh/CliMA/ClimaCoupler.jl/tree/main) are critical and make sure we assign meaningful unit and physics tests there; I can work with @hgpeterson to begin cleaning this up.

ext/ClimaCouplerCMIPExt/oceananigans.jl
hgpeterson: How's this? https://github.com/CliMA/ClimaCoupler.jl/pull/1867/commits/65cb80a48b4b253e45cda43880b824784200fb5d

hgpeterson:
Yeah it'd be nice to have more tests on each of the internal steps.

PR #1877, opened 2026-04-20
Fix command to run simulation in README

jefftyoder:
Update path to `AMIP/run_simulation.jl` shown in AMIP README.md.

### DIFF
--- a/experiments/AMIP/README.md
+++ b/experiments/AMIP/README.md
@@ -8,7 +8,7 @@ ClimaSeaIce, KernelAbstractions). For CMIP, see `experiments/CMIP/`.
 ## Running
 
 ```bash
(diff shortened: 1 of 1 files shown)


experiments/AMIP/README.md
```diff
@@ -8,7 +8,7 @@ ClimaSeaIce, KernelAbstractions). For CMIP, see `experiments/CMIP/`.
 ## Running
 
 ```bash
-julia --project=experiments/AMIP experiments/AMIP/run_simulation.jl
```
juliasloan25, line 11: I think the previous line was actually correct - first we set the project to use, then run the runscript/driver. Did you have issues running this?

juliasloan25:
I'm closing this as I don't think it's needed. If you're still having trouble we can revisit!

PR #1878, opened 2026-04-20
Add time incrementing test

imreddyTeja:
## Purpose 
Add a test for Float64 time at high values. 

## Content
Uses a fake integrator, and starts with t = 10^17. Then it steps many times, and checks that the fake integrator is being stepped as expected.

----

### DIFF
--- a/experiments/test/time_increment_test.jl
+++ b/experiments/test/time_increment_test.jl
@@ -1,0 +1,36 @@
+using Test
+using ClimaCoupler
+import ClimaAtmos
+ClimaAtmosExt = Base.get_extension(ClimaCoupler, :ClimaCouplerClimaAtmosExt)
+
+@testset "Float64 time" begin
+    FT = Float64
+    t = FT(10^17)
+    smallest_dt = FT(60)
+    # use a fake sim and integrator because their functionality is not needed in this test
+    struct FakeIntegrator{FT}
+        dt::FT
+        t::Ref{FT}
+    end
+    Base.getproperty(integrator::FakeIntegrator, s::Symbol) =
+        s == :t ? getfield(integrator, s)[] : getfield(integrator, s)
+    fake_sim = ClimaAtmosExt.ClimaAtmosSimulation(
+        nothing,
+        nothing,
+        FakeIntegrator(6 * smallest_dt, Ref(t)),
+        nothing,
+    )
+    function ClimaCoupler.Interfacer.step!(sim::FakeIntegrator, _, _)
+        getfield(sim, :t)[] = t
+        return
+    end
+    for i in 1:(10^5)
+        t += smallest_dt
+        ClimaCoupler.Interfacer.step!(fake_sim, t)
+        if i % 6 == 0
+            @test fake_sim.integrator.t == t
+        else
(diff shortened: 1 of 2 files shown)


experiments/test/time_increment_test.jl
```diff
@@ -0,0 +1,36 @@
...
+
+@testset "Float64 time" begin
+    FT = Float64
+    t = FT(10^17)
+    smallest_dt = FT(60)
+    # use a fake sim and integrator because their functionality is not needed in this test
+    struct FakeIntegrator{FT}
+        dt::FT
+        t::Ref{FT}
+    end
```
ph-kev, line 14: I don't think you can assume that `FT` is a `Float64`. If you initialize a `AtmosSimulation` with the float type being `Float32`, the float type of time is also `Float32`.

experiments/test/time_increment_test.jl
```diff
@@ -0,0 +1,36 @@
...
+        nothing,
+        nothing,
+        FakeIntegrator(6 * smallest_dt, Ref(t)),
+        nothing,
+    )
+    function ClimaCoupler.Interfacer.step!(sim::FakeIntegrator, _, _)
+        getfield(sim, :t)[] = t
+        return
+    end
+    for i in 1:(10^5)
```
ph-kev, line 27: I think this test would be more useful if it exercises the coupler time stepping loop for all simulations. I don't see much value from how test it is right now since it doesn't test something specific in coupler and test properties of Float64.

experiments/test/time_increment_test.jl
```diff
@@ -0,0 +1,36 @@
...
+    end
+    Base.getproperty(integrator::FakeIntegrator, s::Symbol) =
+        s == :t ? getfield(integrator, s)[] : getfield(integrator, s)
+    fake_sim = ClimaAtmosExt.ClimaAtmosSimulation(
+        nothing,
+        nothing,
+        FakeIntegrator(6 * smallest_dt, Ref(t)),
+        nothing,
+    )
+    function ClimaCoupler.Interfacer.step!(sim::FakeIntegrator, _, _)
```
ph-kev, line 23: Can you also keep track of the number of steps? I think it is simpler to debug the total number of steps rather than floating point time which can accumulate errors.

experiments/test/time_increment_test.jl
imreddyTeja: The step! functions for the sub models do not support Float32

daverumph: This is a coupler test, and Julia has already said that the coupler doesn't need to support Float32.

experiments/test/time_increment_test.jl
daverumph: We could create a coupler instance with fake component models.  The land model would be essentially the same as the atmos model.  The ocean and ice models would need to be the oceananigans equivalent.  That would be more of an integration test.

experiments/test/time_increment_test.jl
daverumph: I'm not sure what would need to be done to handle restart.  Do you know?
Float64 time doesn't have a significant issue with precision in the ranges we need to support.

PR #1879, opened 2026-04-20
Add area fracs to coupler fields diags

### LINKED ISSUE #1549: Add land and sea ice area fraction in coupler diagnostics
This would be useful for debugging and plotting.

hgpeterson:
Closes #1549

Changes:
- The `CoupledSimulation` object now has the `land_area_fraction`, `ocean_area_fraction` and `ice_area_fraction` fields defined on the exchange grid.
- These fields are initialized in the `CoupledSimulation` constructor, updated every time `update_surface_fractions!` is called, and output periodically (land only at the beginning using a custom `OnceSchedule` struct implementing a `ClimaDiagnostics` `AbstractSchedule`).
- These changes are reflected in the docs.

Some potential todos/questions:
- Right now the ocean and sea-ice area fractions will be output regularly even for simulations that don't have these components. Maybe it would make more sense to have some `haskey(cs.model_sims, :ocean_sim)` checks, for instance?
- Looking at the output files, I see that these area fractions are not necessarily between 0 and 1. I think this is due to the way a `ClimaCore` `Field` is interpolated onto a latlon grid for NetCDF output. Would it instead be more useful to output the `Field`s themselves?
- It feels a bit hacky to have the land fraction be in the diagnostics scheduler while only having it saved at the beginning. Perhaps I should just save it when the `CoupledSimulation` is constructed and remove the scheduler?

### DIFF
--- a/src/SimOutput/diagnostics.jl
+++ b/src/SimOutput/diagnostics.jl
@@ -4,6 +4,24 @@ import Dates
 
 export diagnostics_setup
 
+#### Custom Schedule for diagnostics that only get output once at the beginning of the simulation
+
+struct OnceSchedule <: CD.Schedules.AbstractSchedule end
+
+function (::OnceSchedule)(integrator)
+    return integrator.step == 1
+end
+
+function CD.Schedules.short_name(::OnceSchedule)
+    return "once"
+end
+
+function CD.Schedules.long_name(::OnceSchedule)
+    return "once at the first step of the simulation"
+end
+
+#### Diagnostics orchestration and setup functions
+
 """
     CD.orchestrate_diagnostics(cs::CoupledSimulation)
 
@@ -21,7 +39,8 @@ end
 
 Set up the default diagnostics for an AMIP simulation, using ClimaDiagnostics.
 The diagnostics are saved to NetCDF files. Currently, this just includes a
-diagnostic for turbulent energy fluxes.
+diagnostic for turbulent energy fluxes and diagnostics for each of the land, ocean,
+and sea-ice area fractions.
 
 Return a DiagnosticsHandler object to coordinate the diagnostics.
 """
@@ -33,10 +52,14 @@ function diagnostics_setup(
     diagnostics_dt,
     coupled_dt,
 )
-    # Create schedules and writer
-    schedule_everystep = CD.Schedules.EveryStepSchedule()
-    schedule_calendar_dt = CD.Schedules.EveryCalendarDtSchedule(diagnostics_dt; start_date)
-    netcdf_writer = CD.Writers.NetCDFWriter(axes(fields.F_sh), output_dir)
+    # Create a list to hold the scheduled diagnostics
+    scheduled_diags = []
+
+    # Create output writer (shared across all diagnostics since they all live on the boundary space)
+    boundary_space = axes(fields.F_lh)
+    netcdf_writer = CD.Writers.NetCDFWriter(boundary_space, output_dir)
+
+    #### Turbulent energy fluxes diagnostic
 
     # Create the diagnostic for turbulent energy fluxes
     F_turb_energy_diag = CD.DiagnosticVariable(;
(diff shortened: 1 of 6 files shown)


hgpeterson:
@szy21

src/SimOutput/diagnostics.jl
```diff
@@ -4,6 +4,24 @@ import Dates
...
+    return integrator.step == 1
+end
+
+function CD.Schedules.short_name(::OnceSchedule)
+    return "once"
+end
+
+function CD.Schedules.long_name(::OnceSchedule)
+    return "once at the first step of the simulation"
+end
```
szy21, line 21: Should we move this to ClimaDiagnostics?

src/SimOutput/diagnostics.jl
```diff
@@ -65,9 +93,96 @@ function diagnostics_setup(
...
+
+    #### Land area fraction diagnostic (only at beginning of the simulation since it's static)
+
+    # Create the diagnostic for land fraction
+    land_fraction_diag = CD.DiagnosticVariable(;
+        short_name = "land_fraction",
+        long_name = "Land area fraction",
+        standard_name = "land_fraction",
+        units = "1",
+        comments = "Fraction of each grid cell that is land",
```
szy21:
```suggestion
        short_name = "sftlf",
        long_name = "Percentage of the Grid Cell Occupied by Land (Including Lakes)",
        standard_name = "land_area_fraction",
        units = "%",
        comments = "Fraction of each grid cell that is land",
```
(We are trying to follow cmip variable convention as much as we can: https://airtable.com/appYNLuWqAgzLbhSq/shrKcLEdssxb8Yvcp/tblL7dJkC3vl5zQLb.)

src/SimOutput/diagnostics.jl
```diff
@@ -65,9 +93,96 @@ function diagnostics_setup(
...
+        compute! = (out, state, cache, time) -> begin
+            if isnothing(out)
+                return state.land_area_fraction
+            else
+                out .= state.land_area_fraction
+            end
```
szy21, line 109:
```suggestion
            if isnothing(out)
                return state.land_area_fraction * 100
            else
                out .= state.land_area_fraction * 100
            end
```

src/SimOutput/diagnostics.jl
```diff
@@ -65,9 +93,96 @@ function diagnostics_setup(
...
+        output_schedule_func = schedule_once,  # only output at the first step
+    )
+
+    push!(scheduled_diags, land_fraction_diag_sched)
+
+    #### Ocean area fraction diagnostic
+
+    # Create the diagnostic for ocean fraction
+    ocean_fraction_diag = CD.DiagnosticVariable(;
+        short_name = "ocean_fraction",
```
szy21: I didn't find this variable in the above spreadsheet, but I think it's commonly called `stfof`. Maybe search for it?

src/SimOutput/diagnostics.jl
```diff
@@ -65,9 +93,96 @@ function diagnostics_setup(
...
+        output_schedule_func = schedule_calendar_dt,
+    )
+
+    push!(scheduled_diags, ocean_fraction_diag_sched)
+
+    #### Ice area fraction diagnostic
+
+    # Create the diagnostic for ice fraction
+    ice_fraction_diag = CD.DiagnosticVariable(;
+        short_name = "ice_fraction",
```
szy21: same here, I think it's commonly called `sic`.

szy21 (approved):
Thanks!

src/SimOutput/diagnostics.jl
```diff
@@ -4,6 +4,24 @@ import Dates
 
 export diagnostics_setup
 
+#### Custom Schedule for diagnostics that only get output once at the beginning of the simulation
+
+struct OnceSchedule <: CD.Schedules.AbstractSchedule end
+
```
ph-kev, line 10: Another solution to this is to use the `DivisorSchedule` with a divisor of `Inf` or a large integer if that doesn't work. That also feels a little hacky though.

src/SimOutput/diagnostics.jl
```diff
@@ -65,9 +93,96 @@ function diagnostics_setup(
...
+            end
+        end,
+    )
+
+    # Schedule the ice fraction to save and output at diagnostic frequency 
+    ice_fraction_diag_sched = CD.ScheduledDiagnostic(
+        variable = ice_fraction_diag,
+        output_writer = netcdf_writer,
+        compute_schedule_func = schedule_calendar_dt,
+        output_schedule_func = schedule_calendar_dt,
```
ph-kev: This could lead to bugs since `EveryCalendarDtSchedule` is stateful. See https://clima.github.io/ClimaDiagnostics.jl/dev/api/#ClimaDiagnostics.Schedules.EveryCalendarDtSchedule

src/SimOutput/diagnostics.jl
hgpeterson: I see `siconc` for "Sea-Ice Area Percentage (Ocean Grid)" and `siconca` for "Sea-Ice Area Percentage (Atmosphere Grid)". I am just saving the fields on the exchange grid here, so would this be `siconca`?

szy21: Yes, I think `siconca` would be more accurate then.

src/SimOutput/diagnostics.jl
hgpeterson: I didn't see it there either, but [here](https://clipc-services.ceda.ac.uk/dreq/mipVars.html) it is called `stfof` with short name `sea_area_fraction`.

src/SimOutput/diagnostics.jl
hgpeterson: Maybe, or do you think Kevin's suggestion below (i.e., doing `DivsorSchedule(Inf)`) would be better? Or my other idea is to remove the land from the scheduler entirely and just store it when the `CoupledSimulation` object is constructed.

szy21: That feels a bit hacky, but I'm ok with it. I'm not very familiar with diagnostics, but don't you still need a schedule to output land fraction even if you store it in the `CoupledSimulation` object?

src/SimOutput/diagnostics.jl
hgpeterson: Would something more like what I have in https://github.com/CliMA/ClimaCoupler.jl/pull/1879/commits/d7605d35f0ec6b236ff1c508eb04c5025b4bc6aa be better, then?

ph-kev: Yes, it should be fine if you create two different `EveryCalendarDtSchedule`s.

src/SimOutput/diagnostics.jl
hgpeterson: Yeah after trying for a bit it's not immediately obvious to me how to get this to work in the constructor... I think I will keep the `OnceSchedule` method for now. It's more code than Kevin's suggestion, but (at least imho) a little clearer. I'll see if I can get it added to `ClimaDiagnostics`.

szy21: Sounds good. I'm also fine with getting this in and adding it to ClimaDiagnostics later. Up to you.

src/SimOutput/diagnostics.jl
hgpeterson: Discussed [here](https://github.com/CliMA/ClimaCoupler.jl/pull/1879#discussion_r3120534243)

PR #1880, opened 2026-04-21
Add time incrementing test

imreddyTeja:
Use F32 for atmos

## Purpose 
- add a test to show the point at which the coupler and a component model's time will no longer agree

----

### DIFF
--- a/experiments/test/time_increment_test.jl
+++ b/experiments/test/time_increment_test.jl
@@ -1,0 +1,34 @@
+using Test
+using ClimaCoupler
+import ClimaAtmos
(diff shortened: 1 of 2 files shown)


experiments/test/time_increment_test.jl
```diff
@@ -0,0 +1,34 @@
+using Test
+using ClimaCoupler
+import ClimaAtmos
+ClimaAtmosExt = Base.get_extension(ClimaCoupler, :ClimaCouplerClimaAtmosExt)
+
+@testset "Float64 time" begin
```
ph-kev: Can you rename this to something better?

ph-kev (approved):

PR #1882, opened 2026-04-21
Dr/coupler step logic

daverumph:
## Purpose 
Get in some improvements that didn't make it into PR #1794

## Content
Calculate number of steps and use a for loop instead of while

----
- [X] I have read and checked the items on the review checklist.


config/ci_configs/test_concurrent.yml
imreddyTeja, line 1: Is this file used anywhere

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -317,8 +317,11 @@ function Interfacer.step!(sim::BucketSimulation, t::Float64)
     model_t = Float64(sim.integrator.t)
     Δt = t - model_t
```
imreddyTeja: I think Julia recommended changing the name of this var in the last pr.

imreddyTeja (commented):
I think there is something odd happening with git. It says there are 31 commits to merge

config/ci_configs/test_concurrent.yml
daverumph: It wasn't supposed to go into the repo.  I've deleted it.

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
daverumph: Done

imreddyTeja (approved):
This looks good to me. I'm still not sure why it says there are so many commits though

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -315,19 +315,23 @@ end
...
+    # Check to see that we're within 1/8 sec of a time step to avoid floating point issues,
+    # and if so take an integer number of steps to get there
+    if isapprox(time_since_last_model_step, model_dt, atol = 0.125) ||
+       time_since_last_model_step > model_dt
+        n_steps = round(Int, time_since_last_model_step / model_dt)
+        for _ in 1:n_steps
             Interfacer.step!(sim.integrator)
         end
     end
     return nothing
```
juliasloan25, line 329: We can simplify this further as:
```
    model_dt = Float64(sim.integrator.dt)
    n_steps = round(Int, (t - Float64(sim.integrator.t)) / model_dt)
    for _ in 1:n_steps
        Interfacer.step!(sim.integrator, model_dt, true)
    end
    return nothing
```
We may still encounter catastrophic cancellation in the float subtraction `t - Float64(sim.integrator.t`, but as long as the error stays below half of a timestep (which is a very large threshold) the `round(Int, ...)` will make sure we take the correct number of steps.

I'm happy to make this change in a follow-up PR.

PR #1886, opened 2026-04-24
simplify timestepping logic

juliasloan25:
## Purpose 
Simplify timestepping logic when coupler dt is smaller than component models' by performing a loop over an integer number of steps for Float64 and ITime types alike. We no longer hit catastrophic cancellation so the time incrementing test is correct for a larger number of timesteps.

Also adds methods for prescribed sea ice and slab ocean, which were always stepping by the elapsed time rather than by the model's fixed dt, causing them to take partial steps when dt_coupler < dt_model.

Follow-up to #1794 and #1882

## Content
- [x] Float64: compute number of timesteps using integer logic to avoid floating point precision errors
- [x] ITime: take advantage of integer division rounding to remove extra `if` check
- [x] add `step!` methods for prescribed sea ice, slab ocean which handle the case of `dt_coupler < dt_model`
- [x] use consistent time type `TT = Float64` in test
- [x] increase loop limit for test

----


ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -344,27 +344,20 @@ end
...
-    # Check to see that we're within 1/8 sec of a time step to avoid floating point issues,
-    # and if so take an integer number of steps to get there
-    if isapprox(time_since_last_model_step, sim.model_Δt, atol = 0.125) ||
-       time_since_last_model_step > sim.model_Δt
-        n_steps = round(Int, time_since_last_model_step / sim.model_Δt)
-        for _ in 1:n_steps
-            OC.time_step!(sim.ocean, sim.model_Δt)
-        end
+    # `round(Int, ...)` tolerates floating point drift less than `model_dt / 2`
+    n_steps = round(Int, (t - sim.ocean.model.clock.time) / sim.model_Δt)
```
daverumph: Simply rounding doesn't work.  Consider the case where sim.ocean.model.clock.time is still 0.0, sim.model_Δt is 1800.0, and dt_cpl is 360.0 (our typical case).  On the 3rd step, when t reaches 1080.0, the value being rounded, (1080.0 - 0.0) / 360.0 == 0.6, which rounds up to 1.0, and the ocean model takes a step prematurely.  I saw this suggestion from you before and rejected it because of this problem.  I actually spent quite a while trying to figure out how to make it work. That's why the isapprox is there ahead of the call to round, to ensure that we've reached a multiple of the model step time.

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -514,29 +514,19 @@ end
...
 
 function Interfacer.step!(sim::ClimaAtmosSimulation, t::ITime)
-    # Don't step until we've reached a step boundary
-    # (This can happen if the coupler dt is less than this model's)
-    time_since_last_model_step = t - sim.integrator.t
-    if time_since_last_model_step >= sim.integrator.dt
-        while sim.integrator.t < t
-            Interfacer.step!(sim.integrator)
-        end
+    n_steps = div(t - sim.integrator.t, sim.integrator.dt) # integer division; exact for ITime
```
daverumph: I would prefer using // rather than div()

src/Models/prescr_seaice.jl
```diff
@@ -392,8 +393,20 @@ Interfacer.update_field!(
 ) = nothing
 
 # extensions required by FieldExchanger
-Interfacer.step!(sim::PrescribedIceSimulation, t) =
-    Interfacer.step!(sim.integrator, t - sim.integrator.t, true)
+function Interfacer.step!(sim::PrescribedIceSimulation, t::Float64)
+    model_dt = Float64(sim.integrator.dt)
+    # `round(Int, ...)` tolerates floating point drift less than `model_dt / 2`
+    n_steps = round(Int, (t - Float64(sim.integrator.t)) / model_dt)
```
daverumph: Same issue with round() as noted before

src/Models/slab_ocean.jl
```diff
@@ -251,8 +251,20 @@ function Interfacer.update_field!(
...
+function Interfacer.step!(sim::SlabOceanSimulation, t::Float64)
+    model_dt = Float64(sim.integrator.dt)
+    # `round(Int, ...)` tolerates floating point drift less than `model_dt / 2`
+    n_steps = round(Int, (t - Float64(sim.integrator.t)) / model_dt)
+    for _ in 1:n_steps
+        Interfacer.step!(sim.integrator, model_dt, true)
+    end
+end
+function Interfacer.step!(sim::SlabOceanSimulation, t::ITime)
+    n_steps = div(t - sim.integrator.t, sim.integrator.dt) # integer division; exact for ITime
```
daverumph: Consider // instead of div()

experiments/test/time_increment_test.jl
```diff
@@ -4,31 +4,31 @@ import ClimaAtmos
...
+        t::Ref{TT}
     end
     Base.getproperty(integrator::FakeIntegrator, s::Symbol) =
         s == :t ? getfield(integrator, s)[] : getfield(integrator, s)
     fake_sim = ClimaAtmosExt.ClimaAtmosSimulation(
         nothing,
         nothing,
-        # component models can use float32 time
-        FakeIntegrator(Float32(smallest_dt), Ref(Float32(t))),
+        FakeIntegrator(smallest_dt, Ref(t)),
```
ph-kev: I think this should be a `Float32` since the atmos simulation (and probably the land simulation) initializes the time as `Float32` if the float type is `Float32`.

src/Models/slab_ocean.jl
```diff
@@ -251,8 +251,20 @@ function Interfacer.update_field!(
...
+    for _ in 1:n_steps
+        Interfacer.step!(sim.integrator, sim.integrator.dt, true)
+    end
+end
```
ph-kev: Nit: Would it be possible to define a single helper function with the same code as this? I think all `Interfacer.step!` functions involving `ITime` are exactly the same.

experiments/test/time_increment_test.jl
```diff
@@ -4,31 +4,31 @@ import ClimaAtmos
...
         return
     end
-    # 951815 is the number of steps after which the time incrementing test fails
-    for i in 1:951815
+    # Take a sufficiently-large number of steps to test robustness against floating point drift
+    n_steps = 9999999
+    for i in 1:n_steps
         t += smallest_dt
         ClimaCoupler.Interfacer.step!(fake_sim, t)
-        @test fake_sim.integrator.t == t broken = i >= 951815
```
ph-kev, line 32: This is also a nit, but I think this test should also be testing how often the integrator is getting called rather than the current time of the integrator. This is because coupler doesn't have control over how the time stepping is done in the component models, but it can control the number of steps taken by each component model.

experiments/test/time_increment_test.jl
juliasloan25: A few years ago ClimaLand conflated FT with the time type and would sometimes initialize the time as Float32, but this caused a lot of problems and was fixed, so that the time is now either Float64 or ITime. I'm surprised to learn that the atmosphere sets the time type to FT. We should never use Float32 for the tiime

ph-kev: I opened an issue for it in ClimaAtmos: https://github.com/CliMA/ClimaAtmos.jl/issues/4446

src/Models/slab_ocean.jl
juliasloan25: Good idea - I added methods for Float64 and for ITime that work for any model that uses an integrator (so everything except for Oceananigans and ClimaSeaIce)

experiments/test/time_increment_test.jl
juliasloan25: ok I added a check for this, and I also moved the test into test/ since now it's testing the generic method defined in src/

src/surface_stub.jl
```diff
@@ -100,4 +100,5 @@ update_field!(::AbstractSurfaceStub, ::Val{:turbulent_moisture_flux}, field) = n
 
 The stub surface simulation is not updated by this function. Extends `SciMLBase.step!`.
 """
-step!(::AbstractSurfaceStub, _) = nothing
+step!(::AbstractSurfaceStub, ::Float64) = nothing
+step!(::AbstractSurfaceStub, ::ITime) = nothing
```
ph-kev, line 104: You can do this, but this is a stylistic decision.

```suggestion
step!(::AbstractSurfaceStub, ::Union{Float64, ITime}) = nothing
```

ph-kev (approved):

src/surface_stub.jl
juliasloan25: This gave method ambiguities with the methods defined in src/Interfacer.jl because `AbstractSurfaceStub` is more specific than `AbstractComponentModel` and `Union{Float64, ITime}` is less specific than either `Float64` or `ITime`

PR #1888, opened 2026-04-28
Attach a progress callback to the ocean simulation

simone-silvestri:


hgpeterson:
Looks like the build is failing because `ClimaCouplerCMIPExt` doesn't import `Printf`

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -181,6 +181,38 @@ function OceananigansSimulation(
...
+
+        @info msg1 * msg2 * msg3 * msg4 * msg5
+
+        wall_time[] = time_ns()
+
+        return nothing
+    end
+
+    # Attaching a progress function to the ocean
+    add_callback!(ocean, progress, IterationInterval(1))
```
hgpeterson: Looks like we don't have `add_callback!` defined

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -181,6 +181,36 @@ function OceananigansSimulation(
...
+              "maximum(u): ($(round(umax, sigdigits=2)), $(round(vmax, sigdigits=2)), $(round(wmax, sigdigits=2))) m/s, " *
+              "wall time: $(prettytime(step_time))"
+        
+        wall_time[] = time_ns()
+
+        return nothing
+    end
+
+    # Attaching a progress function to the ocean
+    OC.add_callback!(ocean, progress, IterationInterval(1))
```
hgpeterson:
```suggestion
    OC.add_callback!(ocean, progress, OC.IterationInterval(1))
```
?

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -181,6 +181,36 @@ function OceananigansSimulation(
         closure,
     )
 
+    wall_time = Ref(time_ns())
+    
+    function progress(sim)
+        ocean = sim.model
+
+        Tmax = maximum(ocean.tracers.T)
+        Tmin = minimum(ocean.tracers.T)
```
juliasloan25: can we do this for each variable just to be a little less verbose?
```suggestion
        (Tmax, Tmin) = extrema(ocean.tracers.T)
```

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -181,6 +181,36 @@ function OceananigansSimulation(
         closure,
     )
 
+    wall_time = Ref(time_ns())
+    
+    function progress(sim)
```
juliasloan25, line 195:
```suggestion
    """
        progress(sim)

    Output the extrema of some prognostic variables, which can be useful for debugging.
    The frequency with which this is output is determined by the interval passed to
    `OC.add_callback!` below.
    """
    function progress(sim)
```

juliasloan25 (approved):
This looks good once the scope issues are resolved.

@akshaysridhar do you prefer to add a config option for how often this info gets printed rather than hardcoding to 1 for now? i.e. the number that we put in IterationInterval

akshaysridhar:
> This looks good once the scope issues are resolved.
> 
> @akshaysridhar do you prefer to add a config option for how often this info gets printed rather than hardcoding to 1 for now? i.e. the number that we put in IterationInterval

Config argument that populates `IterationInterval` (with some sensible defaults assigned) would be useful while we're debugging - we can eventually use pre-set intervals similar to the atmos `progress_logging_callback` functions.

PR #1893, opened 2026-04-29
Use smaller coupling timestep for CMIP

juliasloan25:
## Purpose 
Right now we use 30 minute timesteps for the coupler, ocean, and sea ice in CMIP. This is causing instabilities because the fluxes are computed very infrequently by the coupler. This PR decreases these dts from 30 to 6 minutes to try to make the simulations more stable.

I'll merge this without running CI since it only changes longruns.

Stable longrun with these changes (manually canceled to free up resources): https://buildkite.com/clima/climacoupler-longruns/builds/1096

### DIFF
--- a/config/longrun_configs/cmip_diagedmf_land.yml
+++ b/config/longrun_configs/cmip_diagedmf_land.yml
@@ -5,11 +5,11 @@ checkpoint_dt: "1months"
 coupler_toml: ["toml/amip_diagedmf.toml"]
 dt_atmos: "120secs" # 2 minutes
 dt_cloud_fraction: "1hours"
-dt_cpl: "1800secs" # 30 minutes
+dt_cpl: "360secs" # 6 minutes
 dt_land: "120secs" # 2 minutes
-dt_ocean: "1800secs" # 30 minutes
+dt_ocean: "360secs" # 6 minutes
 dt_rad: "1hours"
-dt_seaice: "1800secs" # 30 minutes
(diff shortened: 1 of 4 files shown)


hgpeterson:
Nice! Is there a reason you also changed the timesteps in the integrated land setups? I thought the problem was only in the bucket case.

Does this also fix the [crash in the prog. EDM + integrated land case](https://buildkite.com/clima/climacoupler-longruns/builds/1089/canvas?sid=019dc8ce-a51d-4735-bc41-b9678b68da1c&tab=output), or is that just an EDMF problem?

hgpeterson (approved):

akshaysridhar (approved):

juliasloan25:
> Nice! Is there a reason you also changed the timesteps in the integrated land setups? I thought the problem was only in the bucket case.
> 
> Does this also fix the [crash in the prog. EDM + integrated land case](https://buildkite.com/clima/climacoupler-longruns/builds/1089/canvas?sid=019dc8ce-a51d-4735-bc41-b9678b68da1c&tab=output), or is that just an EDMF problem?

This problem is only in the bucket case, but the large coupling dt is likely to make the atmosphere unstable as well. It's probably a good idea to use a setup we're more confident in for now as we debug, and we can try increasing the timestep once we get things working.

This change won't fix the failing prog. EDMF run. That parameterization isn't stable yet and also fails in the AMIP setup. I believe the atmosphere team is working on it

PR #1895, opened 2026-04-30
Add NVTX annotations and profiling runs

### LINKED ISSUE #1430: Add NVTX annotations to profile coupler easily
It would be nice to have `NVTX.@annotate` added to key functions so that we can automatically run the nsight profiler and see more clearly where time is spent in a coupled simulation.

juliasloan25:
## Purpose 
Closes #1430 

[Build without `RUN_PROFILING` set](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/312) (normal benchmarks) -- I have since fixed the `pretty_table` arguments
[Build with `RUN_PROFILING=""`](https://buildkite.com/clima/climacoupler-cpu-gpu-benchmarks/builds/313) (NVTX profiling)
### Content
- [x] add NVTX.jl to AMIP, CMIP environments
- [x] add `NVTX.@annotate` to `step!` of models that don't have annotations internally (everything except ClimaAtmos, ClimaLand)
- [x] add `NVTX.@annotate` to flux calculation and regridding methods in src and ext (`Interfacer.remap`, `Interfacer.remap!`, non-nothing methods of `FluxCalculator.compute_surface_fluxes!`)
- [x] add 1 AMIP and 2 CMIPs with `nsys profile` to buildkite benchmark pipeline, which run for ~4 timesteps each  - e.g. see the ClimaLand profiler run [here](https://github.com/CliMA/ClimaLand.jl/blob/3ce4ef422d5ad5141e209b2d1de9d55f2b7f7d1e/.buildkite/target/pipeline.yml#L142)
- [x] add an ENV var to the benchmarks pipeline, so we run benchmarks by default and profiling if requested
- [x] use `start-later` flag to avoid profiling compilation/initialization - tag the portion of `run!` after 2 steps of compilation with `Profile.@profile`

----


.buildkite/benchmarks/pipeline.yml
```diff
@@ -135,3 +153,75 @@ steps:
...
+    if: build.env("RUN_PROFILING") != null
+    steps:
+      - label: "nsys AMIP with diagnostic EDMF"
+        key: "nsys_amip_diagedmf"
+        command:
+          - mkdir -p output/nsys_amip_diagedmf/artifacts
+          - >
+            nsys profile --start-later=true --capture-range=cudaProfilerApi
+            --trace=nvtx,mpi,cuda,osrt --cuda-memory-usage=true --kill=none
+            --output=output/nsys_amip_diagedmf/artifacts/report
```
juliasloan25: rename report to be less generic

.buildkite/benchmarks/pipeline.yml
```diff
@@ -135,3 +153,66 @@ steps:
...
+
+      - label: "nsys CMIP with diag EDMF + integrated land"
+        key: "nsys_cmip_diagedmf_land"
+        command:
+          - mkdir -p output/nsys_cmip_diagedmf_land/artifacts
+          - >
+            nsys profile --start-later=true --capture-range=cudaProfilerApi --trace=nvtx,mpi,cuda,osrt --cuda-memory-usage=true --kill=none
+            --output=output/nsys_cmip_diagedmf_land/artifacts/report_cmip_diagedmf_land
+            julia --color=yes --project=experiments/CMIP/ experiments/CMIP/profile_simulation.jl
+            --config_file config/longrun_configs/cmip_diagedmf_land.yml --job_id nsys_cmip_diagedmf_land
```
imreddyTeja, line 208: The new nsys steps in this pipeline use the normal config files, but the old benchmarks use specific benchmark configs. Is this intentional?

experiments/CMIP/profile_simulation.jl
imreddyTeja, line 1: this looks pretty similar to experiments/AMIP/profile_simulation.jl. Would it make sense to only have one copy? I'm not sure where that would go though

imreddyTeja (approved):
looks mostly good to me. I left two minor comments

.buildkite/benchmarks/pipeline.yml
juliasloan25: Yeah I wanted to specifically profile the longrun setups we use and I didn't want to duplicate the configs

experiments/CMIP/profile_simulation.jl
juliasloan25: Yeah they are pretty much the same, except one loads `experiments/AMIP/code_loading.jl` and the other loads `experiments/AMIP/code_loading.jl`. I'm not sure what a better design would be. Maybe we could pull everything from line 33 down into a function in `experiments/`, then call it from both?

The two `run_simulation.jl` drivers are also very similar and I haven't come up with a good solution yet

juliasloan25: I'll just merge this for now so we have it, and if we want to redesign the two dirs we can revisit it

PR #1897, opened 2026-05-01
Auto retry lost agent gpu steps on buildkite

imreddyTeja:
## Purpose 
This should automatically retry lost agent jobs. Hopefully this results in less babysitting of buildkite
This PR is basically a copy of  this [atmos pr](https://github.com/CliMA/ClimaAtmos.jl/pull/4458)
## To-do
- make sure this actually works

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -57,6 +57,10 @@ steps:
       - label: "MPI Utilities unit tests"
         key: "utilities_mpi_tests"
         command: "srun julia -O0 --color=yes --project=test/ test/utilities_tests.jl"
+        retry: &retry_policy
+          automatic:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -337,5 +369,6 @@ steps:
   - label: ":chart_with_downwards_trend: build history"
     command:
       - build_history staging # name of branch to plot
```
ph-kev, line 371: Was this removed because it wasn't working anymore?

ph-kev (approved):

.buildkite/pipeline.yml
imreddyTeja: Not sure. I never noticed this step

PR #1898, opened 2026-05-04
Add some averaged diagnostics

simone-silvestri:
for the Oceananigans ocean and the ClimaSeaIce sea ice components


ext/ClimaCouplerCMIPExt.jl
```diff
@@ -34,4 +34,7 @@ include("ClimaCouplerCMIPExt/skin_temperature.jl")
 include("ClimaCouplerCMIPExt/oceananigans.jl")
 include("ClimaCouplerCMIPExt/clima_seaice.jl")
 
+include("ClimaCouplerCMIPExt/ocean_diagnostics.jl")
```
juliasloan25, line 43: Don't these files need to be included before the model files?

ext/ClimaCouplerCMIPExt/seaice_diagnostics.jl
```diff
@@ -0,0 +1,70 @@
...
+                            checkpoint_interval = Dates.Day(90),
+                            filename_prefix = "seaice",
+                            file_splitting_interval = Dates.Day(15))
+
+Attach averaged-output writers to the underlying ClimaSeaIce simulation inside a `ClimaSeaIceSimulation`. 
+Two writers are added to `ice_sim.ice.output_writers`:
+
+1. **Surface diagnostics** (`<prefix>_surface.jld2`): sea-ice concentration, thickness, velocities, and (when present) top 
+   surface temperature, averaged over `surface_averaging_interval`.
+2. **Checkpointer** (`<prefix>_checkpoint`): JLD2 checkpoint of the sea-ice model at `checkpoint_interval`.
```
juliasloan25:
```suggestion
2. **Checkpointer** (`<prefix>_checkpoint.jld2`): JLD2 checkpoint of the sea-ice model at `checkpoint_interval`.
```

juliasloan25: Maybe this is being added as part of @akshaysridhar 's checkpointing PR? To me I would think checkpointing and diagnostics should be set/handled separately

ext/ClimaCouplerCMIPExt/seaice_diagnostics.jl
```diff
@@ -0,0 +1,70 @@
...
+
+    hi = ice.model.ice_thickness
+    ℵi = ice.model.ice_concentration
+    ui, vi = ice.model.velocities
+
+    sitemptop = try
+        ice.model.ice_thermodynamics.top_surface_temperature
+    catch
+        nothing
+    end
```
juliasloan25: Can you add a comment explaining why we need the try/catch? i.e. when would `top_surface_temperature` not be present?

ext/ClimaCouplerCMIPExt/ocean_diagnostics.jl
```diff
@@ -0,0 +1,121 @@
...
+"""
+    add_ocean_diagnostics!(ocean_sim::OceananigansSimulation;
+                           output_dir = ".",
+                           surface_averaging_interval = Dates.Day(1),
+                           field_averaging_interval = Dates.Day(15),
+                           checkpoint_interval = Dates.Day(90),
+                           filename_prefix = "ocean",
+                           file_splitting_interval = Dates.Day(15))
+
+Attach averaged-output writers to the underlying Oceananigans simulation inside an `OceananigansSimulation`.
```
juliasloan25: Is there a way to use this function to add instantaneous diagnostics or will they always be averaged? It would be nice for it to be as general as possible.

juliasloan25 (commented):
This looks good to me! Were you able to run a simulation and verify that the diagnostics were generated as expected?

ext/ClimaCouplerCMIPExt.jl
simone-silvestri: This order is not a problem because julia resolves method at runtime. We can move them up if you prefer.

ext/ClimaCouplerCMIPExt/seaice_diagnostics.jl
simone-silvestri: Ok I ve removed them from here. Let's deal with it separately

ext/ClimaCouplerCMIPExt/ocean_diagnostics.jl
simone-silvestri: restructured to allow changing the interval time and the mode ("averaged" vs "instantaneous") in the config file

ext/ClimaCouplerCMIPExt/seaice_diagnostics.jl
simone-silvestri: removed the try catch (I had it in the other code because I was running also without sea ice)

src/Input.jl
```diff
@@ -262,11 +262,27 @@ function argparse_settings()
         help = "Iteration interval for printing progress information [`nothing`, `<:Integer`] (default: nothing)"
         arg_type = Integer
         default = nothing
+        "--ocean_diagnostic_interval"
```
juliasloan25, line 265: it's kind of annoying but when we add a new config option we also have to update input_test.jl, like we did [here](https://github.com/CliMA/ClimaCoupler.jl/pull/1900/changes#diff-902d08844ef106fd207bf7921fa6fcd6a72eb743f2821d66c2712e08c4ce66d0R73). We should change those tests to not require that - I'll open an issue

juliasloan25 (approved):
LGTM after CI passes

akshaysridhar (approved):

PR #1903, opened 2026-05-05
fix start_date and stop_date of modis LAI for restarting simulations

szy21:
## Purpose 

Currently we don't take into account t_start when creating LAI with CL.prescribed_lai_modis. This results in an error when restarting the simulation (stop_date is before the actual start time of the simulation), e.g. [here](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1102#019df97f-c2ca-4380-b175-65ddb07c1e59/L273). With this fix the restart works: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1103#019dfa0e-59b8-4ba8-a4eb-b10240e8f3ea. cc @kmdeck 

I'm not sure if this is the best solution. @ph-kev Could you take over this PR and see if there is a better fix, and maybe check if start_date and stop_date need to be updated in other places? Thanks!

## Content

----

### DIFF
--- a/ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
+++ b/ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
@@ -113,11 +113,15 @@ function ClimaLandSimulation(
     #  since that's where we compute fluxes for this land model
     atmos_h = Interfacer.remap(surface_space, atmos_h)
 
+    # `tspan` can contain non-zero values (e.g. (720.0, 900.0)) when restarting
+    sim_start_date = start_date + Dates.Second(float(tspan[1]))
+    sim_stop_date = start_date + Dates.Second(float(tspan[2]))
+
     # Set up atmosphere and radiation forcing
     forcing = (;
         atmos = CL.CoupledAtmosphere{FT, typeof(atmos_h)}(atmos_h, gustiness),
         radiation = CL.CoupledRadiativeFluxes{FT}(
-            start_date;
+            sim_start_date;
             latitude = CC.Fields.coordinate_field(domain.space.surface).lat,
             longitude = CC.Fields.coordinate_field(domain.space.surface).long,
             toml_dict,
@@ -125,13 +129,12 @@ function ClimaLandSimulation(
     )
 
     # Set up leaf area index (LAI)
-    stop_date = start_date + Dates.Second(float(tspan[2] - tspan[1]))
     if lai_source == "modis_monthly"
         # Full monthly MODIS LAI data
         LAI = CL.prescribed_lai_modis(
             surface_space,
-            start_date,
-            stop_date;
+            sim_start_date,
+            sim_stop_date;
             time_interpolation_method = LinearInterpolation(),
         )
     elseif lai_source == "modis_monthly_climatology"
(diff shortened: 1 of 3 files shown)


ph-kev:
I can take a look at this. At a preliminary glance, I am not sure where the bug is, since the error message looks right to me.

```julia
julia> Dates.DateTime("2010-01-01") + Dates.Second(31622400)
2011-01-02T00:00:00
```

The simulation is restarting at `2011-01-02T00:00:00` and the error message is evaluating at this date. This might be a bug in ClimaLand?

szy21:
> I can take a look at this. At a preliminary glance, I am not sure where the bug is, since the error message looks right to me.
> 
> ```julia
> julia> Dates.DateTime("2010-01-01") + Dates.Second(31622400)
> 2011-01-02T00:00:00
> ```
> 
> The simulation is restarting at `2011-01-02T00:00:00` and the error message is evaluating at this date. This might be a bug in ClimaLand?

The simulation is restarting at the right time. But the LAI TimeVaryingInput doesn't have that time, which is a bug in the input to `CL.prescribed_lai_modis` I think.

ph-kev:
Yes, I agree with that. For the restart test, when I looked at `tspan`, I get

```julia
infil> tspan
(720.0 seconds (2010-01-01T00:12:00) [counter = 720, period = 1 second, epoch = 2010-01-01T00:00:00], 900.0 seconds (2010-01-01T00:15:00) [counter = 900, period = 1 second, epoch = 2010-01-01T00:00:00])

infil> dt
180.0 seconds (2010-01-01T00:03:00) [counter = 180, period = 1 second, epoch = 2010-01-01T00:00:00]
```

I think `stop_date = start_date + Dates.Second(float(tspan[2] - tspan[1]))` is just wrong.

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
```diff
@@ -113,25 +113,28 @@ function ClimaLandSimulation(
     #  since that's where we compute fluxes for this land model
     atmos_h = Interfacer.remap(surface_space, atmos_h)
 
+    # `tspan` can contain non-zero values (e.g. (720.0, 900.0)) when restarting
+    sim_start_date = start_date + Dates.Second(float(tspan[1]))
```
juliasloan25, line 117: This naming is a little confusing to me but I'm not sure what would be better. We used to have something like `ref_date` was what `start_date` is here, and `start_date` was what `sim_start_date` is now. It could be more clear to go back to that, but it also may require more changes so I'm ok with this

juliasloan25 (approved):

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
```diff
@@ -272,11 +275,11 @@ function ClimaLandSimulation(
     # Convert start_date and stop_date to ITime if using ITime
     if dt isa ITime
         start_date = tspan[1]
-        stop_date = tspan[2]
+        sim_stop_date = tspan[2]
```
szy21, line 278: what's the difference between `sim_stop_date` and `stop_date`? It seems they are the same?

ph-kev: Before, it was `stop_date = start_date + Dates.Second(float(tspan[2] - tspan[1]))`. In this PR, we add `sim_start_date = start_date + Dates.Second(float(tspan[1]))` and `sim_stop_date = start_date + Dates.Second(float(tspan[2]))`, so I needed to use `sim_stop_date` everywhere else in this function. The variable name could have been `stop_date`, but I thought that was confusing.

szy21: Ah ok, thanks!

PR #1908, opened 2026-05-07
Add reset_fluxes! function to FluxCalculator

akshaysridhar:
Generalizes the flux reset in https://github.com/CliMA/ClimaCoupler.jl/pull/1841. FluxCalculator now provides a `reset_fluxes` hook which can be modified per model component.

### DIFF
--- a/ext/ClimaCouplerCMIPExt/oceananigans.jl
+++ b/ext/ClimaCouplerCMIPExt/oceananigans.jl
@@ -576,6 +576,14 @@ function Interfacer.update_field!(sim::OceananigansSimulation, ::Val{:area_fract
     return nothing
 end
 
+function FluxCalculator.reset_fluxes!(sim::OceananigansSimulation)
+    oc_flux_T = surface_flux(sim.ocean.model.tracers.T)
+    OC.interior(oc_flux_T, :, :, 1) .= 0
+    oc_flux_S = surface_flux(sim.ocean.model.tracers.S)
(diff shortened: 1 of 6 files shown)


ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -521,6 +521,14 @@ function Interfacer.update_field!(sim::OceananigansSimulation, ::Val{:area_fract
     return nothing
 end
 
+function FluxCalculator.reset_fluxes!(sim::OceananigansSimulation)
```
juliasloan25, line 579: could you add a brief docstring please?

src/FieldExchanger.jl
```diff
@@ -270,6 +270,7 @@ Iterates `update_sim!` over all component model simulations saved in `cs.model_s
 """
 function update_model_sims!(model_sims, csf)
     for sim in model_sims
+        FluxCalculator.reset_fluxes!(sim)
```
juliasloan25: I'm not sure how I feel about resetting the fluxes here. To me it would make a little more sense to do it explicitly in the coupler loop, after `update_surface_fractions!` and before `exchange!`. That way it's very clear what order things are done in rather than being tucked inside `update_model_sims!`.

juliasloan25 (commented):
Are you planning to extend this for the other components as part of this PR or in a followup?

PR #1909, opened 2026-05-07
CMIP Checkpointing 

akshaysridhar:
Enables checkpoint and restart features for CMIP simulations. Defines simulation getter methods for restarts, adds multi-step restart tests.


ext/ClimaCouplerCMIPExt.jl
```diff
@@ -30,15 +30,14 @@ import ClimaCore as CC
 import ClimaParams as CP
 using KernelAbstractions: @kernel, @index, @inbounds
 
-# Include helper functions first (used by both oceananigans.jl and clima_seaice.jl)
-include("ClimaCouplerCMIPExt/climaocean_helpers.jl")
-
 # Include skin temperature utilities
 include("ClimaCouplerCMIPExt/skin_temperature.jl")
 
 # Include the model files
```
juliasloan25:
```suggestion
```

experiments/test/amip_test.yml
```diff
@@ -15,7 +15,9 @@ mode_name: "amip"
 netcdf_interpolation_num_points: [90, 45, 31]
 netcdf_output_at_levels: true
 output_default_diagnostics: true
+orographic_gravity_wave: ~
 rayleigh_sponge: true
+reproducible_restart: true
```
juliasloan25, line 20: does this option come from atmos? I don't think we set this in any configs outside of experiments/test, but maybe we need to if we want to be able to restart?

.buildkite/pipeline.yml
```diff
@@ -113,24 +113,34 @@ steps:
...
         command: "julia -O0 --color=yes --project=experiments/AMIP/ experiments/test/restart_state_only.jl"
+        retry: *retry_policy
         env:
           CLIMACOMMS_DEVICE: "CUDA"
-        retry: *retry_policy
         agents:
           slurm_gpus: 1
           slurm_mem: 32GB
 
+      - label: "GPU CMIP restarts (state only)"
```
juliasloan25:
```suggestion
      - label: "GPU CMIP restarts (state and cache)"
```
I think this is more appropriate since we're restarting the states and caches for models where it's possible to do so

juliasloan25 (approved):
This looks good to me! Just a few minor comments. Thank you!

experiments/test/amip_test.yml
akshaysridhar: Right, this comes from ClimaAtmos - applied within a cloud fraction calculation

PR #1910, opened 2026-05-07
unify microphysics params in diagnostic and prognostic edmf

szy21:
## Purpose 

Diagnostic EDMF and Prognostic EDMF are using different microphysics parameters now. I changed them to be the same. These are not the best parameters. I will change them again once I have a better set of parameters.

Also removes user specified `netcdf_interpolation_num_points` so all the output are on the same grid, which is easier for comparison. Also extends t_end in nightly 1M jobs.

## Content

----

### DIFF
--- a/config/nightly_configs/amip_coarse_diagedmf_1m_land.yml
+++ b/config/nightly_configs/amip_coarse_diagedmf_1m_land.yml
@@ -26,7 +26,7 @@ output_default_diagnostics: true
 rmse_check: true
 start_date: "20100101"
 surface_setup: "PrescribedSurface"
-t_end: "186days"
+t_end: "366days"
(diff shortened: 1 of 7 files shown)


config/nightly_configs/amip_coarse_diagedmf_1m_land.yml
```diff
@@ -26,7 +26,7 @@ output_default_diagnostics: true
 rmse_check: true
 start_date: "20100101"
 surface_setup: "PrescribedSurface"
-t_end: "186days"
+t_end: "366days"
```
trontrytel, line 29: Woo hoo!

trontrytel (approved):
Thank you!

PR #1902, opened 2026-05-05
Compute bflux, ustar, L_MO from combined fluxes

hgpeterson:
## Purpose 

Closes #1817 

## Content

[x] Remove `buoyancy_flux`, `ustar`, `L_MO` from coupler fields
[x] Compute `buoyancy_flux`, `ustar`, `L_MO` from combined surface fluxes
[x] Update the docs to reflect changes

PR #1914, opened 2026-05-08
set ustar to non zero

szy21:
## Purpose 

Fixes a bug in #1902 where ustar can be zero, which causes issues in EDMF boundary conditions.

## Content

----


szy21:
nightly amip: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1123

costachris (approved):

hgpeterson:
Good catch, thanks @szy21!

hgpeterson (approved):

PR #1916, opened 2026-05-08
Add `check_nan_every` to CI configs

### LINKED ISSUE #1915: Ensure CI fails when NaNs occur
In #1902 we discovered that our CI tests can pass even with most of the coupler fields being NaNs by the end of the simulation (see [here](https://buildkite.com/clima/climacoupler-ci/builds/8793#019e03b1-b414-45a9-93ca-b7fc72e30d4a)). It would be good if we could force the CI to fail for cases like this so that we can catch bugs faster.

hgpeterson:
Closes #1915

### DIFF
--- a/config/ci_configs/amip_albedo_temporal_map.yml
+++ b/config/ci_configs/amip_albedo_temporal_map.yml
@@ -15,3 +15,4 @@ time_varying_trace_gases: ["CO2"]
 vert_diff: "true"
 z_elem: 50
 z_stretch: false
+check_nan_every: 1
--- a/config/ci_configs/amip_coarse_ft64_hourly_checkpoints.yml
+++ b/config/ci_configs/amip_coarse_ft64_hourly_checkpoints.yml
@@ -7,3 +7,4 @@ microphysics_model: "0M"
 t_end: "1days"
 time_varying_trace_gases: ["CO2"]
 vert_diff: "true"
+check_nan_every: 1
--- a/config/ci_configs/amip_coarse_ft64_hourly_checkpoints_restart.yml
+++ b/config/ci_configs/amip_coarse_ft64_hourly_checkpoints_restart.yml
@@ -9,3 +9,4 @@ microphysics_model: "0M"
 t_end: "800secs"
 time_varying_trace_gases: ["CO2"]
 vert_diff: "true"
+check_nan_every: 1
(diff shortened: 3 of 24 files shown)


hgpeterson:
Confirmed that the setup that [erroneously passed](https://buildkite.com/clima/climacoupler-ci/builds/8793/canvas?sid=019e03b1-b347-4fe3-b170-99e5916cdc07&tab=output) last time [correctly failed](https://buildkite.com/clima/climacoupler-ci/builds/8813/canvas?sid=019e0891-a0e2-41fd-9b01-731d3fbef0bd&tab=output) under these changes.

szy21 (approved):
Note that check nan is very slow. It would be good to check if this significantly increases ci time. If so we can increase check_nan_every for the jobs that run longer.

hgpeterson:
The timings between the two cases linked above don't seem too different (although for some reason the sims that fail with NaNs take much longer?).

szy21:
> The timings between the two cases linked above don't seem too different (although for some reason the sims that fail with NaNs take much longer?).

cpu jobs can be a bit random as it depends on which node we get. But if it's not too different it's fine. We can always change later if ci becomes too long.

juliasloan25 (commented):
This looks good, thank you! Could you add some info to the "Debugging tips" section of the ClimaCoupler docs explaining that this option exists and what it does?

juliasloan25:
How does the run that you linked as failing now pass in this PR's CI? Is it because of the fix [here](https://github.com/CliMA/ClimaCoupler.jl/pull/1914)?

hgpeterson:
> How does the run that you linked as failing now pass in this PR's CI? Is it because of the fix [here](https://github.com/CliMA/ClimaCoupler.jl/pull/1914)?

Yes, that run is fixed due to #1914. But I tested that this PR's changes make CI correctly fail pre-#1914 [here](https://buildkite.com/clima/climacoupler-ci/builds/8813/canvas?sid=019e0891-a0e2-41fd-9b01-731d3fbef0bd&tab=output).

akshaysridhar (approved):

akshaysridhar:
Thanks @hgpeterson; following the above comment, if you could add a note to the "debugging" block that addresses this yaml config arg, we can squash and merge this PR. The CI timing seems to be at `1h24m` now vs `1h11m` prior to this callback frequency change.

PR #1840, opened 2026-03-27
use TripolarGrid and ConservativeRegridding for CMIP

juliasloan25:
## Purpose 
new version of #1724

includes the following fixes:
- don't apply SIC mask extra times for heat and salinity fluxes (merged into main, 26 March)
- zero out salinity flux at the start of a timestep

This PR includes the changes from https://github.com/CliMA/ClimaCoupler.jl/pull/1900 that allow us to use a higher-resolution coupler space. We can either merge it in that PR or here.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1918, opened 2026-05-12
use TripolarGrid and ConservativeRegridding for CMIP

juliasloan25:
## Purpose 
New version of TripolarGrid + ConservativeRegridding, to be merged this time.

Replaces #1840

## To-do
- [ ] add a config option for the ocean grid to use (LatitudeLongitude grid [default] vs TripolarGrid); use conservative regridding only with the TripolarGrid
- [ ] update construct_remapper to be generic/construct the appropriate remapper for each case
- [ ] add remap! methods for the conservative and non-conservative cases
- [ ] add construct_remapper methods for both cases

## Content

----


juliasloan25:
not to be merged, it turns out... split into #1919 and #1960

PR #1919, opened 2026-05-13
Use ConservativeRegridding with LatLonGrid

juliasloan25:
## Purpose 
Use ConservativeRegridding with the existing LatitudeLongitudeGrid. This will be a stepping stone on the way to using ConservativeRegridding with the TripolarGrid, so that we can break up the changes into smaller PRs and verify that this runs stably and reliably once we merge it into main

Originally authored by @akshaysridhar 

## To-do
- [x] resolve docs, AMIP envs to include new deps of ClimaCoupler ext
- [x] remove interpolation code
- [x] move code from climaocean_helpers_conservative_regrid.jl into climaocean_helpers.jl
- [x] remove commented out code
- [x] split out C_to_K --> T_melt change into a separate PR https://github.com/CliMA/ClimaCoupler.jl/pull/1920
- [x] use only one polar exclusion mask, check correctness

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1920, opened 2026-05-13
introduce T_melt for sea ice skin temp cap

juliasloan25:
## Purpose 
Introduces T_melt for sea ice skin temp cap, instead of using `C_to_K` for this. For now the default is used (same value as `C_to_K`) so this is not behavior changing.

Also updates some comments in the sea ice code.

Split off from #1919.
Originally authored by @akshaysridhar


juliasloan25:
Not needed for now so I'll close this PR, but we can always revisit it later on

Issue #1533, opened 2025-10-28
Add ECCO datasets to ClimaArtifacts


juliasloan25:
We download some data from ECCO to initialize sea ice, and we may want to do this for the ocean as well. Right now it's downloaded from ECCO using a username and password that we store as secrets, but it would be more straightforward (and generalizable to new machines) if we store these datasets in ClimaArtifacts.

Also EN4 temperature and salinity

PR #1925, opened 2026-05-15
Add sea-ice artifact for initialization

hgpeterson:
## Purpose 

Partially completes #1533

See the corresponding [PR in ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts/pull/160).

## To-do

- [x] Add `ecco4_SIarea_SIheff_2010_01` artifact to `Artifacts.toml`
- [x] Make sure the artifact is loaded in `ext/ClimaCouplerCMIPExt/clima_seaice.jl` when `start_date` is January 1, 2010, otherwise revert to downloading the data

In #1533 the EN4 T and S datasets were also mentioned, but I'll do that in a separate PR (or skip if no longer needed?).

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -74,3 +74,7 @@ git-tree-sha1 = "c1c0a16127bd547124cdf7288140002d74846efe"
 
 [wxquest_initial_conditions]
 git-tree-sha1 = "85b1e3654fb88f19a715ea6c235e1d66f254d2e6"
+
+[ecco4_SIarea_SIheff_2010_01]
+git-tree-sha1 = "889d7209b1a3b066e2dd5b0dbf7a674e32e4a589"
+
--- a/experiments/CMIP/Manifest-v1.11.toml
+++ b/experiments/CMIP/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.9"
 manifest_format = "2.0"
-project_hash = "ffb8414753aaed439d19e599e119e4e795b9030a"
+project_hash = "6cb1bf2558553283a328fb24831b4e206a81f60a"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "bbc22a9a08a0ef6460041086d8a7b27940ed4ffd"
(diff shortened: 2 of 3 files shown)


Artifacts.toml
```diff
@@ -74,3 +74,6 @@ git-tree-sha1 = "c1c0a16127bd547124cdf7288140002d74846efe"
 
 [wxquest_initial_conditions]
 git-tree-sha1 = "85b1e3654fb88f19a715ea6c235e1d66f254d2e6"
+
+[ecco4_SIarea_SIheff_2010_01]
+git-tree-sha1 = "889d7209b1a3b066e2dd5b0dbf7a674e32e4a589"
```
simone-silvestri:
```suggestion
git-tree-sha1 = "889d7209b1a3b066e2dd5b0dbf7a674e32e4a589"

```

simone-silvestri (approved):

akshaysridhar (approved):

experiments/CMIP/Manifest-v1.11.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.9"
+julia_version = "1.11.5"
```
akshaysridhar: PR LGTM - This Manifest seems to have been generated with 1.11.5 locally ? Let's maintain the `julia_version` here.

akshaysridhar:
```test = joinpath(@clima_artifact("ecco4_SIarea_SIheff_2010_01"), "SIheff_2010_01.nc")``` on `clima` points to the correct sampo file now.

experiments/CMIP/Manifest-v1.11.toml
hgpeterson: Ah yes, I was wondering about this. On `clima` when I load `climacommon/2025_05_15`, the julia version is 1.11.5. Is there a reason we were using 1.11.9?

juliasloan25 (approved):

experiments/CMIP/Manifest-v1.11.toml
juliasloan25: We try to stay up to date with the latest patch releases of each julia version so we can include any bugfixes or other improvements that may have been made. Realistically I don't think there's a big difference between 1.11.5 and 1.11.9 though.

Unfortunately climacommon doesn't have a julia 1.11 version above 1.11.5. We could make a new climacommon release with a newer julia 1.11 version if it makes your workflow difficult.

hgpeterson: Ok, no worries. It's not a big deal, just always gives a warning that the Manifest was resolved with a different version and then changes it any time I resolve. I'll revert this and merge!

PR #1926, opened 2026-05-15
update prescribed ocean/sea ice every 1 day

juliasloan25:
## Purpose 
Update SST and SIC at daily frequency for prescribed ocean/sea ice models.  The energy equation solved by prescribed sea ice is still computed at the model dt.

Revived version of #1219

Buildkite results look unchanged

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@ ClimaCoupler.jl Release Notes
 `main`
 -------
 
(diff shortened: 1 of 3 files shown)


szy21 (approved):
Thanks! Would restart still work with this? The test passed so I guess it's ok?

juliasloan25:
> Thanks! Would restart still work with this? The test passed so I guess it's ok?

The restarts should be okay, especially since the tests passed

PR #1935, opened 2026-05-19
don't apply polar mask for ocean/ice interactions

juliasloan25:
## Purpose 
Running CMIP on main with strict bounds checking enforced results in an out-of-bounds memory access error. This comes from applying the polar mask to the momentum fluxes between ocean and sea ice. Because ocean and sea ice exist on the same grid, we don't need to apply the polar mask to their interactions (only to atmos-surface interactions).

This PR removes the polar mask from the ocean/sea ice interactions, and adds `check-bounds=yes` to one CMIP run in CI, so that we can catch similar errors in the future.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -346,9 +346,10 @@ steps:
 
   - group: "CMIP"
     steps:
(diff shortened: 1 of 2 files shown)


hgpeterson (approved):
Nice catch!

PR #1939, opened 2026-05-20
expose diagnostics frequency and reduction for each component

### LINKED ISSUE #1650: Add an option to change diagnostics for each component model
Right now if we want to change the diagnostic saving period or reduction type, we need to manually change it for each component model within the code. This is done in a different place for each component: in the config file for atmos, in the constructors for Oceananigans, ClimaSeaIce, bucket, and integrated land.

It would be helpful if we just have config options that control this. For each component, the user should be able to specify diagnostics frequency and the reduction type (average, instantaneous, etc).

juliasloan25:
## Purpose 
With this change, we can output hourly instantaneous diagnostics of each component (and the coupler fields), which will hopefully facilitate debugging CMIP. When we use these higher frequency diagnostics, they should be visible while the simulation is still running and also afterwards even if the simulation crashes.

Closes #1650 

## Content
- [x] add config options for land and coupler diagnostics period and reduction
- [x] updated CMIP longrun configs to output daily instantaneous diagnostics
- [x] add hourly diagnostics to the two CMIP runs in regular CI so this feature gets tested (and decrease sim lengths)
- [x] added tests for new helper functions
- [x] bugfix: pass `output_dir` to `CL.default_diagnostics` so we call the model-specific method rather than the no-op default

----

### DIFF
--- a/config/longrun_configs/cmip_progedmf_land.yml
+++ b/config/longrun_configs/cmip_progedmf_land.yml
@@ -2,6 +2,8 @@ FLOAT_TYPE: "Float64"
 albedo_model: "CouplerAlbedo"
 atmos_config_file: "config/atmos_configs/climaatmos_progedmf.yml"
 checkpoint_dt: "1months"
+coupler_diagnostics_period: "1days"
+coupler_diagnostics_reduction: "instantaneous"
 coupler_toml: ["toml/amip_progedmf.toml"]
 dt_atmos: "120secs" # 2 minutes
 dt_cloud_fraction: "1hours"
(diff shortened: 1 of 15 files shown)


config/longrun_configs/cmip_progedmf_land.yml
```diff
@@ -13,18 +15,24 @@ dt_seaice: "360secs" # 6 minutes
...
+land_diagnostics_period: "1hours"
+land_diagnostics_reduction: "instantaneous"
 land_model: "integrated"
 land_spun_up_ic: false
 mode_name: "cmip"
 ocean_model: "oceananigans"
-output_default_diagnostics: true
+ocean_diagnostic_interval: "1hours"
+ocean_diagnostic_mode: "instantaneous"
+output_default_diagnostics: false
```
akshaysridhar, line 32: This ('output_default_diagnostics') only applies to `atmos` (the name might suggest that it switches of default diagnostics for all (or coupler field) diagnostics); maybe we can include a note in the `Input` docs.

config/longrun_configs/cmip_progedmf_land.yml
akshaysridhar, line 1: Thanks @juliasloan25 ; I'm guessing the scratch disk quota is higher than what's necessary for these longruns to proceed without interrupting other jobs ?

akshaysridhar: Scratch usage is at 23% right now; LGTM!

juliasloan25: Hm this is a good point, I forgot that the longruns on main don't crash immediately 😅 I'll change them to daily diagnostics and we can switch to hourly when we're actively debugging

config/longrun_configs/cmip_progedmf_land.yml
juliasloan25: I updated the docs!

config/longrun_configs/cmip_progedmf_land.yml
juliasloan25: I changed the diagnostics for the regular CI CMIPs to hourly, and the longruns to daily. I also updated this PR description to emphasize that it adds the ability to change the diagnostics frequency and reduction for each component, which is really the main benefit from it

akshaysridhar (approved):

Issue #1543, opened 2025-10-31
Atmos and surface models update radiation in different frequencies


szy21:
Currently, radiation is called every dt_rad (typically one hour) in the atmosphere, but is called every timestep in surface models (land or sea ice). This is inconsistent and energy is not conserved. To fix this, we can still calculate radiation at every timestep for land and pass the averaged quantities (albedo, T_sfc, emissivity) to the atmosphere. For sea ice we can do the same thing or move radiation to a callback.

PR #1945, opened 2026-05-22
enable flux accumulation

### LINKED ISSUE #1701: accumulate fluxes correctly when components have different dts
Right now when we have different dt for each component, we take the coupling dt to be the maximum. This will lead to conservation issues because e.g. the land is recomputing fluxes every 6 minutes, but the atmosphere sees the same atmos/land fluxes for an entire 30 minute coupling step.

Instead, we should take `dt_cpl = minimum(surface dts)`, and exchange fluxes at the end of each component model's step. We can continue to support the larger coupling dt just to keep the coupler as flexible as possible.

### Approach - timestep changes
- In `Interfacer.step!` for each surface, step the component and update the coupler field with new fluxes (in other words, simplify the logic done in [step_model_sims!](https://github.com/CliMA/ClimaCoupler.jl/blob/bfc631af49361274bbb80d20f708ce655a22450e/src/FieldExchanger.jl#L286-L296))
  -  Do we want to do this for all surfaces or just for the `ImplicitFluxSimulation` surfaces? --> Should be for every surface, since the atmosphere needs frequent flux updates
- Set `dt_cpl = min(surface_dts)` so we compute fluxes and exchange more often
  - Should we skip flux computation of surface models that haven't been updated yet? --> Zhaoyi: atmosphere will be unstable if we keep fluxes constant for 30 minutes, so we should recompute

## Edit March 16 2026: Flux accumulation
Since the atmosphere will be unstable if we keep fluxes constant for 30 minutes, we need to recompute fluxes between the atmosphere and models that haven't stepped yet as well (i.e. ocean and sea ice). This involves flux accumulation over the 30 minute "long surface timestep" so that the atmosphere and slow surface see the same fluxes integrated over this time period.

### Approach - flux accumulation
- For any ExplicitFluxSimulation with a component dt larger than the dt_cpl, we need to accumulate the flux between the atmosphere and this surface model over the time period of its timestep
  - Where right now we update the flux directly (`update_turbulent_fluxes!`), we should instead check if it is time for the surface model to step. If not, add the flux to our accumulator. If so, divide the accumulated flux by the number of coupling timesteps that have passed since this component's last timestep, and update the surface with that average.
  - To minimize regridding operations, it may be best to store the accumulated fluxes on the coupler space. I'm not sure if we can use the existing coupler fields for this, or if we should allocate new fields for accumulating fluxes. If we put it on the coupler field we use to pass the land's fluxes, we might end up with strange values along the continental boundaries.

Note that we'll also need to accumulate precipitation, but that isn't correctly accumulated over an atmosphere timestep yet, so we don't have to worry about it right now. One day we can apply the same logic there that we use for the fluxes here.

Note 2 that we'll also need to accumulate radiation, since this downwelling fluxes are computed via an infrequent callback, but the surface models recompute the upwelling radiation and T/emissivity more often. We may want to accumulate the upwelling fluxes over a radiation timestep, then use those to compute a T_sfc/emissivity that gets sent to the atmosphere.

Note 3: We need to think about how this will work with stepping component models concurrently. For example, if we want to accumulate fluxes from `t = 0` to `t = 30` and send the averaged fluxes to ocean/sea ice, ocean/sea ice can't run until after atmos/land has run for 30 minutes. This imposes an ordering that will make concurrency more complicated. We may need to have the models step offset from each other - so the concurrency would be between atmos/land from `t=30` to `t=60` and ocean/sea ice from time `t=0` to `t=30` etc.

juliasloan25:
## Purpose
When a surface model’s timestep is longer than `dt_cpl`, it previously received instantaneous turbulent fluxes every coupling step, which breaks conservation over the surface’s own timestep. This PR time-averages turbulent fluxes over each "slow" surface’s dt window before pushing them to that surface’s boundary conditions, while the atmosphere continues to see freshly computed fluxes every `dt_cpl`.

For example, in a simulation where we have `dt_cpl = 6mins` and `dt_ocean = dt_seaice = 30mins`, previously the ocean and sea ice models would receive instantaneous fluxes every 30 minutes. Now, the fluxes will be computed and accumulated every 6 minutes, then averaged at the 30 minute intervals and provided to the ocean/sea ice.

CMIP fluxes look very similar to main before this PR, with general structures the same and minor changes to extrema

Closes #1701

#### integrated land fluxes
Since these fluxes are computed implicitly in the land model, with this implementation they would not be recomputed at the coupler timestep when we have `dt_land > dt_cpl`. This would result in the atmosphere using out-of-date fluxes at some timesteps. E.g. if we have `dt_cpl = 6mins` and `dt_land = 12mins`, the atmos/land fluxes would be recomputed every 12 mins, so on odd timesteps (6, 18, 24 mins, etc) the atmos would be using stale fluxes from the previous coupler dt. In comparison, with the ocean/sea ice/bucket accumulation the fluxes are recomputed with the updated atmosphere state every `dt_cpl`. For now we can avoid this issue by setting `dt_land == dt_cpl`, but it would be good to remove this restriction in the future.

#### precipitation accumulation
This is not currently accumulated across stages in the atmosphere, so accumulating it as we do for the turbulent fluxes here would be incorrect anyway. Once the atmosphere does accumulate correctly over its own step, we can easily set up precipitation accumulation in the same way we do it for turbulent fluxes in this PR. See https://github.com/CliMA/ClimaCoupler.jl/issues/1972

#### radiation accumulation
Radiation is computed by an infrequent callback (controlled by `dt_rad`), but surface models recompute upwelling longwave radiation as a function of their updated T_sfc at each timestep. This results in an inconsistency where the radiation provided by the atmosphere changes slowly, but surface LW_u changes more quickly. This is a bit different from the flux accumulation being addressed in this PR, so I'll leave it as a piece of future work: see https://github.com/CliMA/ClimaCoupler.jl/issues/1543

## Content
- [x] add FluxAccumulator object to hold accumulated fluxes for each slow explicit surface
- [x] add flux_accumulators to CoupledSimulation 
- [x] add `sim_dt(sim)` and `will_step(sim, t)` helpers
- [x] checkpoint / restart support for `flux_accumulators`
- [x] add extra scratch space/buffers to BucketSimulation to enable flux accumulation even though its fluxes are computed on the land space (not the boundary space)
- [x] update docs, add unit tests
- [x] add integration tests in experiments/test/fluxes_test.jl

Note: CMIP configs withdt_cpl < dt_ocean/dt_seaice(e.g.cmip_oceananigans_climaseaice*.yml) will have behavior changes


ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -371,7 +371,14 @@ function FluxCalculator.compute_surface_fluxes!(
     sim::BucketSimulation,
     atmos_sim::Interfacer.AbstractAtmosSimulation,
     thermo_params,
+    accumulator = nothing,
 )
+    # Bucket land's `update_flux_fields!` does not push fluxes back to the
```
juliasloan25: fix this (needs to use flux accumulator pathway)

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -295,26 +328,10 @@ function Interfacer.update_field!(sim::BucketSimulation, ::Val{:air_velocity}, u
             sim.integrator.p.bucket.scratch2,
         )
 end
-function Interfacer.update_field!(
```
juliasloan25, line 298: these two `update_field!` methods have been unused for a while

hgpeterson:
Awesome! I've just taken a quick look so far; might not get to do a thorough review until next week...

docs/src/fluxcalculator.md
```diff
@@ -43,6 +43,32 @@ Note also that [`FluxCalculator.turbulent_fluxes!`](@ref) only
 computes turbulent fluxes, not radiative fluxes. Currently, these are computed
 within the atmospheric model.
 
+## Turbulent flux accumulation for slow surfaces
+
+When a surface model's own timestep is larger than the coupling timestep
+(e.g. `dt_ocean > Δt_cpl`), pushing a fresh flux to it every coupling step is wasteful:
+the surface only consumes its boundary fluxes when it steps, andthe intermediate writes
```
hgpeterson:
```suggestion
the surface only consumes its boundary fluxes when it steps, and the intermediate writes
```

docs/src/fluxcalculator.md
```diff
@@ -43,6 +43,32 @@ Note also that [`FluxCalculator.turbulent_fluxes!`](@ref) only
...
+are typically overwritten by the next coupling-stepexchange. For correctness and
```
hgpeterson:
```suggestion
are typically overwritten by the next coupling step exchange. For correctness and
```

docs/src/fluxcalculator.md
```diff
@@ -43,6 +43,32 @@ Note also that [`FluxCalculator.turbulent_fluxes!`](@ref) only
...
+fluxes rather than a single instantaneous snapshot.
+
+To support this, the coupler allocates a [`FluxCalculator.FluxAccumulator`](@ref)
+for each slow explicit surface (see [`Interfacer.sim_dt`](@ref),
+[`Interfacer.will_step`](@ref)), and:
+
+- Each coupling step, [`FluxCalculator.turbulent_fluxes!`](@ref) calls
+  [`FluxCalculator.accumulate!`](@ref) on the accumulator instead of writing the
+  flux directly to the surface via `update_turbulent_fluxes!`. The area-weighted
+  combined `csf.F_*` fields that the atmosphere reads are still updated every call.
```
hgpeterson:
```suggestion
  combined `CoupledSimulation.Fields.F_*` fields that the atmosphere reads are still updated every call.
```

docs/src/checkpointer.md
```diff
@@ -108,6 +108,20 @@ forward, but there are still several challenges that need to be solved:
...
+coupler maintains a `cs.flux_accumulators` `NamedTuple` of partial turbulent
+flux sums (see [Turbulent flux accumulation for slow surfaces](@ref)). These
+partial sums are part of the simulation state — restarting in the middle of a
+slow-surface window without them would feed the wrong time-average to the
+surface at the next push. `Checkpointer.checkpoint_sims` writes the
+accumulators to a per-rank JLD2 file whenever they are non-empty, and
+[`Checkpointer.restart!`](@ref) calls
+[`Checkpointer.restart_flux_accumulators!`](@ref) on the matching file during
+a restart. Configurations without slow surfaces leave `flux_accumulators`
+empty and no accumulator file is written.
```
hgpeterson:
```suggestion
The `CoupledSimulation` object stores `flux_accumulators` to handle explicit surfaces 
with timesteps larger than `Δt_cpl`. This field is a `NamedTuple` of partial turbulent
flux sums (see [Turbulent flux accumulation for slow surfaces](@ref)). These
partial sums are part of the simulation state, and are required for properly restarting 
in the middle of a slow-surface window. `Checkpointer.checkpoint_sims` writes the
accumulators to a per-rank JLD2 file whenever they are non-empty, and
[`Checkpointer.restart!`](@ref) calls
[`Checkpointer.restart_flux_accumulators!`](@ref) on the matching file during
a restart. Configurations without slow surfaces leave `flux_accumulators`
empty and no accumulator file is written.
```

docs/src/coupledsimulation.md
```diff
@@ -119,6 +120,33 @@ configure which variables are saved and at what frequency.
...
+
+When an accumulator is present, the coupler:
+
+- adds the per-surface turbulent flux to the accumulator each coupling step
+  (instead of calling `update_turbulent_fluxes!` on that surface), and
+- if the surface is about to step, divides the accumulator by
+  `n_steps`, writes the time-averaged flux to the surface boundary conditions
+  via `update_turbulent_fluxes!`, and resets the accumulator.
+
+The area-weighted combined `csf.F_*` fields (which the atmosphere reads) are
```
hgpeterson, line 146:
```suggestion
The area-weighted combined `cs.fields.F_*` fields (which the atmosphere reads) are
```

ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
```diff
@@ -413,67 +423,70 @@ function FluxCalculator.compute_surface_fluxes!(
...
+    # area-weighted contribution to `csf.F_*`, and (slow case) also add the masked value
+    # to the accumulator.
     area_fraction = Interfacer.get_field(sim, Val(:area_fraction))
-    FT = CC.Spaces.undertype(axes(csf))
-
-    # Combine turbulent energy fluxes from the bucket model
-    # Use temporary variables to avoid allocating
-    Interfacer.remap!(csf.scalar_temp1, fluxes.lhf)
-    Interfacer.remap!(csf.scalar_temp2, fluxes.shf)
+    ρ_liq = LP.ρ_cloud_liq(sim.model.parameters.earth_param_set)
```
hgpeterson: Move to [line 445](https://github.com/CliMA/ClimaCoupler.jl/pull/1945/changes#diff-b34e13cbe11899a640b06592e2a1aacdbf14cc9d4a6731c481d24db7f357871eR445)?

hgpeterson (approved):
I think this looks good, and it's an important change; thank you! I still find the special treatment for land confusing... but, for now, making sure that `dt_land = dt_cpl` is fine with me. 

If we don't have issues tracking "what this PR does not do," we should make some.

docs/src/checkpointer.md
juliasloan25: Thank you! This is much clearer

akshaysridhar (approved):

PR #1951, opened 2026-05-27
Change CA to a weakdep

imreddyTeja:
## Purpose 
Allows the coupler to be used with Julia versions before 1.10.7

## Content
Made CA a weakdep.
I'm not too sure about what the behavior of the albedo functions should be if CA is not loaded

----


Project.toml
```diff
@@ -65,7 +65,7 @@ ClimaTimeSteppers = "0.8.11, 0.9"
 ClimaUtilities = "0.1.22"
 Dates = "1"
 GeoMakie = "0.7"
-Insolation = "1.1"
+Insolation = "1.1.1"
```
imreddyTeja: this change is not needed

src/Input.jl
```diff
@@ -371,8 +371,12 @@ function get_coupler_config_dict(config_file)
...
     coupler_config_dict = YAML.load_file(config_file)
 
-    # Get ClimaAtmos default configuration dictionary
-    atmos_default = CA.default_config_dict()
+    # Get ClimaAtmos default configuration dictionary. This method only exists when the
+    # ClimaCouplerClimaAtmosExt is loaded, so we check if it ClimaAtmos is
+    # before calling it. If it doesn't exist, we return an empty dictionary.
+    atmos_default =
+        any(k -> Symbol(k.name) == :ClimaAtmos, keys(Base.loaded_modules)) ?
+        atmos_default_config_dict() : Dict()
```
juliasloan25: Could we just return `Dict()` from the default method defined below, and avoid the loaded module check here?

src/Models/prescr_ocean.jl
```diff
@@ -269,43 +266,4 @@ the atmospheric wind. The albedos are calculated using the `surface_albedo_direc
 and `surface_albedo_diffuse` functions from the `ClimaAtmos` module, so this
 is dependent on running with `ClimaAtmosSimulation` as the atmosphere simulation.
 """
-function set_albedos!(sim::PrescribedOceanSimulation, t)
```
juliasloan25, line 272: the docstring can be moved to the atmos extension too

src/Models/prescr_ocean.jl
```diff
@@ -269,43 +266,4 @@ the atmospheric wind. The albedos are calculated using the `surface_albedo_direc
...
-    )
-    Interfacer.update_field!(
-        sim,
-        Val(:surface_diffuse_albedo),
-        CA.surface_albedo_diffuse(α_model).(λ, cos_zenith, wind_atmos),
-    )
-
-    return nothing
-end
+function set_albedos! end
```
juliasloan25: for this method can you throw an error and explain this model needs to be used with ClimaAtmos, but the extension isn't loaded?

juliasloan25 (commented):
Thank you Teja! This looks good, I just left some small comments

src/Input.jl
juliasloan25:
```suggestion
    @assert any(k -> Symbol(k.name) == :ClimaAtmos, keys(Base.loaded_modules)) "ClimaAtmos.jl is required to be loaded to run coupled simulations. Please make sure the extension is correctly loaded."
    atmos_default = atmos_default_config_dict()
```

and add `atmos_default_config_dict` to required atmos functions in interfacer docs

src/Input.jl
```diff
@@ -371,8 +371,11 @@ function get_coupler_config_dict(config_file)
     # Load the coupler config file into a dictionary
     coupler_config_dict = YAML.load_file(config_file)
 
-    # Get ClimaAtmos default configuration dictionary
-    atmos_default = CA.default_config_dict()
+    # Get ClimaAtmos default configuration dictionary. This method is only defined when the
+    # ClimaCouplerClimaAtmosExt is loaded or a user defines it
+    # if it is not defined, give an informative error message. `hasmethod` should be compile time constant
+    @assert hasmethod(atmos_default_config_dict, ()) "ClimaAtmos.jl is required to be loaded to use the default `atmos_default_config_dict()` method. Please make sure the extension is correctly loaded, or define `atmos_default_config_dict()`."
```
imreddyTeja: @juliasloan25 this is a bit different than what you suggested

src/Models/prescr_ocean.jl
```diff
@@ -261,51 +258,8 @@ function FieldExchanger.import_atmos_fields!(csf, sim::PrescribedOceanSimulation
...
-        sim,
-        Val(:surface_diffuse_albedo),
-        CA.surface_albedo_diffuse(α_model).(λ, cos_zenith, wind_atmos),
-    )
-
-    return nothing
+function set_albedos!(model::Interfacer.AbstractSurfaceStub, _)
+    model isa PrescribedOceanSimulation &&
+        @error "The default `set_albedos!` for PrescribedOceanSimulation requires ClimaAtmos.jl to be loaded. Please make sure the extension is correctly loaded, or define `set_albedos!`"
+    @error "set_albedos! is not defined for $(typeof(model))"
```
juliasloan25: This is fine, or we can define one method for PrescribedOceanSimulation and another for AbstractSurfaceStub (which should probably go in src/surface_stub.jl)

src/Input.jl
juliasloan25: This looks good! Thank you

juliasloan25 (approved):

PR #1956, opened 2026-05-29
explicitly check atmos albedo after init

juliasloan25:
## Purpose 
Testing to make sure https://github.com/CliMA/ClimaAtmos.jl/pull/4535 is ok to merge

I don't think we want to keep this change in the drivers, but maybe in e.g. the fluxes test.


akshaysridhar:
Closing since ClimaAtmos#4535 has been merged and this PR has served its purpose as a downstream check.

PR #1962, opened 2026-05-31
use prognostic edmf in benchmark

szy21:
## Content

----


szy21:
manually merging this so the benchmark tonight can run with prognostic edmf.

PR #1963, opened 2026-06-01
tag v0.2.0

juliasloan25:
## Purpose 
ClimaCoupler's last release was in December of 2024, which might make it look like the package is out of date and/or not maintained, and we've made a lot of interface changes since then (moving components to extensions, streamlining the drivers, etc). This PR tags a minor release to reflect these changes.

We can also think about switching to a more regular release schedule, now that we have some external users.


akshaysridhar (approved):

akshaysridhar:
We can follow up with another release once the latest changes from ClimaOcean, and the TripolarGrid + ConservativeRegridding features have been merged. Thanks @juliasloan25!

PR #1965, opened 2026-06-02
re-enable central longruns [skip ci]

juliasloan25:
## Purpose 
The "Cess climate sensitivity" longrun failed this weekend, which I believe is because the comparison data is downloaded during init (which was done on clima), but the run tries to access it when running on central.

We disabled central longruns a few months ago when we were having module issues on that cluster. The issues have been resolved so I think it's fine to re-enable the central init and two aquaplanets. The downside is that we have to init 4 environments (AMIP + CMIP on central + clima) instead of 2 (AMIP + CMIP on clima only), but I think it's worth it to be able to run on central and to test longer slabplanets.

## Content
- [x] init AMIP and CMIP environments on central
- [x] add back aquaplanet and slabplanet longruns

After review, I'll merge this without running CI because it only changes longruns.

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -15,26 +15,26 @@ env:
 timeout_in_minutes: 1800 # 30 hours
 
 steps:
-  # - label: "init :computer:"
-  #   key: "init_cpu_env"
-  #   command:
-  #     - "echo $$JULIA_DEPOT_PATH"
+  - label: "init :computer:"
+    key: "init_cpu_env"
+    command:
+      - "echo $$JULIA_DEPOT_PATH"
 
-      # - echo "--- Instantiate AMIP env"
-      # - "julia --project=experiments/AMIP/ -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
-      # - "julia --project=experiments/AMIP/ -e 'using Pkg; Pkg.precompile()'"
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Thanks! Should we just run the cess sensitivity job on clima?

juliasloan25:
> Thanks! Should we just run the cess sensitivity job on clima?

Yeah that would be the other option. I thought it would be good to continue testing a couple slabplanet cases in the longruns, but we can also just remove them and run the climate sensitivity on clima. I didn't want to add them back and put them on clima because that would be a lot of long runs on clima.

szy21:
> > Thanks! Should we just run the cess sensitivity job on clima?
> 
> Yeah that would be the other option. I thought it would be good to continue testing a couple slabplanet cases in the longruns, but we can also just remove them and run the climate sensitivity on clima. I didn't want to add them back and put them on clima because that would be a lot of long runs on clima.

Yes, I meant running the climate sensitivity on clima. The slabplanet can stay on central. The climate sensitivity test should only be dependent on the two amip jobs running clima?

juliasloan25:
> Yes, I meant running the climate sensitivity on clima. The slabplanet can stay on central. The climate sensitivity test should only be dependent on the two amip jobs running clima?

Ohh right it needs to be on clima to access the artifacts. I changed it to run on clima

PR #1971, opened 2026-06-04
Update to ClimaOcean 0.10.0

### LINKED ISSUE #1969: Use improved SE conservative regridding
As of #1919, we now use ConservativeRegridding.jl for CMIP runs. This is a great improvement, however the regridding algorithm being used effectively treats the spectral element space as finite volume. This reduces the resolution of regridding and causes a lot of grid imprinting.

An improved spectral element regridding algorithm is being implemented in https://github.com/JuliaGeo/ConservativeRegridding.jl/pull/99. After this is merged and released, we should use this for CMIP runs.

simone-silvestri:
Update ClimaCoupler to use ClimaOcean 0.10.0. 
1) Adds the TripolarGrid ocean configuration. 
2) Removes Lat-long grid as part of standard CMIP configurations, removes polar masking functions
3) Uses ConservativeRegridding0.2.5 which introduces consistent spectral element remapping. 
4) Updates CMIP runs to use `float64` clock. Future PRs will re-introduce the ITime clock following changes across ocean repositories.


juliasloan25:
It looks like with the new interface, we can't pass the model clock for the ocean or sea ice models to use anymore. In main it gets passed to `ClimaOcean.ocean_simulation` [here](https://github.com/CliMA/ClimaCoupler.jl/blob/8df84bad498b1e894791e8d1d6322baf5d66e9f5/ext/ClimaCouplerCMIPExt/oceananigans.jl#L178) and to `ClimaOcean.SeaIces.sea_ice_simulation` [here](https://github.com/CliMA/ClimaCoupler.jl/blob/8df84bad498b1e894791e8d1d6322baf5d66e9f5/ext/ClimaCouplerCMIPExt/clima_seaice.jl#L91).

We now call:
- `ClimaOcean.OceanConfigurations.latitude_longitude_ocean` -->
- `NumericalEarth.Oceans.ocean_simulation` -->
- `NumericalEarth.Oceans.hydrostatic_ocean_simulation` -->
- `Oceananigans.Models.HydrostaticFreeSurfaceModel`
 
The only function in this callstack that exposes the `clock` kwarg is `Oceananigans.Models.HydrostaticFreeSurfaceModel`. The default clock uses Float64, so this PR means that we can't use ITime/DateTime times in CMIP anymore. I think this is still worth merging so we can use it to push the TripolarGrid change through, but we should re-enable non-Float time types.

I didn't trace through the callstack for the sea ice model but it has the same issue.

The fix would be to open PRs in NumericalEarth and ClimaOcean to expose the clock arg for the functions listed above, as well as the corresponding functions for the sea ice model. For now I switched all the CMIP models to use float time (`use_itime = false`).

juliasloan25:
I got this to a point where the ocean simulation can be constructed successfully. Now it fails when trying to access the sea ice artifact: `ERROR: LoadError: IOError: mktempdir("/resnick/groups/esm/ClimaArtifacts/artifacts/ecco4_SIarea_SIheff_2010_01"): permission denied (EACCES)` (full stacktrace [here](https://buildkite.com/clima/climacoupler-ci/builds/9082#019e94f7-41b6-4457-9894-c8f5d9a9524b)).

The fix that was applied in ClimaOcean [here](https://github.com/CliMA/ClimaOcean.jl/pull/795) needs to be moved over into NumericalEarth.jl. After that's resolved, we may need to make more changes in the sea ice setup here, or in ClimaOcean/NumericalEarth/ClimaSeaIce, but we'll have to re-run CMIP to see.

hgpeterson:
> I got this to a point where the ocean simulation can be constructed successfully. Now it fails when trying to access the sea ice artifact: `ERROR: LoadError: IOError: mktempdir("/resnick/groups/esm/ClimaArtifacts/artifacts/ecco4_SIarea_SIheff_2010_01"): permission denied (EACCES)` (full stacktrace [here](https://buildkite.com/clima/climacoupler-ci/builds/9082#019e94f7-41b6-4457-9894-c8f5d9a9524b)).
> 
> The fix that was applied in ClimaOcean [here](https://github.com/CliMA/ClimaOcean.jl/pull/795) needs to be moved over into NumericalEarth.jl. After that's resolved, we may need to make more changes in the sea ice setup here, or in ClimaOcean/NumericalEarth/ClimaSeaIce, but we'll have to re-run CMIP to see.

The corresponding PR in NumericalEarth for the sea-ice dataset is [here](https://github.com/NumericalEarth/NumericalEarth.jl/pull/331)

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -92,74 +92,16 @@ function OceananigansSimulation(
     # Retrieve EN4 data (monthly)
     # (It requires username and password)
     dates = range(start_date, step = Dates.Month(1), stop = stop_date)
-    en4_temperature = CO.Metadata(:temperature; dates, dataset = CO.EN4.EN4Monthly())
-    en4_salinity = CO.Metadata(:salinity; dates, dataset = CO.EN4.EN4Monthly())
+    en4_temperature = CO.Metadata(:temperature; dates, dataset = CO.EN4Monthly())
+    en4_salinity = CO.Metadata(:salinity; dates, dataset = CO.EN4Monthly())
     download_dataset(en4_temperature)
     download_dataset(en4_salinity)
```
akshaysridhar:
```suggestion
CO.download_with_fallback(en4_temperature)
CO.download_with_fallback(en4_salinity)
```

akshaysridhar: After which the `model_clock` kwarg is the next failure point in `hydrostatic_ocean_simulation`

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -92,74 +92,16 @@ function OceananigansSimulation(
...
 
-    # Set up ocean grid (1 degree)
-    resolution_points = (360, 160, 32)
-    Nz = last(resolution_points)
-    depth = 4000 # meters
```
akshaysridhar, line 103:
```
The vertical range of the NumericalEarth.DataWrangling.EN4.EN4Monthly() dataset (5500.0 m) is smaller than the target grid (6000.0 m). Some vertical levels cannot be filled with data.
``` 
(when using new default from `CO.OceanConfigurations` )

simone-silvestri: We should initialize with WOAAnnual

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -172,18 +114,10 @@ function OceananigansSimulation(
...
-        timestepper = :SplitRungeKutta3,
-        momentum_advection,
-        tracer_advection,
-        free_surface,
-        closure,
-    )
+
+    ocean = CO.OceanConfigurations.latitude_longitude_ocean(arch; model_clock, depth=5500)
+    ocean.stop_time = stop_time
+    ocean.Δt = Δt
```
akshaysridhar:
```suggestion
    ocean.Δt = float(dt)
```

akshaysridhar:
Needs https://github.com/CliMA/ClimaAtmos.jl/pull/4574

.buildkite/longruns/pipeline.yml
akshaysridhar, line 1: Keep only CMIP runs (test only - revert before merge).

config/longrun_configs/cmip_edonly_bucket.yml
akshaysridhar, line 1: Shortened for <12h turnaround : revert.

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -172,18 +114,10 @@ function OceananigansSimulation(
...
-        stop_time,
-        Δt = float(model_Δt),
-        timestepper = :SplitRungeKutta3,
-        momentum_advection,
-        tracer_advection,
-        free_surface,
-        closure,
-    )
+
+    ocean = CO.OceanConfigurations.orca_grid(arch; clock = model_clock, depth=5500)
```
hgpeterson: Maybe best to put the orca grid in a separate PR?

akshaysridhar: We run into parameter memory issues with the orca default set up in any case; reverted this change; we can revisit orca in a future PR.

daverumph (approved):
This looks good to me.

Is the removed support for passed in the model clock, noted by Julia S., being addressed?

hgpeterson (approved):
Approved once we sort out the parameter memory issues!

akshaysridhar:
> This looks good to me.
> 
> Is the removed support for passed in the model clock, noted by Julia S., being addressed?

PRs enabling clock kwargs in the ocean model components, and then ITime in ClimaCoupler will follow this one.

PR #1975, opened 2026-06-08
Focus buildkite pipelines on prognostic EDMF runs. (CMIP)

akshaysridhar:
Remove diagnostic EDMF configs, and corresponding entries from CMIP pipeline.


szy21:
Oh, let's keep amip with diagnostic EDMF for now. I'll remove it when we are more confident about prognostic EDMF. I was just thinking for cmip we don't need to use diagnostic EDMF anymore.

config/atmos_configs/climaatmos_progedmf_1m.yml
```diff
@@ -26,11 +26,11 @@ rayleigh_sponge: true
 surface_setup: "DefaultMoninObukhov"
 t_end: "120days"
 time_varying_trace_gases: ["CO2", "O3"]
-toml: [toml/longrun_aquaplanet_diagedmf.toml]
+toml: [toml/longrun_aquaplanet_progedmf.toml]
```
akshaysridhar, line 29: @szy21 Do we still want the diagedmf toml here ? (And above in the 0M config)

szy21: These toml are not doing anything in the coupler I think, as we always have a coupler_toml. So this change is fine.

szy21 (approved):

config/atmos_configs/climaatmos_progedmf_1m.yml
akshaysridhar: @szy21 Right, the coupler toml will override these - but the config files seemed misleading that's all. Thanks!

PR #1985, opened 2026-06-15
Add model run # and start_date as diagnostic global attributes; exclude HDF5 from VM upload

AlexisRenchon:
## Purpose

Make the AMIP diagnostics that are sent to the visualization VM self-describing, and stop shipping restart checkpoints to that VM.

### Global attributes on coupler & land diagnostics
- New helper `Utilities.diagnostics_global_attribs(start_date)` builds the `global_attribs` dict passed to `ClimaDiagnostics.Writers.NetCDFWriter`:
  - `start_date` — always (as a file-level global attribute, in addition to the existing per-variable one).
  - `buildkite_build_number` — the "model run #", read from `ENV["BUILDKITE_BUILD_NUMBER"]`; omitted on local runs where the variable is unset.
- Wired into the three coupler-side writers:
  - `src/SimOutput/diagnostics.jl` (coupler diagnostics)
  - `ext/ClimaCouplerClimaLandExt/climaland_integrated.jl`
  - `ext/ClimaCouplerClimaLandExt/climaland_bucket.jl`

These files (`clima_coupler/`, `clima_land/`) are exactly what gets rsynced to the VM, so the metadata travels with the data.

> Note: the atmos diagnostics writer is built inside ClimaAtmos and does not expose `global_attribs`, so atmos `.nc` files are not covered here.

### Exclude HDF5 restart checkpoints from the VM upload
- Added `--exclude='*.hdf5'` to the rsync in `.buildkite/amip/pipeline.yml`. ClimaViz/ClimaAnalysis only read the NetCDF diagnostics; the checkpoints (~2.4 GB/run) otherwise fill the VM's 16 GB ephemeral disk. Checkpoints are still written/kept on the cluster for restarts.

## Verification
- Edited files parse cleanly.
- Helper verified: local run → `{"start_date" => ...}`; with `BUILDKITE_BUILD_NUMBER=1234` → adds `"buildkite_build_number" => "1234"`.
- After the first AMIP run on this code, confirm with `ncdump -h` on a `clima_coupler`/`clima_land` `.nc` on the VM that `buildkite_build_number` is present (verifies env propagation through `srun`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.buildkite/amip/pipeline.yml
+++ b/.buildkite/amip/pipeline.yml
@@ -60,7 +60,10 @@ steps:
 
       - label: "Send simulation output to Azure VM for visualization"
         command:
-          - rsync -avz output/amip/output_active/* azure-arenchon:/home/arenchon/GitHub/ClimaViz.jl/data/climacoupler-amip/
+          # Exclude restart checkpoints (*.hdf5): ClimaViz/ClimaAnalysis only read the
+          # NetCDF diagnostics, and the checkpoints (~2.4 GB/run) otherwise fill the VM's
+          # 16 GB ephemeral disk. They remain written/kept on the cluster for restarts.
+          - rsync -avz --exclude='*.hdf5' output/amip/output_active/* azure-arenchon:/home/arenchon/GitHub/ClimaViz.jl/data/climacoupler-amip/
 
       - label: ":envelope: Slack report: current AMIP"
         command:
--- a/ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
+++ b/ext/ClimaCouplerClimaLandExt/climaland_bucket.jl
@@ -192,8 +192,13 @@ function BucketSimulation(
 
     # Add diagnostics
     if use_land_diagnostics
-        output_writer =
-            CD.Writers.NetCDFWriter(domain.space.subsurface, output_dir; start_date)
+        global_attribs = Utilities.diagnostics_global_attribs(start_date)
+        output_writer = CD.Writers.NetCDFWriter(
+            domain.space.subsurface,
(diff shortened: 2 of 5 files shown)


AlexisRenchon:
@nefrathenrici Hey Nat, I am not sure who should review this...

The goal of this PR is to have the coupler AMIP diagnostics (netcdf files) have a global attribute of the model run number. My intention is to have it for the dashboards, but in general, I think it will be good to have. Right now there's no way to know what model was run from the diagnostics.

AlexisRenchon:
I think alternatively, I could just send a .txt file containing the model run #. That would just be a 1 line PR in .buildkite (along with sending data to the VM). maybe it's better

nefrathenrici:
> I think alternatively, I could just send a .txt file containing the model run #. That would just be a 1 line PR in .buildkite (along with sending data to the VM). maybe it's better

I think it depends on what we will use this for -- it is currently specific to buildkite. Do you think we will expand the functionality to include more global attributes?

ph-kev:
I think it is okay to put the buildkite build number in the global attributes in the NetCDF files. I can see this being used in other contexts such as putting the hash of the commit. 

If we don't want to do that for whatever reason, I would imagine that you can just access the environment variable `BUILDKITE_BUILD_NUMBER` in the visualization step in Buildkite.

nefrathenrici (approved):

ph-kev (approved):

ph-kev:
One thing that would be good to check if you have time is the restarts and if the `start_date` attribute still make sense there. I think it should be fine though.

PR #1993, opened 2026-06-23
Dr/time step sweep

daverumph:
## Purpose 
This PR is never intended to be merged into main.  It exists to use the CI machinery to run a step time sweep test on the CI cluster.

----


.buildkite/pipeline.yml
```diff
@@ -45,6 +46,7 @@ steps:
     concurrency: 1
     concurrency_group: 'depot/climacoupler-ci'
     agents:
+      slurm_gres: "gpu:type:nvidia_l40s:1"
```
akshaysridhar: `gpu:nvidia_l40s:1` should fix this!

PR #1998, opened 2026-06-25
wire in the snowfall

simone-silvestri:
with the new ClimaOcean (0.10) we also have a snow model attached to the sea ice by default.
This PR wires in the snow model
- the top temperature we solve for is the snow temperature
- the resistance for temperature computation is the resistance in series $\kappa^{snow} / h^{snow} + \kappa^{ice} / h^{ice}$
- we reroute the snowfall to increase the snow layer above sea ice

### DIFF
--- a/ext/ClimaCouplerCMIPExt/clima_seaice.jl
+++ b/ext/ClimaCouplerCMIPExt/clima_seaice.jl
@@ -213,6 +213,8 @@ Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:ice_concentration}) =
     sim.ice.model.ice_concentration
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:ice_thickness}) =
     sim.ice.model.ice_thickness
+Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:snow_thickness}) =
+    sim.ice.model.snow_thickness
 # Internal temperature in Kelvin (same convention as :surface_temperature)
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:internal_temperature}) =
     sim.ocean_ice_interface.temperature + sim.ice_properties.C_to_K
@@ -230,9 +232,13 @@ Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:surface_direct_albedo}) 
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:surface_diffuse_albedo}) =
     eltype(sim.ice.model)(0.7)
 
+top_thermodynamics(sim::ClimaSeaIceSimulation) =
+    isnothing(sim.ice.model.snow_thermodynamics) ? sim.ice.model.ice_thermodynamics :
+    sim.ice.model.snow_thermodynamics
+
 # Surface temperature from last timestep (Celsius → Kelvin)
 Interfacer.get_field(sim::ClimaSeaIceSimulation, ::Val{:surface_temperature}) =
-    sim.ice_properties.C_to_K + sim.ice.model.ice_thermodynamics.top_surface_temperature
+    sim.ice_properties.C_to_K + top_thermodynamics(sim).top_surface_temperature
 
 """
     FluxCalculator.compute_surface_fluxes!(csf, sim::ClimaSeaIceSimulation, atmos_sim, thermo_params)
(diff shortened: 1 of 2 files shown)


ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -259,28 +265,37 @@ NVTX.@annotate function FluxCalculator.compute_surface_fluxes!(
 
     # Sea ice parameters for `update_T_sfc_cb` (load into boundary-space scratch Fields first
     # to avoid GPU scalar indexing when building the closure)
+
+    # Conductive resistance of the column R = h_ice/k_ice + h_snow/k_snow [m² K W⁻¹].
+    # Snow and ice add in series; with no snow layer (or h_snow = 0) this reduces to h_ice/k_ice.
+    ice_heat_flux = sim.ice.model.ice_thermodynamics.internal_heat_flux
+    k_ice = hasfield(typeof(ice_heat_flux), :conductivity) ? FT(ice_heat_flux.conductivity) :
+        convert(FT, 2) # default conductivity [W m⁻¹ K⁻¹]
```
hgpeterson: Maybe add a warning for the case when 2 W m⁻¹ K⁻¹ gets used here? Although then it will trigger every time we compute surface fluxes... so perhaps something during initialization?

ext/ClimaCouplerCMIPExt/skin_temperature.jl
```diff
@@ -3,32 +3,34 @@
...
 Solve the steady-state flux balance for surface temperature Tₛ:
 
-            κ
-Jᵃ(Tₛ) + --- (Tₛ - Tᵢ) = 0
-            δ
-
-where Jᵃ is the net upward surface flux and κ/δ·(Tₛ - Tᵢ) is the conductive flux.
+         Tₛ - Tᵢ
+Jᵃ(Tₛ) + ──────── = 0
+            R
```
hgpeterson:
```suggestion
Jᵃ(Tₛ) + (Tₛ - Tᵢ)/R = 0
```

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -470,6 +481,16 @@ function FieldExchanger.update_sim!(sim::ClimaSeaIceSimulation, csf)
             (OC.interior(ice_concentration, :, :, 1) .> 0) .*
             (-(1 .- α) .* remapped_SW_d .- ϵ .* remapped_LW_d)
     end
+
+    # Snow precipitation drives snow accumulation (sign flip: downward atmospheric mass 
+    # flux to positive `snowfall`).
+    snowfall = sim.ice.model.snowfall
+    if snowfall isa OC.Field
```
hgpeterson, line 489: Just curious, when is `snowfall` not an `OC.Field`, and what else would it be in that case? Is it when it is not defined?

ext/ClimaCouplerCMIPExt/clima_seaice.jl
```diff
@@ -259,28 +265,37 @@ NVTX.@annotate function FluxCalculator.compute_surface_fluxes!(
...
     Interfacer.get_field!(csf.scalar_temp1, sim, Val(:ice_thickness))
+    R = csf.scalar_temp1
+    snow_thermo = sim.ice.model.snow_thermodynamics
+    if isnothing(snow_thermo)
+        R .= R ./ k_ice
+    else
+        k_snow = FT(snow_thermo.internal_heat_flux.conductivity)
+        Interfacer.get_field!(csf.scalar_temp2, sim, Val(:snow_thickness))
+        R .= R ./ k_ice .+ csf.scalar_temp2 ./ k_snow
+    end
```
hgpeterson, line 284: Similarly, it'd be nice to know when this gets triggered.

hgpeterson (approved):
Looks good, thanks @simone-silvestri! I think just adding some kind of `@info` upon initialization to say whether or not the snow model is being used (an what its parameters are) would be all I'd want. Maybe this already happens somewhere in ClimaOcean internals?

ext/ClimaCouplerCMIPExt/clima_seaice.jl
akshaysridhar: This is fixed at model construction so in our CMIP runs we'd always be running the else branch.

akshaysridhar (approved):

PR #1999, opened 2026-06-25
change resolution of nightly amip

szy21:
## Purpose 

Use the same vertical resolution and a lower horizontal resolution than our target. We can only run it for a shorter period overnight, but I think this setup is more useful for diagnosing problems.

These nightly amips will probably break with the latest ClimaAtmos. We will look into it.

## Content

----


szy21:
After review I will merge it manually as it only changes nightly configs.

costachris (approved):

PR #2001, opened 2026-06-26
Add config option for ORCA grid 

### LINKED ISSUE #1379: investigate open/closed basins in bathymetry
In our current simulations using OceananigansSimulation, the Mediterranean sea is a closed basin (i.e. not labeled as ocean), but we want it to be an open basin. This behavior is seen when running at 1 degree resolution on a capped Lat/Lon grid. The basin becomes open if we increase the resolution slightly or (in the case of standalone Oceananigans) use a TripolarGrid. We could try one of these options, or increase the value of `major_basins` when regridding the bathymetry. Note that we want to move towards using the TripolarGrid anyway, so that could be a good solution.

akshaysridhar:
Introduces the option to use `CO.OceanConfigurations.orca_ocean` and makes this the default configuration for weekly (CMIP) longruns. 

Closes https://github.com/CliMA/ClimaCoupler.jl/issues/1379


ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -87,6 +87,48 @@ function tripolar_ocean_simulation(
...
+
+Build an Oceananigans `Simulation` on either the standard one-degree tripolar grid
+or the NEMO eORCA1 mesh (`orca`).
+"""
+function ocean_simulation(arch, ocean_grid::Symbol; simple_ocean, closure, clock, kwargs...)
+    substeps = simple_ocean ? 70 : 150
+
+    if ocean_grid == :orca
+        if simple_ocean
+            @info "simple_ocean=true: using one_deg_tripolar grid instead of ORCA1 grid"
```
hgpeterson:
```suggestion
            @warn "`simple_ocean=true` not supported for ORCA1 grid,  using one_deg_tripolar grid instead"
```

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -87,6 +87,48 @@ function tripolar_ocean_simulation(
...
+                arch;
+                closure,
+                clock,
+                substeps,
+                kwargs...,
+            )
+        end
+    end
+
+    @info "Using standard one-degree tripolar ocean grid"
```
hgpeterson:
```suggestion
    @info "Using one-degree tripolar ocean grid"
```

ext/ClimaCouplerCMIPExt/oceananigans.jl
```diff
@@ -153,9 +197,6 @@ function OceananigansSimulation(
     end
 
     # Tripolar active_cells_map kernels exceed the 4 KiB sm_60 limit on P100.
```
hgpeterson:
```suggestion
```

hgpeterson (approved):

PR #2004, opened 2026-06-29
use default land model

kmdeck:
## Purpose 
This switches to using the default land model, rather than one constructed by the coupler with different parameterization choices. The motivation is that the land team is developing and improving a model which is created with our default constructor, and the coupler should use this instead of something else.

The parameterization changes are:
- Pmodel instead of Farquhar + Medlyn
- Peicewise soil moisture stress instead of Tuzet

In order to use the pmodel, the land model sets initial conditions using making use of T, P, q, c_co2 for the atmosphere at the surface. *these are not required* to set up the land simulation, as the land can be restarted from its' Y only. However, the atmosphere values provide a very convenient guess for reasonable initial conditions, so we use them.

For now, we approximate these using the orographic adjusted T, a surface elevation correction for Pressure and air density, and q = q_sat given the above assumptions. This can be improved in the future if it is any issue. It may not be, though, as the whole point is just to provide reasonable IC that shorten spinup time, and these may be reasonable enough already.

## Content

----


szy21 (approved):

AlexisRenchon:
It seems to me that all that needs to be done is rebase this branch once we release a new version of ClimaLand (to have the time integrated variables in latest release), and it should be ready to go?

kmdeck:
> It seems to me that all that needs to be done is rebase this branch once we release a new version of ClimaLand (to have the time integrated variables in latest release), and it should be ready to go?

i hope so! we may need to save the cache (to get p.snow.T_sfc) for exact reproducibility for the tests, but the coupler has options for saving Y and p in the restart file.

experiments/AMIP/Manifest-v1.11.toml
```diff
@@ -497,9 +497,11 @@ weakdeps = ["CUDA"]
 
 [[deps.ClimaLand]]
 deps = ["ClimaComms", "ClimaCore", "ClimaDiagnostics", "ClimaParams", "ClimaTimeSteppers", "ClimaUtilities", "Dates", "DocStringExtensions", "Insolation", "Interpolations", "LazyArtifacts", "LazyBroadcast", "LinearAlgebra", "NCDatasets", "NVTX", "RootSolvers", "StaticArrays", "SurfaceFluxes", "Thermodynamics"]
-git-tree-sha1 = "16e3ad009325373200d8beda4d57e1b5b403b776"
+git-tree-sha1 = "8371779e614c602d51311807700f82cc590f02db"
+repo-rev = "kd/update_subseasonal_ic_pmodel"
+repo-url = "https://github.com/CliMA/ClimaLand.jl.git"
 uuid = "08f4d4ce-cf43-44bb-ad95-9d2d5f413532"
-version = "1.11.2"
+version = "1.11.1"
```
kmdeck: my CL is not rebased off of CL main. If this first test passes, I will rebase the CL branch and merge it, then change this coupler branch to run with climaland main, and try a nightly run

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
```diff
@@ -150,30 +150,9 @@ function ClimaLandSimulation(
...
-    model = CL.LandModel{FT}(
-        forcing,
-        LAI,
-        toml_dict,
-        domain,
-        dt;
-        prognostic_land_components,
-        canopy,
-    )
+    model =
```
costachris, line 155: This PR will also change the soil moisture stress model (from TuzetMoistureStressModel to PiecewiseMoistureStressModel) since the canopy isn’t constructed and passed anymore, so the LandModel default is used. That can be noted in the PR as an additional difference

ext/ClimaCouplerClimaLandExt.jl
```diff
@@ -18,6 +18,11 @@ import Dates
 import ClimaUtilities.TimeVaryingInputs:
     LinearInterpolation, PeriodicCalendar, TimeVaryingInput
 import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.Regridders: InterpolationsRegridder
+import Interpolations
+regridder_type = :InterpolationsRegridder
```
costachris: `make_set_initial_state_from_era5land` defaults to `Interpolations.Linear()` for `tsn` / `swvl` / `stl`. The coupler reads `skt` from the same file with nearest-neighbor `Constant()`, then uses it as `p.drivers.T` (canopy T *and* P-model `T_air`). Previous code interped `skt` to `Y.canopy.energy.T` with Linear. I think just use the same interpolator for `skt` as for the rest of the file, and pass the interpolation_method into `make_set_initial_state_from_era5land()` so they're always the same.

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
```diff
@@ -164,8 +167,31 @@ function _coupler_set_ic!(Y, p, t, model::CL.Bucket.BucketModel, atmos_T, set_ic
...
+"""
+function _coupler_set_ic!(Y, p, t, model::CL.LandModel, T_sfc, surface_elevation, set_ic!)
+    FT = eltype(T_sfc)
+    p.drivers.T .= T_sfc
+    elevation_correction = @. exp(-surface_elevation/FT(8000))
+    p.drivers.P .= FT(101325) .* elevation_correction
+    thermo_params = LP.thermodynamic_parameters(model.canopy.earth_param_set)
+    R = TD.Parameters.R_d(thermo_params)
+    ρ_sfc = @. p.drivers.P/(R*p.drivers.T)
+    @. p.drivers.q = TD.q_vap_saturation(thermo_params, T_sfc, ρ_sfc)
```
costachris, line 195: This is assuming saturated air at the surface. We could instead pass the real atmos `T/P/q/CO2` (already available before land is built) if it ends up mattering, or recompute the P-model capacities once land has the real atmos air. Maybe add a comment or TODO noting this assumption?

costachris (approved):

ext/ClimaCouplerClimaLandExt.jl
```diff
@@ -18,6 +18,11 @@ import Dates
...
+extrapolation_bc = (Interpolations.Periodic(), Interpolations.Flat(), Interpolations.Flat())
+interpolation_method = Interpolations.Constant()
```
imreddyTeja: I think these should either be made into `const`s with more specific names, or just placed into the one function they are used in

ext/ClimaCouplerClimaLandExt.jl
```diff
@@ -18,6 +18,11 @@ import Dates
 import ClimaUtilities.TimeVaryingInputs:
     LinearInterpolation, PeriodicCalendar, TimeVaryingInput
 import ClimaUtilities.SpaceVaryingInputs: SpaceVaryingInput
+import ClimaUtilities.Regridders: InterpolationsRegridder
+import Interpolations
```
imreddyTeja, line 22: also imported on line 39

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
```diff
@@ -164,8 +167,31 @@ function _coupler_set_ic!(Y, p, t, model::CL.Bucket.BucketModel, atmos_T, set_ic
...
+    p.drivers.c_co2 .= FT(4.2e-4)
```
imreddyTeja: Could you add a comment that this value is set to match the default from `PrescribedAtmosphere` in ClimaLand, or something along those lines

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
```diff
@@ -164,8 +167,31 @@ function _coupler_set_ic!(Y, p, t, model::CL.Bucket.BucketModel, atmos_T, set_ic
...
+required to compute radiation in the atmosphere.
+
+Note that when running a restarted simulation, any values set here will be overwritten by
+the saved state and cache values in the restart file.
+"""
+function _coupler_set_ic!(Y, p, t, model::CL.LandModel, T_sfc, surface_elevation, set_ic!)
+    FT = eltype(T_sfc)
+    p.drivers.T .= T_sfc
+    elevation_correction = @. exp(-surface_elevation/FT(8000))
+    p.drivers.P .= FT(101325) .* elevation_correction
```
imreddyTeja: It might be a good idea to add a comment explaining what these numbers represent. It wasn't immediately obvious to me, and I had to look in ClimaLand to figure out what they mean.

imreddyTeja (approved):

ext/ClimaCouplerClimaLandExt/climaland_integrated.jl
kmdeck: I updated the PR description.

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
kmdeck: I added a comment in the PR describing this, and will add a comment to the code.

kmdeck: I think it would be more desirable not to have to pass in the atmos IC to set up the land IC, so hopefully these work fine.

kmdeck: We will recompute the capacities during first timestep using the values in p.drivers (as they are prognostic vars), which the coupler updates correctly. This is only about setting IC. I think it should be OK

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
kmdeck: I added:
```
    p.drivers.c_co2 .= FT(4.2e-4) # Reasonable value; in the future we may wish to ensure this is consistent with atmos, but this is only used for the initial conditions
```

ext/ClimaCouplerClimaLandExt/climaland_helpers.jl
kmdeck: I added some comments. thanks for pointing this out.

ext/ClimaCouplerClimaLandExt.jl
kmdeck: moved to the function

ext/ClimaCouplerClimaLandExt.jl
kmdeck: I changed this to be linear interpolation. and moved the definitions here to be clearer. I can make it so make_set...takes the interpolation method too, but that would require another PR in CL. let me know what you prefer!

I dont think it matters if we use the same "T" for canopy T and Pmodel Tair because this is just about setting up IC that are not too far from steady state. But did you want me to add some comments making this clearer?

kmdeck: (I changed all the functions to now use "surface T" instead of atmos T, in the coupler, so hopefully that is clearer)

PR #2005, opened 2026-06-29
Add MAC LWP dataloader

nefrathenrici:
This PR adds a dataloader for the MAC LWP dataset and updates the `pressure_levels.jl` calibration config to use LWP.

### DIFF
--- a/experiments/calibration/amip/config/pressure_levels.jl
+++ b/experiments/calibration/amip/config/pressure_levels.jl
@@ -14,12 +14,12 @@ const CALIBRATE_CONFIG = CalibrationTools.CalibrateConfig(;
     config_file,
     # Note: Pressure-level variables require model output with
     # pressure_coordinates: true in config
-    short_names = ["ta", "hur"],
+    short_names = ["lwp", "ta", "hur"],
     minibatch_size = 1,
     n_iterations = 5,
     sample_date_ranges,
     extend = Dates.Month(1),
-    spinup = Dates.Day(0),
+    spinup = Dates.Day(7),
     output_dir,
     rng_seed = 42,
 )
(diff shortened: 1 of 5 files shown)


experiments/calibration/amip/config/pressure_levels.jl
```diff
@@ -14,12 +14,12 @@ const CALIBRATE_CONFIG = CalibrationTools.CalibrateConfig(;
     config_file,
     # Note: Pressure-level variables require model output with
     # pressure_coordinates: true in config
-    short_names = ["ta", "hur"],
```
costachris, line 17: Do we only want to do LWP vs all three? If the aim here is primarily for CI and the error wasn't decreasing before, that's reasonable and the others can be added back later.

costachris (approved):
Thanks!

experiments/calibration/amip/config/pressure_levels.jl
costachris: also, may be better to map to `mac_lwp` or similar to be explicit as there are different sources for LWP estimates

nefrathenrici: The aim here is primarily CI but I can update it to use all three variables.

nefrathenrici: I will also verify that the composite dataloader is using MAC data instead of MODIS.

ph-kev: Are we running any real calibration with this on CI? I'm only aware of the calibration test that run on [ClimaCoupler-CI](https://buildkite.com/clima/climacoupler-ci) for each PR.

src/CalibrationTools.jl
```diff
@@ -948,4 +948,73 @@ function emulate_diagnostics!(sim, t_end)
     return nothing
 end
 
+
+"""
+    MACDataLoader
+
+A struct for loading preprocessed MAC data on pressure levels as `OutputVar`s.
```
ph-kev: Can you move this section to be with the rest of the other data loaders. I don't want a section of data loaders, stuff about emulating diagnostics and regridding, and then another data loader.

src/CalibrationTools.jl
```diff
@@ -948,4 +948,73 @@ function emulate_diagnostics!(sim, t_end)
...
+    return preprocess(loader, var, Val(Symbol(short_name)))
+end
+
+function preprocess(::MACDataLoader, var, ::Union{Val{:iwp}, Val{:lwp}})
+    var = _preprocess_var(var)
+    not_nans = filter(!isnan, var.data)
+    global_mean = sum(not_nans) / length(not_nans)
+    @info "$(ClimaAnalysis.short_name(var)): Imputing global mean $global_mean for NaNs"
+    replace!(x -> isnan(x) ? global_mean : x, var)
+    var = ClimaAnalysis.convert_units(var, "kg m^-2", conversion_function = x -> 0.001 * x)
```
ph-kev: Is this a good idea for data imputation? I don't trust replacing `NaN`s with the global mean.

experiments/calibration/amip/config/pressure_levels.jl
nefrathenrici: This should run weekly in https://buildkite.com/clima/climacoupler-calibration/

src/CalibrationTools.jl
nefrathenrici: Agreed, we shouldn't replace NaNs here

PR #2006, opened 2026-06-30
Re-enable ITime in ocean and sea-ice models

### LINKED ISSUE #1992: Fix clock in CMIP configurations
Following PR#1971, ITime is temporarily disabled in CMIP simulations. 

TODO: Update the model clock kwarg in upstream packages and reset `use_itime=true` in CMIP configurations. Requires changes to 
- [ ] NumericalEarth
- [ ] ClimaOcean

hgpeterson:
Closes #1992

Following the update to ClimaOcean v0.10 (#1971), `stop_time` would default to `Inf` in the ocean/sea-ice models, which forced us to use `Float`s instead of `Date`s. [This PR](https://github.com/NumericalEarth/NumericalEarth.jl/pull/395) in NumericalEarth.jl enables us to set `stop_time` as a kwarg. 

This PR:
- [x] Updates how `clock` and `stop_time` are set in the ocean model
- [x] Updates how `clock` and `stop_time` are set in the sea-ice model
- [x] Re-enables `use_itime: true` in config files

This PR also fixes a completely unrelated bug in `experiments/test/compare_cmip.jl` that emerged from an update in NumericalEarth (see [this PR](https://github.com/NumericalEarth/NumericalEarth.jl/pull/346); ocean surface salinity is now passed as a `view` instead of with `interior`, changing how we need to define `_compare(v1, v2)` for `SubArray`s).


hgpeterson:
The [NumericalEarth PR](https://github.com/NumericalEarth/NumericalEarth.jl/pull/395) is merged, but because NumericalEarth is not a direct dependency of ClimaCoupler (it comes in through ClimaOcean), we'll have to wait for a new release there for this PR to work.

akshaysridhar (approved):

hgpeterson:
An issue that came up that I believe is unrelated to the clock changes: `experiments/test/restart_cmip.jl` started to fail [here](https://github.com/CliMA/ClimaCoupler.jl/blob/77287c5563a8950978035845f12dff4fe0debe25/experiments/test/restart_cmip.jl#L115-L120) with a CUDA scalar indexing error. I believe it has something to do with how the sea-ice model is initialized with `SubArray`s of `OffsetArray`s of `CuArray`s from the `ocean` (e.g. `IceWaterThermalEquilibrium.salinity`). To fix the test, I had to add lines of the form
```julia
h1 = OffsetArrays.no_offset_view(on_architecture(CPU(), v1))
```
to `experiments/test/compare_cmip.jl`, which required adding `OffsetArrays` to our CMIP `Project.toml`. 

@simone-silvestri Does it make sense to you why separate initializations of the sea-ice model might have different array offsets in their `view`s?

@akshaysridhar Let me know if you're okay with adding `OffsetArrays` or if we should find a different solution.

simone-silvestri:
In https://github.com/NumericalEarth/NumericalEarth.jl/pull/346 we have changed the ocean surface salinity to pass 
`view(S.data, :, :, Nz)` instead of `interior(S, :, :, Nz)` because we need the halos of the surface salinity to compute fluxes inside the halo itself.

https://github.com/NumericalEarth/NumericalEarth.jl/blob/01ab2a6fc1d5e3d6e6b77c20897936495e5c914b/src/Oceans/Oceans.jl#L83-L91

Therefore, the output of that function switched from begin a subarray of a plain array to a subarray of an offsetarray of an array.

I think `on_architecture(CPU(), subarray)` is enough because it moves everything on the CPU so you avoid the scalar indexing

experiments/test/compare_cmip.jl
```diff
@@ -144,15 +146,24 @@ function _compare(pass, v1::T, v2::T; name, ignore) where {T <: CC.DataLayouts.A
...
 function _compare(
     pass,
     v1::SubArray{FT},
     v2::SubArray{FT};
     name,
     ignore,
 ) where {FT <: AbstractFloat}
-    return pass && _compare(collect(v1), collect(v2); name, ignore)
+    h1 = OffsetArrays.no_offset_view(on_architecture(CPU(), v1))
+    h2 = OffsetArrays.no_offset_view(on_architecture(CPU(), v2))
```
simone-silvestri:
```suggestion
    h1 = on_architecture(CPU(), v1)
    h2 = on_architecture(CPU(), v2)
```

experiments/test/compare_cmip.jl
```diff
@@ -7,6 +7,8 @@ import ClimaCore as CC
 import Oceananigans as OC
 import ClimaSeaIce as CSI
 using ClimaSeaIce.SeaIceThermodynamics.HeatBoundaryConditions: IceWaterThermalEquilibrium
+using Oceananigans.Architectures: on_architecture, CPU
+import OffsetArrays
```
simone-silvestri:
```suggestion
```

experiments/test/compare_cmip.jl
hgpeterson: Hmm this doesn't work because then we do `Array(h1)` in `_error()` on R25 we get a `DimensionMismatch` error:
```julia
DimensionMismatch: axes must agree, got (Base.OneTo(370), Base.OneTo(190), Base.OneTo(1)) and (Base.IdentityUnitRange(OffsetArrays.IdOffsetRange(values=-4:365, indices=-4:365)), Base.IdentityUnitRange(OffsetArrays.IdOffsetRange(values=-4:185, indices=-4:185)), Base.OneTo(1))
```
So I think we need the `no_offset_view`, unless you know a way to make `_error()` work for `OffsetArray`s?

hgpeterson: Aha, I think I figured out a simpler way: just do
```
h1 = vec(on_architecture(CPU(), v1))
h2 = vec(on_architecture(CPU(), v2))
```
so that there's no `DimensionMismatch` issue when `Array()` is called in `_error()`!

simone-silvestri: I think you could also do 
```julia
h1 = parent(on_architecture(CPU(), v1))
h2 = parent(on_architecture(CPU(), v2))
```
to retrieve the underying array?

hgpeterson: Hmm just tried that but it errors. I think it needs two `parent`s? But then it might fail for other cases that just need one...

hgpeterson: Test passes with 
```
    h1 = parent(parent(on_architecture(CPU(), v1)))
    h2 = parent(parent(on_architecture(CPU(), v2)))
```
What do you think is the more canonical solution? I'll defer to your judgement.

simone-silvestri (approved):

PR #2013, opened 2026-07-02
Use dates instead of floating-point for spinup

ph-kev:
The spinup was computed using `spinup * 31 * 86400.0` and we rely on `ClimaAnalysis.window` to find the closest month. For most cases, this works, but the [latest Buildkite run](https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/1288#019f208e-f5be-4f7c-a6f0-721e80410b67) shows that for short simulation run, it doesn't work.

To fix this, we use absolute time (dates) instead of relative time (floating-point).

### DIFF
--- a/ext/ClimaCouplerMakieExt/leaderboard/leaderboard.jl
+++ b/ext/ClimaCouplerMakieExt/leaderboard/leaderboard.jl
@@ -52,9 +52,9 @@ function Plotting.compute_leaderboard(
         obs_var = obs_var_dict[short_name](sim_var.attributes["start_date"])
 
         # Remove first spin_up_months from simulation
-        spinup_cutoff = spinup * 31 * 86400.0
-        ClimaAnalysis.times(sim_var)[end] >= spinup_cutoff &&
-            (sim_var = ClimaAnalysis.window(sim_var, "time", left = spinup_cutoff))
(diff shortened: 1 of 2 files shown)


ext/ClimaCouplerMakieExt/leaderboard/leaderboard.jl
```diff
@@ -52,9 +52,9 @@ function Plotting.compute_leaderboard(
         obs_var = obs_var_dict[short_name](sim_var.attributes["start_date"])
 
         # Remove first spin_up_months from simulation
-        spinup_cutoff = spinup * 31 * 86400.0
-        ClimaAnalysis.times(sim_var)[end] >= spinup_cutoff &&
-            (sim_var = ClimaAnalysis.window(sim_var, "time", left = spinup_cutoff))
+        spinup_cutoff_date = first(ClimaAnalysis.dates(sim_var)) + spinup * Dates.Month(1)
```
akshaysridhar, line 55: Awesome, thank you ; if you can include the corresponding fix in `src/Plotting.jl` where we currently use 
[hardcoded values](https://github.com/CliMA/ClimaCoupler.jl/blob/29e88d917eecf722e952c1a5359f63e3af8c9a6b/src/Plotting.jl#L146)
we can merge this PR.

akshaysridhar (approved):

PR #2014, opened 2026-07-03
patch release 0.2.2; update manifests

imreddyTeja:
## Purpose 
Weekly release 
## Content
Updates NEWS.md. The last coupler release never got registered. I did that as a prerequisite for this release.
Does not update any compats, just the manifests

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,11 +4,23 @@ ClimaCoupler.jl Release Notes
 `main`
 -------
 
+v0.2.2
+-------
+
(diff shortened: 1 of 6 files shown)


update_deps_coupler.sh
szy21, line 1: do we want to add this file?

szy21 (approved):

update_deps_coupler.sh
imreddyTeja: No... good catch

PR #2017, opened 2026-07-04
use max_newton_iters_ode 3

szy21:
## Purpose 

Temporarily set Newton iterations to 3 until the top boundary issue is sorted out.

## Content

----


akshaysridhar (approved):
LGTM, we can set up an issue to track these config args for future runs

szy21:
> LGTM, we can set up an issue to track these config args for future runs

Yes, that would be good.

PR #2026, opened 2026-07-10
Revamp wall time reporter

hgpeterson:
Removed `capped_geometric_walltime_cb` in `TimeManager` in favor of a `WalltimeReporter` type that
- prints every `"1days"` or at least `(t_end - t_start) / 10` (but never less than `dt_cpl`) by default,
- prints the *coupler* step (no longer `ClimaAtmos`'s), and
- is more compact and visually pleasing.

### DIFF
--- a/ext/ClimaCouplerClimaAtmosExt.jl
+++ b/ext/ClimaCouplerClimaAtmosExt.jl
@@ -727,7 +727,9 @@ function get_atmos_config_dict(
     # can pick up from where we have left. NOTE: This should not be needed, but
     # there is no easy way to initialize ClimaAtmos with a different t_start
     atmos_config["dt_save_state_to_disk"] = coupler_config["checkpoint_dt"]
-    atmos_config["log_progress"] = coupler_config["atmos_log_progress"]
+
+    # disable Atmos's progress logging
+    atmos_config["log_progress"] = false
 
     # The Atmos `get_simulation` function expects the atmos config to contains its timestep size
     # in the `dt` field. If there is a `dt_atmos` field in coupler_config, we add it to the atmos config as `dt`
--- a/src/Input.jl
+++ b/src/Input.jl
@@ -121,6 +121,10 @@ function argparse_settings()
         help = "Time interval for checkpointing [\"90days\" (default)]"
         arg_type = String
         default = "90days"
+        "--walltime_dt"
+        help = "Time interval for walltime reporting [nothing (default): a tenth of the simulation length, at most 1 day and at least one coupling step; allowed formats: \"Nsecs\", \"Nmins\", \"Nhours\", \"Ndays\", \"Nmonths\", \"never\"]"
+        arg_type = String
+        default = nothing
         # Space information
         "--h_elem"
         help = "Number of horizontal elements to use for the atmosphere horizontal space [16 (default)]"
@@ -208,10 +212,6 @@ function argparse_settings()
         help = "An optional YAML file used to overwrite the default model parameters."
         arg_type = String
         default = nothing
-        "--atmos_log_progress"
-        help = "Use the ClimaAtmos walltime logging callback instead of the default ClimaCoupler one [`false` (default), `true`]"
-        arg_type = Bool
-        default = false
         "--atmos_progress_interval"
         help = "Time interval for printing atmosphere progress information [\"never\" (default); allowed formats: \"Nsecs\", \"Nmins\", \"Nhours\", \"Ndays\", \"Nmonths\", \"never\"]"
         arg_type = String
@@ -543,6 +543,15 @@ function get_coupler_args(config_dict::Dict)
     # Checkpointing information
     checkpoint_dt = config_dict["checkpoint_dt"]
 
+    # Walltime reporting information
+    walltime_dt = get(config_dict, "walltime_dt", nothing)
+    if isnothing(walltime_dt)
+        # default to a tenth of the simulation length (capped at one day, but never
+        # shorter than one coupling step)
+        walltime_dt_secs = max(min(float(t_end - t_start) / 10, 86400.0), float(Δt_cpl))
+        walltime_dt = "$(walltime_dt_secs)secs"
+    end
+
     # Atmos progress reporting information
     atmos_progress_interval = config_dict["atmos_progress_interval"]
 
@@ -665,6 +674,7 @@ function get_coupler_args(config_dict::Dict)
         h_elem_coupler,
         saveat,
         checkpoint_dt,
+        walltime_dt,
         atmos_progress_interval,
         detect_restart_files,
         restart_dir,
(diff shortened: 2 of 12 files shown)


hgpeterson:
Example from CI:
```
┌ Info: Progress
│   time = 2010-01-01T00:36:00 (36 m)
│   step = 6 (20.0%)
│   walltime remaining ≈ 20 m 54 s (2026-07-21T13:56:09)
└   sypd ≈ 0.01886
```

src/TimeManager.jl
```diff
@@ -102,27 +101,86 @@ function (::NeverSchedule)(args...)
...
-            nsteps = steps_taken(integrator)
-            # skip first two steps for compilation
-            (nsteps <= 2) && return false
-            return nsteps % five_percent_steps == 0 || ispow2(nsteps)
+struct WalltimeReporter
+    n_calls::Base.RefValue{Int}
+    wall_time_last::Base.RefValue{Float64}
+    sim_time_last::Base.RefValue{Float64}
+    wall_time_elapsed::Base.RefValue{Float64}
+end
```
ph-kev, line 131: This seems like we are reinventing the wheel since isn't this functionality already supported in ClimaUtilities (see docs at https://clima.github.io/ClimaUtilities.jl/dev/onlinelogging/).

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -661,7 +661,9 @@ function get_atmos_config_dict(
     # can pick up from where we have left. NOTE: This should not be needed, but
     # there is no easy way to initialize ClimaAtmos with a different t_start
     atmos_config["dt_save_state_to_disk"] = coupler_config["checkpoint_dt"]
-    atmos_config["log_progress"] = coupler_config["atmos_log_progress"]
+
+    # disable Atmos's progress logging
+    atmos_config["log_progress"] = false
```
ph-kev, line 732: @imreddyTeja Does the performance monitoring use the logs from atmos?

src/TimeManager.jl
```diff
@@ -102,27 +101,86 @@ function (::NeverSchedule)(args...)
...
-first two steps to avoid compilation time noise.
+A callable object that logs the progress of a coupled simulation.
+
+A `WalltimeReporter` is meant to be used as the function of a `Callback`: it is
+called with the `CoupledSimulation` as its only argument and reads the current
+time, start date, time span, and coupling time step from it.
+
+The wall time of the first call is dominated by compilation, so it is not
+measured; instead, the first measurement (between the first and second calls) is 
+scaled up to estimate the compilation-free wall time of the steps before it.
```
ph-kev, line 124: I think @imreddyTeja faced issues with this before, but he can explain it better than I can.

ext/ClimaCouplerClimaAtmosExt.jl
imreddyTeja: I don't think so. I believe it just uses the final reported SYPD, and atmos logging should be disabled

imreddyTeja:
If I remember correctly, after #1581, all coupled runs should be reporting the coupler step, unless explicitly overridden.

src/TimeManager.jl
hgpeterson: That's fair, I initially tried to get things to work the way I wanted with `ClimaUtilities.OnlineLogging.WallTimeInfo`, but I found that `_update!(wt::WallTimeInfo, integrator)` and `report_walltime(wt, integrator)` were too restrictive for what I wanted to do (we don't use an `integrator` for the coupler, for instance). To keep things streamlined, I decided to make a simple, custom type catered for the coupler. Would you instead recommend still using the `WallTimeInfo` type from `ClimaUtilities` but building new functions around it?

hgpeterson:
> If I remember correctly, after #1581, all coupled runs should be reporting the coupler step, unless explicitly overridden.

Hmm but I think that PR builds the walltime reporter on the atmos's `integrator`:
```julia
(coupled_sim) -> report_walltime(wt, coupled_sim.model_sims.atmos_sim.integrator)
```
Causing it to report the atmos's step rather than the coupler's step. For example, in the [latest longrun](https://buildkite.com/clima/climacoupler-longruns/builds/1182/list?sid=019f7964-bdb5-4431-b611-3663f504237b&tab=output) you can see
```
┌ Info: Progress
│   simulation_time = "5 weeks, 1 day"
│   n_steps_completed = 26352
│   ...
```
The coupler timestep is `dt_cpl = 360`. 5 weeks + 1 day = 3110400 s, so that would be 3110400 / 360 = 8640 steps, not 26352, as reported. That's because the atmos timestep is `dt_atmos = 120`, which checks out: 3110400 / 120 = 26352.

Since the coupler doesn't have an integrator, I thought it more sensible to build a small amount of custom code to report its progress, rather than trying to force it to fit into the `ClimaUtilities` mold.

src/TimeManager.jl
ph-kev: For now, I think it is fine to merge this in if you want the improved logging now (assuming it doesn't break anything since I think some of the performance monitoring relies on logging). In the future, the `ClimaUtilities.OnlineLogging.WallTimeInfo` can be refactored to be more composable or states what interface that need to be implemented.

I don't think it is a good idea to build functions around the `WallTimeInfo` type.

ext/ClimaCouplerClimaAtmosExt.jl
hgpeterson: Where does the performance monitoring code live? I can check to make sure this PR doesn't break anything.

src/TimeManager.jl
hgpeterson: Discussed offline. Issue to track this added to ClimaUtilities here: https://github.com/CliMA/ClimaUtilities.jl/issues/237

ext/ClimaCouplerClimaAtmosExt.jl
imreddyTeja: I believe it is somewhere [here](https://github.com/petebachant/clima-perf/blob/main/calkit.yaml)

src/Input.jl
```diff
@@ -543,6 +543,13 @@ function get_coupler_args(config_dict::Dict)
     # Checkpointing information
     checkpoint_dt = config_dict["checkpoint_dt"]
 
+    # Walltime reporting information
+    walltime_dt = get(config_dict, "walltime_dt", nothing)
+    if isnothing(walltime_dt)
+        # default to quarter of the simulation length (capped at one day)
```
imreddyTeja: could we default to something a little more frequent? If it called four times, but the first time is skipped, then the first useful report is halfway through the simulation

imreddyTeja (approved):

src/Input.jl
hgpeterson: Sure, I made it 1/10th now and also made sure that it is never less than the coupler step.

PR #2030, opened 2026-07-10
Add weatherbench2 config

costachris:
Add weatherbench2 config for 10 day global weather forecast runs.

### DIFF
--- a/config/weather/weatherbench_progedmf.yml
+++ b/config/weather/weatherbench_progedmf.yml
@@ -1,0 +1,105 @@
+# Under development — not run in CI yet. Currently used for WeatherBench / ProgEDMF
+# experiments on Derecho (NCAR); paths below assume that machine.
+FLOAT_TYPE: "Float32"
+albedo_model: "CouplerAlbedo"
+atmos_config_file: "config/atmos_configs/climaatmos_progedmf.yml"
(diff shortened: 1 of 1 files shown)


config/weather/weatherbench_progedmf.yml
```diff
@@ -0,0 +1,103 @@
+FLOAT_TYPE: "Float32"
+albedo_model: "CouplerAlbedo"
+atmos_config_file: "config/atmos_configs/climaatmos_progedmf.yml"
+strict_params: false
+mode_name: "subseasonal"
+
+# derecho machine data paths
+era5_initial_condition_dir: "/glade/campaign/univ/ucit0011/cchristo/wxquest_data/initial_conditions/initial_conditions_0p25deg_dev/no_hz"
+coupler_output_dir: "/glade/derecho/scratch/cchristo/coupler_runs_v2/weatherbench"
```
szy21, line 11: we usually don't add machine dependent configs. do you want to run this in ci, or do you just want to merge in this config?

costachris: agree a case should be in CI long term. For now it requires derecho but I can update after the IC.jl changes are finalized.

szy21 (approved):

PR #2034, opened 2026-07-14
Use artifacts for EN4 T/S init

### LINKED ISSUE #1533: Add ECCO datasets to ClimaArtifacts

hgpeterson:
Closes #1533 

See corresponding [ClimaArtifacts PR](https://github.com/CliMA/ClimaArtifacts/pull/178).


akshaysridhar (approved):
See also https://github.com/CliMA/ClimaArtifacts/pull/178 (adds artifacts for EN4 T, S) ; source : https://www.metoffice.gov.uk/hadobs/en4/

PR #2036, opened 2026-07-15
Update RRTMGP API (includes gates for CMIP env compat with older version) 

akshaysridhar:
- ClimaCouplerClimaAtmosExt: query the radiation method via RRTMGP.radiation_method(solver) instead of reading the removed `.radiation_mode` field off the solver, which would throw on non-gray coupled runs (the neighboring flux accesses were migrated, this one missed).
- Rename rrtmgp_model -> rrtmgp_solver to match ClimaAtmos.


akshaysridhar:
@szy21 The `pkgversion` checks will be removed once the NumericalEarth changes are in; right now it's a guard mainly to keep both AMIP and CMIP tests / restart checks running. I'll update the coupler version when that happens.

szy21 (approved):

PR #2039, opened 2026-07-16
Detach ClimaOcean to ClimaCoupler (Part 1)

simone-silvestri:
There are not too many lines we need to port over to completely detach ClimaCoupler from ClimaOcean and just depend on Oceananigans and ClimaSeaIce. 

This PR does the initial port.

PR #2050, opened 2026-07-20
Update CMIP env to use Atmos 0.42, RRTMGP 0.22.2

akshaysridhar:
Update deps in CMIP env.


szy21 (approved):
Hmm, when I tried to update it I got a compilation error, and it seems NumericalEarth 0.6.0 is not compatible with ClimaOcean 0.10.0

akshaysridhar:
> Hmm, when I tried to update it I got a compilation error, and it seems NumericalEarth 0.6.0 is not compatible with ClimaOcean 0.10.0

I'll run this locally to check; I think I should merge https://github.com/CliMA/ClimaCoupler.jl/pull/2039 first ; it should simplify the deps a bit ; (there may still be issues with ClimaOcean latest at that stage but it should be a simpler dep tree to manage)  

This PR's changes compiled successfully (local) on Julia 1.11.9 on the latest commit with this branch (there is a warning `could not import ORCA.ORCA1 into OceanConfigurations`); ClimaOcean0.10.0 adds NumericalEarth main to its sources so I don't expect compat issues there.

hgpeterson (approved):

PR #2059, opened 2026-07-28
Final removal of ClimaOcean

simone-silvestri:
This PR completely removes ClimaOcean by using hardcoded data passed through by clima artifacts

PR #2060, opened 2026-07-28
Fix error from ocean's output writer

simone-silvestri:
`CO` is still wired in which means that temperature and salinity top bcs pick from NumericalEarth where we use real fluxes for salinity and temperature, which in turn means that T and S top boundary conditions are `DiscreteBoundaryFunctions` which we need to extract.

This correction will not be needed after #2059, but is propaedeutic to when we will need to implement real fluxes


daverumph (approved):
This looks good to me.  As the commit title says, it's a simple fix.

PR #2064, opened 2026-07-29
change coupler logo

szy21:
## Content

----


szy21:
merging it manually as it seems there is no objection:)

PR #2069, opened 2026-08-05
Walltime reporting for debugging

hgpeterson:
Geometric walltime reporting was removed in #2026, but some find it useful for debugging, so I've added it back in here in the form of a config flag: 

- **`walltime_debug::Bool`** (default `false`) — also report the walltime on power-of-two steps, on top of the `walltime_dt` interval. If `walltime_dt` is `"never"`, those steps become the only reporting.

To accomplish this, I've added two new schedules to `TimeManager`:

- **`PowerOfTwoSchedule`** — true on coupling steps 1, 2, 4, 8, 16, …
- **`OrSchedule(schedules...)`** — true when any child schedule is true.

`maybe_trigger_callback` now passes `(; t, step)`, where `step` is a new field in `CoupledSimulation`. Note: `step` restarts to 0 after a restart, so just counts the number of steps in the *current* run.

All coupler schedules now live in `TimeManager` and subtype `ClimaDiagnostics.Schedules.AbstractSchedule` so that they can be used for diagnostics (and could in principle be moved to `ClimaDiagnostics`/`ClimaUtilities` someday).

All CI, nightly, and longrun configs now have `walltime_debug: true`.

(I also noticed that two `Input.jl` help strings were wrong, so I fixed those).


src/TimeManager.jl
```diff
@@ -62,27 +62,46 @@ simulated_years_per_day(t_start, t_end, walltime) =
...
 """
-struct Callback{SCHEDULE, FUNC}
-    schedule::SCHEDULE
-    func::FUNC
+struct Callback{S, F}
+    schedule::S
+    func::F
 end
 
+# TODO: Add `step` to the coupler struct?
```
ph-kev: I think it would be okay to add `step` as a field to the coupler struct.

src/TimeManager.jl
```diff
@@ -62,27 +62,46 @@ simulated_years_per_day(t_start, t_end, walltime) =
 """
     Callback
 
-A small struct containing a schedule, and the function to be executed if the
+A small struct containing a schedule and the function to be executed if the
 schedule is true.
 
-A `schedule` is a callable object (ie, a function) that takes an integrator-type
-of object and returns true or false.
+A `schedule` is a callable object (ie, a function) that returns true or false. It
```
ph-kev:
```suggestion
A `schedule` is a callable object (i.e., a function) that returns `true` or `false`. It
```

src/TimeManager.jl
```diff
@@ -62,27 +62,46 @@ simulated_years_per_day(t_start, t_end, walltime) =
...
+"""
+    coupler_step_number(cs)
+
+Return the number of coupling steps taken so far in this run.
+
+This function works for both floating-point times and `ITime`s. Note that this is 
+counted from `cs.tspan[1]`, which is the start of the *current* run and not of the 
+overall simulation.
+"""
+coupler_step_number(cs) = round(Int, (cs.t[] - cs.tspan[1]) / cs.Δt_cpl)
```
ph-kev: I think this can be wrong for very large `cs.t[]` although I don't know if we will ever get there in practice.

src/TimeManager.jl
```diff
@@ -62,27 +62,46 @@ simulated_years_per_day(t_start, t_end, walltime) =
...
 
+# TODO: Add `step` to the coupler struct?
+"""
+    coupler_step_number(cs)
+
+Return the number of coupling steps taken so far in this run.
+
+This function works for both floating-point times and `ITime`s. Note that this is 
+counted from `cs.tspan[1]`, which is the start of the *current* run and not of the 
+overall simulation.
```
ph-kev: Why is this the behavior we want? Right now, this matters for restarted simulation.

src/TimeManager.jl
```diff
@@ -99,17 +118,119 @@ function callbacks!(cs)
...
 
 """
-A schedule that is never true. Useful to disable something.
+    NeverSchedule()
 
-TODO: Move this to ClimaUtilities once we move Schedules there
+A schedule that is never true. Useful to disable something.
 """
-struct NeverSchedule end
+struct NeverSchedule <: CD.Schedules.AbstractSchedule end
```
ph-kev, line 114: This doesn't need to be done now, but it would be good to open an issue stating what a new schedule struct need to interface to be a `ClimaDiagnostics.Schedules.AbstractSchedule`.

Also, is this needed because the schedules are passed to the diagnostics?

src/TimeManager.jl
```diff
@@ -99,17 +118,119 @@ function callbacks!(cs)
...
+end
+
+OrSchedule(schedules...) = OrSchedule(schedules)
+
+function (schedule::OrSchedule)(integrator)::Bool
+    # Must call *every* schedule: some of them (e.g., EveryDtSchedule) only 
+    # advance their internal state when they return true. This is why the 
+    # results are collected before being reduced.
+    triggered = map(s -> s(integrator), schedule.schedules)
+    return any(triggered)
```
ph-kev, line 185: Can you use this form instead: `any(p, itr) -> Bool`? This short-circuits instead of evaluating all of the schedules.

src/TimeManager.jl
```diff
@@ -99,17 +118,119 @@ function callbacks!(cs)
...
+"""
+    walltime_schedule(walltime_dt, walltime_debug, start_date)
+
+Return the schedule to be used for walltime reporting, or `nothing` if walltime
+reporting is disabled.
+
+Reporting happens every `walltime_dt` (a string, as described in
+[`time_to_period`](@ref)), or never if `walltime_dt` is `"never"`.
+
+If `walltime_debug` is true, reporting also happens on every coupling step whose
```
ph-kev:
```suggestion
If `walltime_debug` is `true`, reporting also happens on every coupling step whose
```

ph-kev (approved):

src/TimeManager.jl
hgpeterson: Right, I guess I thought it would be nice to see more frequent logs after a restart just in case anything went wrong with the checkpointing. @trontrytel what do you think?

src/TimeManager.jl
hgpeterson: Yeah I made sure they were `AbstractSchedule`s with `short_name` and `long_name` methods so that they would work in the output diagnostics if you wanted to use them there. Currently `OnceSchedule` is being used there.

src/TimeManager.jl
hgpeterson: I triggered them all on purpose here; some schedules (like `EveryCalendarDtSchedule` are stateful, so they need to be triggered. A short-circuit wouldn't do that. This is described in the comment, but let me know if I can word it better.

src/TimeManager.jl
hgpeterson: I agree, I've done it in the latest commit. For now I set it such that the first step of every simulation (even restarts) is 1... Maybe it would be better to add the field to the checkpoints, though, and then report total step of the simulation.

src/TimeManager.jl
hgpeterson: Removed in favor of `cs.step`

src/TimeManager.jl
trontrytel: I think it makes sense for restarts to have more frequent logs at the beginning. Thank you

trontrytel (approved):
Thank you!

PR #2071, opened 2026-08-07
Seed coupler schedules with t_start so restarts keep calendar phase

nefrathenrici:
This PR adds the function `calendar_dt_schedule`, which returns a schedule that works for restarted simulations. This function could probably live in ClimaDiagnostics. Previously, restarted simulations did not pass `t_start` to the diagnostic schedule, causing it to immediately trigger, as the simulation starts after `t_start`.

### DIFF
--- a/src/TimeManager.jl
+++ b/src/TimeManager.jl
@@ -194,7 +194,25 @@ function CD.Schedules.long_name(schedule::OrSchedule)
 end
 
 """
-    walltime_schedule(walltime_dt, walltime_debug, start_date)
+    calendar_dt_schedule(period, start_date, t_start)
+
+Build an `EveryCalendarDtSchedule` for `period` (a time string or a
+`Dates.Period`) whose `date_last` is seeded from `t_start`.
+
+Without the seeding, a schedule built at `t_start > 0` believes it is overdue and
+fires on the first step of a restarted simulation, staying one coupling step out
+of phase from then on. Seeding `date_last` keeps a restarted schedule on the same
+calendar boundaries as the original run. For a fresh simulation `t_start` is 0 and
+the behavior is unchanged.
+"""
+function calendar_dt_schedule(period, start_date, t_start)
+    period isa Dates.Period || (period = time_to_period(period))
+    date_last = start_date + Dates.Millisecond(round(Int64, 1000 * float(t_start)))
+    return CD.Schedules.EveryCalendarDtSchedule(period; start_date, date_last)
+end
+
+"""
+    walltime_schedule(walltime_dt, walltime_debug, start_date, t_start = 0)
 
 Return the schedule to be used for walltime reporting, or `nothing` if walltime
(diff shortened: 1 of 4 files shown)


test/timemanager_tests.jl
```diff
@@ -116,6 +116,27 @@ end
     @test CD.Schedules.long_name(named) == "every power-of-two iteration or never"
 end
 
+@testset "calendar_dt_schedule" begin
```
ph-kev, line 120: Can you add a test using `ITime` here?

src/TimeManager.jl
```diff
@@ -194,7 +194,25 @@ function CD.Schedules.long_name(schedule::OrSchedule)
...
+Without the seeding, a schedule built at `t_start > 0` believes it is overdue and
+fires on the first step of a restarted simulation, staying one coupling step out
+of phase from then on. Seeding `date_last` keeps a restarted schedule on the same
+calendar boundaries as the original run. For a fresh simulation `t_start` is 0 and
+the behavior is unchanged.
+"""
+function calendar_dt_schedule(period, start_date, t_start)
+    period isa Dates.Period || (period = time_to_period(period))
+    date_last = start_date + Dates.Millisecond(round(Int64, 1000 * float(t_start)))
+    return CD.Schedules.EveryCalendarDtSchedule(period; start_date, date_last)
```
ph-kev: For `ITime` inputs, I think you can just get the date directly?

src/TimeManager.jl
```diff
@@ -194,7 +195,29 @@ function CD.Schedules.long_name(schedule::OrSchedule)
 end
 
 """
-    walltime_schedule(walltime_dt, walltime_debug, start_date)
+    calendar_dt_schedule(period, start_date, t_start)
+
+Build an `EveryCalendarDtSchedule` for `period` (a time string or a
```
ph-kev: What is a time string? Maybe it is not obvious to me what is doing the conversion from a string to a `Dates.Period` object.

src/TimeManager.jl
```diff
@@ -194,7 +195,29 @@ function CD.Schedules.long_name(schedule::OrSchedule)
...
+`Dates.Period`) whose `date_last` is seeded from `t_start`. `t_start` may be an
+`ITime`, in which case its date is used directly, or a number of seconds since
+`start_date`.
+
+Without the seeding, a schedule built at `t_start > 0` believes it is overdue and
+fires on the first step of a restarted simulation, staying one coupling step out
+of phase from then on. Seeding `date_last` keeps a restarted schedule on the same
+calendar boundaries as the original run. For a fresh simulation `t_start` is 0 and
+the behavior is unchanged.
```
ph-kev: Can you make this into a admonition and clean it up by telling the user why it should be used? I think it can be a but shorter.

ph-kev (approved):

PR #2072, opened 2026-08-07
Add parameter sweep utility

nefrathenrici:
Adds `experiments/calibration/param_sweep`: a sensitivity sweep for the coupled AMIP model, driven by a small Julia experiment file with two required inputs: a coupler YAML and the parameters to vary.

How it works

- A sweep is a single ClimaCalibrate iteration with no ensemble update, so each member is an independent forward run of the same CouplerModelInterface the amip calibration uses.
- For each parameter to vary, you can set a fixed value, a list to iterate over, or a distribution.
- Parameter names are validated against the ClimaParams registry and the coupler TOML files before running

See README: experiments/calibration/param_sweep/README.md

Usage: `julia --project=experiments/AMIP experiments/calibration/param_sweep/run_sweep.jl <sweep_file.jl>`. The example file given is `experiments/calibration/param_sweep/example_sweep.jl`.

TODO
- Add tests. I'm not sure where to test this, ideally Derecho or our GPU server.
- We could probably generalize this and move it to ClimaCalibrate, but I am opening a PR here to make it available earlier.

### DIFF
--- a/experiments/calibration/param_sweep/README.md
+++ b/experiments/calibration/param_sweep/README.md
@@ -46,8 +46,9 @@ wall time.
 Rerunning the experiment file skips members that ClimaCalibrate recorded
 as completed, and a member that stopped partway resumes from its last
 checkpoint, so a partially failed sweep is resumable with the same
-command. The experiment file is the only record of the design, so keep it
-next to the sweep.
+command. A resumed member's diagnostics are affected by the restart, see
+Costs below. The experiment file is the only record of the design, so
+keep it next to the sweep.
 
 ## What a sweep produces
 
@@ -67,7 +68,14 @@ capped at 64 to catch a value list that accidentally asks for hundreds of
 these; raise `max_members` if a large sweep is what you want.
 
 Derecho charges exclusive jobs for all 4 GPUs of a node, so `q = "preempt"`
-in `worker_options` costs 20% of the regular charge. Members checkpoint
-every 10 simulated days and resume from the last checkpoint, so a
-preempted member loses at most 10 days of work; the tradeoff is that its
-walltime becomes unpredictable.
+in `worker_options` costs 20% of the regular charge, at the cost of an
+unpredictable walltime.
+
+A preempted member loses at most 10 simulated days of stepping, but the
+diagnostics are not that safe. Checkpoints are every 10 simulated days
+and do not line up with the monthly averaging windows the diagnostics
+use, and a checkpoint does not store the partially accumulated averages.
+A month that a restart falls inside is therefore averaged over only the
+part after the restart. Use `preempt` when the sweep is exploratory, and
+the regular queue when the diagnostics have to be comparable across
+members.
--- a/experiments/calibration/param_sweep/README.md
+++ b/experiments/calibration/param_sweep/README.md
@@ -1,0 +1,73 @@
+# Parameter sweep utility
+
+Run an AMIP parameter sweep from two inputs: a coupler YAML and the
+parameters to vary. Parameters mix freely: exact value lists, priors
+(swept over their quantiles), and pinned scalars. The members are every
+combination of the swept values. The sweep is a single ClimaCalibrate
+iteration with no ensemble update, so each member is an independent
+forward run of the same `CouplerModelInterface` that the `amip`
+calibration uses. Each member writes its own model output; comparing the
+members is left to you.
+
+A prior is swept over evenly spaced quantiles, `i / (prior_points + 1)`
+for `i in 1:prior_points`, so the default of 3 points gives the 25th,
+50th, and 75th percentiles. The endpoints are left out because a prior's
+0th and 100th percentiles are often unphysical.
+
+## Quickstart
+
+1. Clone ClimaCoupler on Derecho.
+2. Copy `example_sweep.jl` and edit the inputs. Point `output_dir` at
+   scratch.
+3. From the repository root:
+
+```
+module purge && module load climacommon/2025_02_25
+julia --project=experiments/AMIP -e 'using Pkg; Pkg.instantiate()'
+julia --project=experiments/AMIP \
+    experiments/calibration/param_sweep/run_sweep.jl my_sweep.jl
+```
+
+This checks every parameter name against the ClimaParams registry (a
+typo is a hard failure before any GPU time), then `add_workers` submits
+one GPU worker per member and the `WorkerBackend` hands members to
+workers as they free up. The driver has to stay alive while the members
+run, so start it in a session that survives, for example with `nohup`,
+`screen`, or `tmux`. Workers are cancelled when the driver exits. Pass
+`--no-submit` to only write the member inputs out.
+
+Set `n_workers` to run the members on fewer workers than there are
+members. Pick a divisor of the member count: any remainder leaves a final
+wave with idle workers, and a wave takes as long as its slowest member
+either way. With 9 members, 9 or 3 workers keep every worker busy, while
+8 leaves one member to run alone in a second wave and roughly doubles the
+wall time.
+
+Rerunning the experiment file skips members that ClimaCalibrate recorded
+as completed, and a member that stopped partway resumes from its last
+checkpoint, so a partially failed sweep is resumable with the same
+command. The experiment file is the only record of the design, so keep it
+next to the sweep.
+
+## What a sweep produces
+
+Everything lands in `output_dir`:
+
+- `iteration_000/member_NNN/parameters.toml`: the values each member ran
+  with.
+- `iteration_000/member_NNN/`: each member's model output, holding the
+  diagnostics the coupler YAML asks for. The forward model logs go to the
+  workers instead, in `worker_N.log` next to where the driver runs.
+
+## Costs
+
+One member is a 38-day AMIP run (one week of spin-up plus the sampled
+month), about 1.2 wall hours on one Derecho GPU node. The member count is
+capped at 64 to catch a value list that accidentally asks for hundreds of
+these; raise `max_members` if a large sweep is what you want.
+
+Derecho charges exclusive jobs for all 4 GPUs of a node, so `q = "preempt"`
+in `worker_options` costs 20% of the regular charge. Members checkpoint
+every 10 simulated days and resume from the last checkpoint, so a
+preempted member loses at most 10 days of work; the tradeoff is that its
+walltime becomes unpredictable.
--- a/experiments/calibration/param_sweep/example_sweep.jl
+++ b/experiments/calibration/param_sweep/example_sweep.jl
@@ -1,0 +1,39 @@
+# Example parameter sweep. Copy, edit, then from the repository root:
+#
+#   julia --project=experiments/AMIP \
+#       experiments/calibration/param_sweep/run_sweep.jl my_sweep.jl
+#
+# Two inputs: config_file (which model) and parameters (what to vary).
+
+config_file =
+    joinpath(pkgdir(ClimaCoupler), "config", "amip_configs", "amip_calibration.yml")
+
+# Absolute path on scratch. Model output is large, never use home.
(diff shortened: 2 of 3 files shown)


experiments/calibration/param_sweep/example_sweep.jl
```diff
@@ -0,0 +1,43 @@
...
+#   julia --project=experiments/AMIP \
+#       experiments/calibration/param_sweep/run_sweep.jl my_sweep.jl
+#
+# Two inputs: config_file (which model) and parameters (what to vary).
+
+config_file =
+    joinpath(pkgdir(ClimaCoupler), "config", "amip_configs", "amip_calibration.yml")
+
+# Absolute path on scratch. Model output is large, never use home.
+output_dir = "/glade/derecho/scratch/CHANGE_ME/amip_my_sweep"
```
ph-kev, line 12: Would it be possible to throw an error if this isn't changed?

experiments/calibration/param_sweep/example_sweep.jl
```diff
@@ -0,0 +1,43 @@
...
+
+# Diagnostic variables for the response analysis (monthly means from the
+# model output). Optional, this is the default.
+short_names = ["lwp", "pr", "rsut", "rlut", "tas"]
+
+# Per parameter: a vector sweeps exactly those values, a prior sweeps its
+# quantiles (default 3, set prior_points to change), and a scalar pins the
+# value in every member. The ensemble is the full factorial across the swept
```
ph-kev: I think the proper term is cartesian product and not factorial. You can also use the word "combination" too.

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,76 @@
...
+Run an AMIP parameter sweep from two inputs: a coupler YAML and the
+parameters to vary. Parameters mix freely: exact value lists, priors
+(swept over their quantiles), and pinned scalars, expanded to the full
+factorial. The sweep is a single ClimaCalibrate iteration with no
+ensemble update, so each member is an independent forward run of the same
+`CouplerModelInterface` that the `amip` calibration uses. The analysis
+compares the members' monthly-mean output against near-identical
+replicate runs with a latitude-weighted RMS difference, so it measures
+whether each parameter moves the model above its internal variability.
+There is no comparison to observations.
```
ph-kev: Could we separate the parameter sweep from the analysis? I can imagine some use case where people are interested in running simulation over different sets of parameters but want to do their own analysis.

experiments/calibration/param_sweep/example_sweep.jl
```diff
@@ -0,0 +1,43 @@
...
+
+config_file =
+    joinpath(pkgdir(ClimaCoupler), "config", "amip_configs", "amip_calibration.yml")
+
+# Absolute path on scratch. Model output is large, never use home.
+output_dir = "/glade/derecho/scratch/CHANGE_ME/amip_my_sweep"
+
+# Diagnostic variables for the response analysis (monthly means from the
+# model output). Optional, this is the default.
+short_names = ["lwp", "pr", "rsut", "rlut", "tas"]
```
ph-kev: Is this hard coded to be monthly averages?

experiments/calibration/param_sweep/example_sweep.jl
```diff
@@ -0,0 +1,43 @@
...
+
+# Near-copies of the first member that measure the internal-variability
+# noise floor.
+replicates = 3
+
+# Options for ClimaCalibrate.add_workers. Optional, these are the defaults.
+# `time` is the worker walltime in minutes. Anything add_workers does not
+# consume goes to the scheduler, so `q = "preempt"` (20% charge) is the cheapest
+# way to run: sweep members are independent and resumable, so preemption costs
+# the least here.
```
ph-kev: Is the comment outdated or just giving an example? I am not sure how good/bad preemption is for this since the simulation need to reach a checkpoint.

experiments/calibration/param_sweep/example_sweep.jl
```diff
@@ -0,0 +1,43 @@
...
+output_dir = "/glade/derecho/scratch/CHANGE_ME/amip_my_sweep"
+
+# Diagnostic variables for the response analysis (monthly means from the
+# model output). Optional, this is the default.
+short_names = ["lwp", "pr", "rsut", "rlut", "tas"]
+
+# Per parameter: a vector sweeps exactly those values, a prior sweeps its
+# quantiles (default 3, set prior_points to change), and a scalar pins the
+# value in every member. The ensemble is the full factorial across the swept
+# parameters, here 3 x 3 = 9 members.
```
ph-kev:
```suggestion
# parameters, here 3 x 3 x 1 = 9 members.
```

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,76 @@
+# Parameter sweep utility
+
+Run an AMIP parameter sweep from two inputs: a coupler YAML and the
+parameters to vary. Parameters mix freely: exact value lists, priors
+(swept over their quantiles), and pinned scalars, expanded to the full
```
ph-kev: What values of quantiles are being used here?

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Quickstart
+
+1. Clone ClimaCoupler on Derecho.
+2. Copy `example_sweep.jl` and edit the inputs. Point `output_dir` at
+   scratch.
+3. From the repository root:
+
+```
+module purge && module load climacommon/2025_02_25
+JULIA=/glade/campaign/univ/ucit0011/software/julia/julia-1.11.3/bin/julia
```
ph-kev: Would it not be enough to use `julia` loaded from the modules?

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,76 @@
...
+    experiments/calibration/param_sweep/run_sweep.jl my_sweep.jl
+```
+
+This checks every parameter name against the ClimaParams registry (a
+typo is a hard failure before any GPU time), then `add_workers` submits
+one GPU worker per member and the `WorkerBackend` runs the members on the
+pool as the workers join. Set `n_workers` in the experiment file to use
+fewer workers than members; the members then run in waves. The driver has
+to stay alive while the members run, so start it in a session that
+survives, for example with `nohup` or `screen`. Workers are cancelled when
```
ph-kev:
```suggestion
survives, for example with `nohup`, `screen`, or `tmux`. Workers are cancelled when
```

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,76 @@
...
+$JULIA --project=experiments/AMIP -e 'using Pkg; Pkg.instantiate()'
+$JULIA --project=experiments/AMIP \
+    experiments/calibration/param_sweep/run_sweep.jl my_sweep.jl
+```
+
+This checks every parameter name against the ClimaParams registry (a
+typo is a hard failure before any GPU time), then `add_workers` submits
+one GPU worker per member and the `WorkerBackend` runs the members on the
+pool as the workers join. Set `n_workers` in the experiment file to use
+fewer workers than members; the members then run in waves. The driver has
```
ph-kev: A recommendation of what number of workers to choose would be good. For example, choose `n_workers = 10` when the number of runs is 11 is not a good idea.

experiments/calibration/param_sweep/run_sweep.jl
```diff
@@ -0,0 +1,401 @@
...
+                "parameter $n must be a number, a vector of values, or a " *
+                "ParameterDistribution, got $(typeof(spec))",
+            )
+        end
+    end
+    n_members = prod(length.(value_axes))
+    n_members <= 64 || error(
+        "The factorial sweep has $n_members members. Shrink the value lists " *
+        "or pin more parameters (limit 64).",
+    )
```
ph-kev: Is there a reason why we limit the number of ensemble members to less than 64?

experiments/calibration/param_sweep/example_sweep.jl
nefrathenrici: Done, read_experiment now errors if output_dir still contains CHANGE_ME

experiments/calibration/param_sweep/example_sweep.jl
nefrathenrici: Agreed, I replaced it with "every combination of the swept values"

experiments/calibration/param_sweep/README.md
nefrathenrici: I removed the analysis, it is not that useful and can easily be added later.

experiments/calibration/param_sweep/example_sweep.jl
nefrathenrici: outdated, I removed it

experiments/calibration/param_sweep/README.md
nefrathenrici: 3 per prior: 25th, 50th, and 75th

experiments/calibration/param_sweep/README.md
nefrathenrici: yes

experiments/calibration/param_sweep/run_sweep.jl
nefrathenrici: It's just intended to prevent huge ensembles, no cap is really needed.

experiments/calibration/param_sweep/README.md
nefrathenrici: recommendation:
Set `n_workers` to run the members on fewer workers than there are
members. Pick a divisor of the member count: any remainder leaves a final
wave with idle workers, and a wave takes as long as its slowest member
either way. With 9 members, 9 or 3 workers keep every worker busy, while
8 leaves one member to run alone in a second wave and roughly doubles the
wall time.

experiments/calibration/param_sweep/README.md
```diff
@@ -0,0 +1,73 @@
...
+One member is a 38-day AMIP run (one week of spin-up plus the sampled
+month), about 1.2 wall hours on one Derecho GPU node. The member count is
+capped at 64 to catch a value list that accidentally asks for hundreds of
+these; raise `max_members` if a large sweep is what you want.
+
+Derecho charges exclusive jobs for all 4 GPUs of a node, so `q = "preempt"`
+in `worker_options` costs 20% of the regular charge. Members checkpoint
+every 10 simulated days and resume from the last checkpoint, so a
+preempted member loses at most 10 days of work; the tradeoff is that its
+walltime becomes unpredictable.
```
ph-kev: Doesn't the checkpointing need to match the schedules of the diagnostic for this to be true?

experiments/calibration/param_sweep/run_sweep.jl
```diff
@@ -0,0 +1,277 @@
...
+
+# Guard against a value list that accidentally asks for hundreds of 1.2-hour GPU
+# runs. Raise it with `max_members` in the experiment file if a large sweep is
+# what you want.
+const DEFAULT_MAX_MEMBERS = 64
+
+# `short_names` is required by `CalibrateConfig` because a calibration needs
+# observations to compare against. A sweep never reads them: the diagnostics
+# each member writes are whatever the coupler YAML asks for.
+const PLACEHOLDER_SHORT_NAMES = ["ta"]
```
ph-kev: Could we just hardcode this instead if the user don't need to deal with it and move the comment over there too?

ph-kev (approved):

experiments/calibration/param_sweep/README.md
nefrathenrici: Yes, I extended the comment to clarify that diagnostics are generally not saved properly when restarting and that preempt queue should mainly be used for exploratory sweeps.

PR #2075, opened 2026-08-10
use atmos 1m config in amip 1m longrun

szy21:
## Purpose 

Missed in #2074.

## Content

----


szy21:
manually merging this as it only changes longrun config.

PR #2078, opened 2026-08-11
Update Manifests to use CC 0.15

imreddyTeja:
Updates Manifest to use new version of ClimaCore and ClimaLand


szy21:
It seems ClimaLand is not changed except for AMIP/Manifest.toml. Is this intended?

imreddyTeja:
> It seems ClimaLand is not changed except for AMIP/Manifest.toml. Is this intended?

No it was not... good catch. The Julia package registry has been extremely slow since yesterday, so releases are not visible to the package manager. I'll fix it by manually overriding my local. registry settings

imreddyTeja:
@szy21 There is a conflict in the CMIP environment that prevents using the latest ClimaCore release. Should I wait to update the dependencies until both the AMIP and CMIP envs can use the latest ClimaCore? I'd need to update ConservativeRegridding and wait for them to release it

szy21:
> @szy21 There is a conflict in the CMIP environment that prevents using the latest ClimaCore release. Should I wait to update the dependencies until both the AMIP and CMIP envs can use the latest ClimaCore? I'd need to update ConservativeRegridding and wait for them to release it

Yes, let's wait for that.

akshaysridhar:
https://github.com/JuliaGeo/ConservativeRegridding.jl/pull/130 cc @szy21 @imreddyTeja

szy21 (approved):

PR #2079, opened 2026-08-12
Test CloudMicrophysics on main in nightly pipeline

petebachant:
AMIP 1M performance seems to be driven by changes in CloudMicrophysics, so it would be nice to see those consequences without needing to release and propagate through the stack.

Will this actually work without changing compat entries though?


szy21 (approved):

szy21:
It should work

ph-kev (approved):

szy21:
manually merging this as it only changes the nightly pipeline.

PR #2080, opened 2026-08-12
Add hotfix for LAI PeriodicCalendar bug.

costachris:
Add temporary hotfix for LAI PeriodicCalendar bug, which resulted in a shift of climatological LAI values as years deviated from 2000.


szy21:
Do we want to call the prescribed_climatological_lai_modis directly?

kmdeck (approved):

szy21 (approved):

PR #2090, opened 2026-08-20
Use ClimaParams for some ocean parameters

hgpeterson:
In the future, we should add *all* ocean/sea-ice parameters to ClimaParams and load them from there in the Coupler. But for now, this PR loads the following parameters from ClimaParams in the CMIP OceananigansSimulation:
- `gravitational_acceleration` (was: `9.80665`, now: `9.81`)
- `angular_velocity_planet_rotation` (was: `7.292115e-5`, now: `7.2921159e-5`)
- `density_ocean_reference` (was: `1026`, now: `1035`) <- this one is the largest change; perhaps worth changing the value in ClimaParams if we want to keep 1035? @simone-silvestri 
- `planet_radius` (was: `6.371e6`, now: `6371000`)
- `specific_heat_ocean` (was: `3991`, now: `3991.86795711963`)

While I was at it, I also changed the default slab ocean albedo from 0.38 to `idealized_ocean_albedo` (also 0.38).


simone-silvestri:
0.38 for the ocean albedo seems gigantic (typically it is around 0.05-0.1)

simone-silvestri (approved):

akshaysridhar (approved):

hgpeterson:
> 0.38 for the ocean albedo seems gigantic (typically it is around 0.05-0.1)

Agreed! I think it's because this value is meant to be used in very idealized setups with just a slab ocean and no clouds, so it acts as more of an Earth average albedo. In ClimaParams, they cite [O'Gorman and Schneider (2008)](https://journals.ametsoc.org/downloadpdf/view/journals/clim/21/15/2007jcli2065.1.pdf) for this value. When we go through all the ocean/sea-ice parameters and add them to ClimaParams, maybe we can consider renaming this one to be more clear.

PR #2096, opened 2026-08-31
Reduce AMIP pipeline limit to 36hours

nefrathenrici:
This PR removes the global time limit from the AMIP pipeline and reduces the actual run's time limit to 36 hours.

### DIFF
--- a/.buildkite/amip/pipeline.yml
+++ b/.buildkite/amip/pipeline.yml
@@ -1,6 +1,5 @@
 agents:
   queue: central
-  slurm_time: 96:00:00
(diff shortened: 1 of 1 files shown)


.buildkite/amip/pipeline.yml
szy21, line 11: do we need to change this?

szy21 (approved):

.buildkite/amip/pipeline.yml
nefrathenrici: This adds a time limit to buildkite rather than slurm, it is not needed but does not need to be changed.

PR #2102, opened 2026-09-04
Always use zstar grid

hgpeterson:
`simple_ocean = true` (used in CI and software testing) previously used a linear free surface, which turned out to dramatically violate energy conservation.


simone-silvestri (approved):

hgpeterson:
Agh, this runs out of parameter memory though...

PR #2105, opened 2026-09-08
Compute sea salt emissions as part of turbulent surface fluxes

ziglaser:
Restructure in light of comments on Atmos PR  @https://github.com/CliMA/ClimaAtmos.jl/pull/4756, adding `compute_sslt_emission!` to FluxCalculator, and calling it from `compute_surface_fluxes`. A few MOST variables are added to `get_surface_fluxes` to enable computation of 10m winds.

----
- [x ] I have read and checked the items on the review checklist.

### DIFF
--- a/config/atmos_configs/climaatmos_wx_progedmf_SSA.yml
+++ b/config/atmos_configs/climaatmos_wx_progedmf_SSA.yml
@@ -1,0 +1,37 @@
+aerosol_radiation: true
+approximate_linear_solve_iters: 2
+cloud_model: "quadrature"
+dt: "40secs"
+dt_cloud_fraction: "1hours"
+dt_rad: "1hours"
+dt_save_state_to_disk: "30days"
+dz_bottom: 30.0
+edmfx_detr_model: "Generalized"
+edmfx_entr_model: "Generalized"
+edmfx_filter: true
+edmfx_nh_pressure: true
+edmfx_sgs_diffusive_flux: true
+edmfx_sgs_mass_flux: true
+edmfx_upwinding: "first_order"
+edmfx_vertical_diffusion: true
+h_elem: 30
+implicit_diffusion: true
+insolation: "timevarying"
+netcdf_output_at_levels: true
+ode_algo: ARS222
+prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "DST05", "OC1", "OC2", "SO4", "SSLT01", "SSLT02", "SSLT03", "SSLT04", "SSLT05"]
+prognostic_aerosols: ["SSLT"]
+prognostic_tke: true
+rad: "allskywithclear"
+rayleigh_sponge: true
+surface_setup: "DefaultMoninObukhov"
+t_end: "120days"
+time_varying_trace_gases: ["CO2", "O3"]
+toml: [toml/prognostic_edmfx_1M.toml]
+turbconv: "prognostic_edmfx"
+viscous_sponge: true
+z_elem: 63
+z_max: 60000.0
+microphysics_model: "1M"
+non_orographic_gravity_wave: true
+orographic_gravity_wave: "raw_topo"
--- a/ext/ClimaCouplerClimaAtmosExt.jl
+++ b/ext/ClimaCouplerClimaAtmosExt.jl
@@ -559,6 +559,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
 end
 
 """
+    FluxCalculator.compute_sslt_emission!(csf, ocean_sim, atmos_sim::ClimaAtmosSimulation, fluxes)
+
+Compute sea salt emission mass flux using ocean `area_fraction` and MOST solution, 
+writing to Atmos cache via`ClimaAtmos.set_sslt_surface_fluxes!`. Called from [`compute_surface_fluxes!`](@ref).
+Requires Atmos to carry prognostic sea salt.
+"""
+function FluxCalculator.compute_sslt_emission!(
+    csf,
+    ocean_sim::Union{Interfacer.AbstractOceanSimulation, Models.PrescribedOceanSimulation},
+    atmos_sim::ClimaAtmosSimulation,
+    fluxes,
+)
+    FT = eltype(atmos_sim.integrator.u)
+    p = atmos_sim.integrator.p
+    hasproperty(p.tracers, :sslt_sfc_fluxes) || return nothing
+    sfp = FluxCalculator.get_surface_params(atmos_sim)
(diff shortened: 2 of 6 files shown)


ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -558,6 +558,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
...
+"""
+function FluxCalculator.compute_sslt_emission!(
+    csf,
+    ocean_sim::Union{Interfacer.AbstractOceanSimulation, Models.PrescribedOceanSimulation},
+    atmos_sim::ClimaAtmosSimulation,
+    fluxes,
+)
+    FT = eltype(atmos_sim.integrator.u)
+    p = atmos_sim.integrator.p
+    hasproperty(p.tracers, :sslt_sfc_fluxes) || return nothing
```
trontrytel, line 576: This will just give unused key warning rather than fail. It would be better to do something like what radiation is doing `hasradiation(integrator) =  !isnothing(integrator.p.atmos.radiation_mode)`. This way the choice is done based on Atmos model struct and not the names of fields in Atmos cache.

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -558,6 +558,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
...
+    sfp = FluxCalculator.get_surface_params(atmos_sim)
+    area_fraction = Interfacer.get_field(ocean_sim, Val(:area_fraction))
+
+    u₁₀ = csf.scalar_temp3
+    @. u₁₀ = CA.wind_at_height(FT(10), fluxes.ustar, fluxes.L_MO, sfp)
+
+    u₁₀_atmos = p.scratch.ᶠtemp_field_level
+    ocean_fraction_atmos = p.scratch.ᶠtemp_field_level_2
```
trontrytel, line 584: I already commented in Atmos. I don't think we need a second scratch field. We could compute the scalar wind factor, ocean fraction times the wind power law, on the coupler space, remap it once, and pass one field. That deletes the new scratch field from the atmosphere PR, halves the remaps, and is the correct order of operations because remapping is linear.

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -558,6 +558,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
...
+    atmos_sim::ClimaAtmosSimulation,
+    fluxes,
+)
+    FT = eltype(atmos_sim.integrator.u)
+    p = atmos_sim.integrator.p
+    hasproperty(p.tracers, :sslt_sfc_fluxes) || return nothing
+    sfp = FluxCalculator.get_surface_params(atmos_sim)
+    area_fraction = Interfacer.get_field(ocean_sim, Val(:area_fraction))
+
+    u₁₀ = csf.scalar_temp3
```
trontrytel, line 580: Claude pointed out that `scalar_temp3` is used in `get_roughness_params`. It's fine in this configuration, but an ocean type with constant roughness would let the wind overwrite the roughness field while it is still referenced. I'm not sure if such a configuration is feasible though.

szy21: yes, we should make the constant roughness work too.

config/atmos_configs/climaatmos_wx_progedmf_SSA.yml
```diff
@@ -0,0 +1,37 @@
+aerosol_radiation: true
```
trontrytel, line 1: A lot of things are duplicated here from other configs. Could you follow the pattern and for example re-sue the atmos config as we do in amip or weatherquest setups?

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -558,6 +558,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
     return nothing
 end
 
+"""
+    FluxCalculator.compute_sslt_emission!(csf, ocean_sim, atmos_sim::ClimaAtmosSimulation, fluxes)
```
trontrytel, line 562: Could you add some unit tests for the FluxCalculator? Would it be possible to also test that the area fraction masking is working and that the per-bin diagnostics are correct?

ext/ClimaCouplerClimaAtmosExt.jl
```diff
@@ -558,6 +558,42 @@ function FluxCalculator.update_turbulent_fluxes!(sim::ClimaAtmosSimulation, fiel
...
+
+Compute sea salt emission mass flux using ocean `area_fraction` and MOST solution, 
+writing to Atmos cache via`ClimaAtmos.set_sslt_surface_fluxes!`. Called from [`compute_surface_fluxes!`](@ref).
+Requires Atmos to carry prognostic sea salt.
+"""
+function FluxCalculator.compute_sslt_emission!(
+    csf,
+    ocean_sim::Union{Interfacer.AbstractOceanSimulation, Models.PrescribedOceanSimulation},
```
trontrytel, line 570: I'm not super familiar with the Coupler patterns. But I think that rather than doing the above it would be better to follow what is done for `roughness_model`.  So something more along these lines:

```
# src/Interfacer.jl: a capability with a safe default
get_field(::AbstractSurfaceSimulation, ::Val{:emits_sea_salt}) = false

# the two ocean models opt in, next to their other getters
Interfacer.get_field(::PrescribedOceanSimulation, ::Val{:emits_sea_salt}) = true
Interfacer.get_field(::SlabOceanSimulation, ::Val{:emits_sea_salt}) = true
# (Oceananigans ocean in the CMIP extension likewise)

# src/FluxCalculator.jl: generic entry point, no knowledge of ocean types
function compute_sslt_emission!(csf, sim, atmos_sim, fluxes)
    Interfacer.get_field(sim, Val(:emits_sea_salt)) || return nothing
    compute_sslt_emission!(csf, sim, atmos_sim, fluxes, Val(true))
end
compute_sslt_emission!(csf, sim, atmos_sim, fluxes, ::Val{true}) = nothing  # atmos models without sea salt

# ext/ClimaCouplerClimaAtmosExt.jl: the ClimaAtmos implementation, one method
function FluxCalculator.compute_sslt_emission!(csf, sim, atmos_sim::ClimaAtmosSimulation, fluxes, ::Val{true})
    hasseasalt(atmos_sim.integrator) || return nothing 
    ... 
end
```

trontrytel (changes requested):
Left some comments. Probably needs revisiting

Issue #2041, opened 2026-07-16
Sync up ocean/sea-ice fluxes


hgpeterson:
The ocean needs to reset its surface temperature at every timestep to get the frazil heat flux right. This requires connecting the two components more explicitly.

Maybe @simone-silvestri can explain this better?

simone-silvestri:
Basically what should happen is that

1) the ocean time steps, the temperature can dip below freezing
2) the ocean temperature is correct by clamping it to freezing. This equates to a heat flux that is collected at the surface (the frazil heat flux)
3) the sea ice time steps and grows following this (negative) heat flux

At the moment, the frazil correction is done in the sea ice - ocean flux computation that is performed at the coupler time step.
This time step can differ from the ocean and sea ice time step. 
Two scenarios:

dt_cpl > dt_ocean = dt_sea_ice

The ocean is not corrected at each time step, ergo we have several time steps with Toce < Tfreezing. When the correction is done, the frazil ice heat flux is computed and then applied multiple times to the sea ice (double counting)

dt_cpl < dt_ocean = dt_sea_ice

The ocean is corrected multiple times (the correction is idempotent), so that we reset the frazil ice heat flux at each subsequent coupling step (such that Q = 0) and the sea ice does not "see" the growth implied by the ocean temperature correction: we loose heat in the system.

The best way to approach this is to have a single
```
step!(ocean, seaice, dt)
```
that steps ocean and sea ice together and computes the exchange fluxes independently from the coupler time step

PR #2111, opened 2026-09-10
Add correct flux accumulation and connect ocean and sea ice

simone-silvestri:
This PR does three main things:
- introduces flux accumulation for radiation and precipitation
- fixes a small bug where the sea-ice atmosphere fluxes did not account for the ice concentration (I flipped-flopped a lot on the convention but now it's stable in ClimaSeaIce, meaning we need per-ice not per-cell fluxes)
- corrects the scaling of temperature in sea-ice solver (I don't think this is the cause of the negative ice temperatures we see, I tried on this branch and we still get significantly negative temperatures)
- Ties the ocean step to the sea ice step and computes the sea-ice ocean fluxes independently of the coupler time step


src/FluxCalculator.jl
hgpeterson, line 89:
```suggestion
Time-accumulator for the fluxes between a surface model and the atmosphere,
```

src/FluxCalculator.jl
```diff
@@ -95,32 +96,53 @@ Used by `turbulent_fluxes!` to push the time-averaged flux to a
...
 surfaces; fast surfaces and implicit-flux surfaces do not have an accumulator.
 """
-struct FluxAccumulator{F <: CC.Fields.Field}
+struct FluxAccumulator{F <: CC.Fields.Field, D <: NamedTuple}
     F_lh::F
     F_sh::F
     F_turb_moisture::F
     F_turb_ρτxz::F
     F_turb_ρτyz::F
+    fluxes::D
```
hgpeterson, line 105: I think `fluxes` would actually hold things like `u_int` for the bucket, so the name is a bit misleading. Maybe something like `coupler_fields`? But then I guess `FluxAccumulator` is also a misleading name...

src/FieldExchanger.jl
```diff
@@ -291,17 +293,35 @@ function update_sim!(sim::Interfacer.AbstractSurfaceSimulation, csf)
...
+
+"""
+    update_model_sims!(model_sims, csf, flux_accumulators = (;))
 
 Iterates `update_sim!` over all component model simulations saved in `cs.model_sims`.
 
+For a surface in `flux_accumulators`, the fields it declared in
+`accumulated_coupler_fields` are also added to the accumulator, and
+`FluxCalculator.push_and_reset!` overwrites them with the window average just before
+the surface steps.
```
hgpeterson, line 312:
```suggestion
`FluxCalculator.push_ready_accumulators!` overwrites them with the 
window average just before the surface steps.
```

src/FluxCalculator.jl
hgpeterson, line 161: Just noting that I think `n_steps` only gets incremented here and not when `fluxes` is accumulated, which is maybe worth documenting.

ext/ClimaCouplerCMIPExt/clima_seaice.jl
hgpeterson, line 316: Should I be worried that these are still instantaneous values and not accumulated? Maybe it makes sense because you want T_s to be for the current step's radiation, but then later the other terms in the heat flux are accumulated, no?

hgpeterson (approved):
Approved; just some minor comments about naming conventions and a question about consistency with the sea-ice skin temperature calculation.

Also, does this PR close #2041?

PR #2117, opened 2026-09-15
Add ClimaTimeSteppers v1 to compat

ph-kev:
This PR adds ClimaTimeSteppers v1 to the compat so that nightly AMIP can run.

Note that compat for nightly AMIP will not instantiate due to CloudMicrophysics. It will be able to instantiate when https://github.com/CliMA/ClimaAtmos.jl/pull/4825 is merged.


imreddyTeja:
should CMIP be bumped as well?

imreddyTeja (approved):

ph-kev:
> should CMIP be bumped as well?

I don't think it matter since you can't use it until ClimaAtmos and ClimaLand make releases (unless you are using the main branches of them), but I updated it anyway.

PR #2118, opened 2026-09-16
Parse YYYYMMDD-HHMM start_date

costachris:
Support arbitrary start dates/times for subseasonal and weatherbench setups (HHMM in ERA5 IC filenames). Pairs with [#4836](https://github.com/CliMA/ClimaAtmos.jl/pull/4836)


src/Input.jl
```diff
@@ -1116,34 +1117,32 @@ function get_era5_filepaths(
 )
     era5_initial_condition_dir = resolve_era5_dir(era5_initial_condition_dir)
     datestr = Dates.format(start_date, Dates.dateformat"yyyymmdd")
+    timestr = Dates.format(start_date, Dates.dateformat"HHMM")
+    stamp = "$(datestr)_$(timestr)"
```
nefrathenrici, line 1121: I think this will silently be inconsistent with Atmos, specifically because `weather_model_data_path` hardcodes the time `_0000`.  `start_date: "20191231-1200"` would pair the 12Z surface state with the 00Z atmosphere state.

src/CalibrationTools.jl
```diff
@@ -932,7 +933,7 @@ precision beyond days (e.g. hours, seconds, etc.) are not used for setting the
 start date.
```
nefrathenrici:
```suggestion
The `start_date` and `end_date` are converted to strings and the keys
"start_date" and "t_end" in `config_dict` are updated accordingly.
```

nefrathenrici (approved):
Thanks!

src/CalibrationTools.jl
nefrathenrici: The suggestion isn't working, the last sentence of this docstring can be removed.

src/Input.jl
costachris: opened [#4836](https://github.com/CliMA/ClimaAtmos.jl/pull/4836)

PR #2120, opened 2026-09-17
Remove land and coupler space flexibility

hgpeterson:
We don't actually have an efficient, conservative SE -> SE regridder, so we never actually use `share_surface_space: false` in practice. The same is true for building a custom land space. To keep things simple, I removed this option (along with `h_elem_coupler` and `nh_ploy_coupler`) and cleaned up anything that dealt with general SE spaces.


costachris (approved):
LGTM

kmdeck (approved):
This seems fine because it simplifies things and I dont think we were able to have the land model run with its own grid anyways. so I dont think we lose any capability here

